PR #69, opened 2013-11-12
Expose public default constructors to enable serialization

scottmeyer:
I would like to be able to serialize `LogEvent` to send it over the wire before writing to a sink.

### DIFF
--- a/src/Serilog/Events/DictionaryValue.cs
+++ b/src/Serilog/Events/DictionaryValue.cs
@@ -27,6 +27,14 @@ namespace Serilog.Events
         private readonly IReadOnlyDictionary<ScalarValue, LogEventPropertyValue> _elements;
 
         /// <summary>
+        /// Construct an empty <see cref="DictionaryValue"/>.
+        /// </summary>
+        public DictionaryValue()
+        {
+            
+        }
+
+        /// <summary>
         /// Create a <see cref="DictionaryValue"/> with the provided <paramref name="elements"/>.
         /// </summary>
         /// <param name="elements">The key-value mappings represented in the dictionary.</param>
--- a/src/Serilog/Events/LogEvent.cs
+++ b/src/Serilog/Events/LogEvent.cs
@@ -30,6 +30,11 @@ namespace Serilog.Events
         private readonly Dictionary<string, LogEventPropertyValue> _properties;
 
         /// <summary>
+        /// Construct an empty <seealso cref="LogEvent"/>.
+        /// </summary>
+        public LogEvent() {}
+
+        /// <summary>
         /// Construct a new <seealso cref="LogEvent"/>.
         /// </summary>
         /// <param name="timestamp">The time at which the event occurred.</param>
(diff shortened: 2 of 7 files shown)


sstorie:
I know this PR was closed, but I'm curious if anyone else has any insight on how this could be done, or if anyone's already created a DTO object to help address use case. I have a REST API that I'd like our client software to use for logging, so I have a custom sink that'll take a rendered LogEvent and send it to the appropriate endpoint. However, I'd like to send the raw LogEvent object and use it to do more interesting processing on the API side.

I understand the reasoning for keeping the LogEvent immutable, but it sure would be nice to have a simple way to send that object over the wire for remote processing. I don't want all the logging configuration to be done on each client application in our environment (we have well over 50 different apps).

I'll keep exploring ways to design the DTO so we could try that approach, but I don't want to re-invent the wheel if I don't need to :)

sstorie:
Ok, this might have been *much* easier than I originally anticipated. Here's what I have after exploring the data types a bit more:

```c#
public class LogEventDto
{
    public DateTimeOffset Timestamp { get; set; }

    public LogEventLevel Level { get; set; }

    public Exception Exception { get; set; }

    public MessageTemplate MessageTemplate { get; set; }

    public List<LogEventProperty> Properties { get; }

    public LogEventDto()
    {
        this.Properties = new List<LogEventProperty>();
    }
}

public static class LogEventExtensions
{
    public static LogEventDto ToLogEventDto(this LogEvent logEvent)
    {
        var logEventDto = new LogEventDto
        {
            Timestamp = logEvent.Timestamp,
            Level = logEvent.Level,
            Exception = logEvent.Exception,
            MessageTemplate = logEvent.MessageTemplate,
        };

        // Convert the logEvent's properties back into a list
        //
        foreach (var entry in logEvent.Properties)
        {
            logEventDto.Properties.Add(new LogEventProperty(entry.Key, entry.Value));
        }

        return logEventDto;
    }

    public static LogEvent ToLogEvent(this LogEventDto logEventDto)
    {
        var logEvent = new LogEvent(logEventDto.Timestamp, logEventDto.Level, logEventDto.Exception, logEventDto.MessageTemplate, logEventDto.Properties);

        return logEvent;
    }
}
```

sstorie:
I've spent the day trying to build out an appropriate DTO object that would let me send the LogEvent over the wire as JSON, but I am finding myself re-building the entire stack of types that go into the LogEvent because of the immutability. I get the nagging feeling that is adding too much coupling between my code and the internals of Serilog just to send this object as JSON, so I'm going to move away from this idea for now. If anyone comes along with a good approach I'd love to hear more.

*I also tried to just use binary serialization and forget about JSON, but these classes aren't marked as serializable so that was a dead end too. I experimented both with BinaryFormatter() and with the SharpSerializer library. Neither one can get around how these classes are designed.*

sstorie:
Ok, one final comment about another non-Serilog problem that makes this challenging to pull off. Even if I came up with some way to serialize and de-serialize the LogEvent so I could send it over the wire, there is a glaring problem. Given Serilog's ability to contain arbitrary classes in the logs, it's possible the client application could include some type that only exists in that app. So when the log hit the server, it wouldn't be able to re-build the LogEvent because there are types that don't exist.

I'm going to forgo any more work on this and create a custom sink that renders the log locally and sends it to the API.

nblumhardt:
Thanks for the follow up @sstorie - I've mostly come to the same conclusions; better to serialize at the source into the destination format and transfer it that way. Cheers!

PR #85, opened 2014-02-15
Add source file / line logging via [CallerFileName] and friends

anaisbetts:
This PR adds source file / line number logging via the C# [CallerFileName] attribute. However, to do this, we have to do some Controversial Stuff (tm), so I'm pushing up this PR before I do the work :cat: 
## The Problem

So, the way you use `[CallerFileName]` is that it must be a Default Parameter. This conflicts Pretty Badly with the `params` keyword, since if the log parameters happen to be `string` and `int`, we now match `sourceFile` and `line`, not the `params` like we wanted. :poop:
## Here's what we do

Instead of using `params`, we T4 Template some explicit overloads via template methods, similar to what we do in ReactiveUI in [VariadicTemplates.cs](https://github.com/reactiveui/ReactiveUI/blob/rxui6-master/ReactiveUI/VariadicTemplates.tt). Something like:

``` cs
public static void Warn<T0>(this ILogger log, string messageTemplate, T0 item0, [CallerFilePath] sourceFile = "", [CallerLineNumber] int line = 0);

public static void Warn<T0,T1>(this ILogger log, string messageTemplate, T0 item0, T1 item1, [CallerFilePath] sourceFile = "", [CallerLineNumber] int line = 0);
```

These are all Extension Methods, because they're super boring, and asking every implementer to implement 3 million methods that just proxy to another one is annoying. 
## TODO:
- [ ] Move most of the methods on `ILogger` onto an extension class that just proxies to `Write(LogEntry entry)`
- [ ] Create a T4 Template to render the overloads of Info / Warn / Etc

### DIFF
--- a/src/Serilog/Events/LogEvent.cs
+++ b/src/Serilog/Events/LogEvent.cs
@@ -28,6 +28,8 @@ namespace Serilog.Events
         private readonly Exception _exception;
         private readonly MessageTemplate _messageTemplate;
         private readonly Dictionary<string, LogEventPropertyValue> _properties;
+        private readonly string _sourceFile;
+        private readonly int _line;
 
         /// <summary>
         /// Construct a new <seealso cref="LogEvent"/>.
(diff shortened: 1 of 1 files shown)


mattxo:
Could it use System.Diagnostics.StackFrame perhaps? 

https://msdn.microsoft.com/en-us/library/system.diagnostics.stackframe.getfilelinenumber(v=vs.110).aspx

SimonCropp:
@mattxo i expect that would be prohibitively expensive in terms of perf cost

mattxo:
@SimonCropp Mm, it's not cheap, about 25ms/call on my machine.

PR #203, opened 2014-08-31
Azure DocumentDB sink

robdmoore:
Notes:

* Not thoroughly tested yet
* Not batching yet although this should be possible to add using a stored proc as per http://azure.microsoft.com/en-us/documentation/articles/documentdb-get-started/
* The SDK is pre-release so likely to change over time - I wonder if we should lock the nuget dependency version for now to avoid later versions breaking this sink?


nblumhardt:
Merged to dev, still straightening out CI but should have this in master/NuGet very soon. Thanks Rob!

nblumhardt:
This has gone out, not sure what happened with the merge. Closing now.

@robdmoore this is moving to https://github.com/serilog/serilog-sinks-azuredocumentdb - just let me know if you'd like to take ownership of that one. Cheers!! :)

robdmoore:
Yeah sure - can't hurt. Good idea splitting them out. I'll watch the Azure ones :)

Issue #290, opened 2014-11-17
ApplicationLifecycleModule request logging should be optional


waynebrantley:
I do want the handler to catch the global errors, but not log every request in production.
This should be configurable.
Perhaps a ApplicationLifecycleModule with delegates for the events - we can specify them or not and we can handle this how we want also:  https://github.com/serilog/serilog/issues/279 and this https://github.com/serilog/serilog/issues/285

Issue #285, opened 2014-11-12
All exceptions are "error caught in global handler"


waynebrantley:
When the 'global handler' catches an exception, it logs the exception with the message above.  This results in a pretty useless looking log:

![image](https://cloud.githubusercontent.com/assets/157082/5017112/75b9ff5c-6a79-11e4-9a9d-44aca22b53ec.png)

Would be better to make it just do a Log.Error(ex) or Log.Error(ex, ex.Message)?

Issue #279, opened 2014-11-11
Configurable log level for "Beginning HTTP Request" in Extras.Web


oliverjanik:
Currently it's hard-coded to Information, which is a bit too high for my taste :-)

PR #292, opened 2014-11-18
Update ApplicationLifecycleModule.cs

waynebrantley:
Add way to hook in and change behavior of events.  https://github.com/serilog/serilog/issues/290, https://github.com/serilog/serilog/issues/279 and https://github.com/serilog/serilog/issues/285


nblumhardt:
Hi Wayne - thanks for the PR! I think we'll pass on this as users who want to significantly customise behaviour can move away from _Serilog.Extras.Web_ to their own implementation with a small amount of work. Cheers!

PR #319, opened 2014-12-05
Added ByTransformingFrom extension method to support destructuring by in...

eamodio:
...terfaces and base classes

(Didn't want to apply it to ByTransforming directly since it would be a breaking change and would likely have performance implications)

### DIFF
--- a/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using System;
+using System.Reflection;
 using Serilog.Core;
 using Serilog.Policies;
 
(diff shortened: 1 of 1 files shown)


nblumhardt:
Hi there!

Slow progress on this one, I know. The "catch-22" we've been in is that the Serilog project became a bit of a sprawl. We've just about finished working through #344, which improves things by moving most sub-projects to their own repositories or independent organisations.

I've started https://github.com/serilog/serilog/issues/414 to move a whole bunch of destructuring-related "extras" to a new project. This'd be a great candidate to have over there, despite being a bit more effort to do. What do you think?

Cheers!
Nick

nblumhardt:
@eamodio the repository split is now done - custom destructuring policies are now over at https://github.com/destructurama - let me know if you want to explore this concept over there. I don't think we'll incorporate this into the core at this point (just a philosophical desire to start developing extensions separately) but thanks very much for the submission. Cheers!

PR #325, opened 2014-12-06
Added option to provide LogentriesSink with ITextFormatter

blachniet:
[Logentries](https://logentries.com/) has support for receiving JSON formatted log events. When it receives JSON logs, it allows users to search for events with certain properties or property values. It also allows you to create tags and alerts based on properties and property values.

Before these changes, the `LogentriesSink` only allowed simple display messages to be sent to Logentries. With these changes, the developer can provide their own `ITextFormatter` when creating the `LogentriesSink`. With this support, a developer could configure the sink to send JSON logs.

```csharp
            .WriteTo.Logentries(logentriesToken, new JsonFormatter(), useSsl:false)
            .CreateLogger();
```

### DIFF
--- a/src/Serilog.Sinks.Logentries/LoggerConfigurationLogentriesExtensions.cs
+++ b/src/Serilog.Sinks.Logentries/LoggerConfigurationLogentriesExtensions.cs
@@ -16,6 +16,7 @@ using System;
 using Serilog.Configuration;
 using Serilog.Events;
 using Serilog.Sinks.Logentries;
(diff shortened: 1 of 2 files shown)


nblumhardt:
Howdy - due to some repository reorganisation this needed to be merged manually; it's now at https://github.com/serilog/serilog-sinks-logentries - which the CI build will push out as 1.5 of the sink shortly.

Cheers!
Nick

PR #339, opened 2014-12-11
Add pre-built binaries for ASP.NET 5

nblumhardt:
VS2015 is in preview and not really suitable as a CI dependency; but, a lot of people are interested in using Serlog on ASP.NET 5.0 core (this was contributed by the ASP.NET team). Checking in the binaries lets the production NUPKG support these people while we wait for a closer-to-final VS release we can use in CI.

(Waiting for the CI build of this PR to verify that all is good. @kanchanm may have some input/corrections on the diff.)


nblumhardt:
#355 is pushing this one forward now. Thanks!

PR #343, opened 2014-12-22
Expanded the log timing abilities

eamodio:
Added BeginTimedScope extension to ILogger which returns a disposable ITimedScope
The BeginTimedScope usage allows for the same logging style as ILogger (messageTemplate and params object[]) to keep consistency as well as more powerful logging
ITimedScope supports the (optional) setting of a completion message (if not used the message provided to BeginTimedScope will be used again with a completion preamble)

Here is an example usage:

```c#
using (var logScope = Logger.BeginTimedDebugScope("{HttpMethod} Request {@Request}", CurrentHttpMethod, request))
{
    try
    {
        var response = ProcessRequest(request);
        logScope.SetCompletetionMessage("{HttpMethod} Response {@Response}", CurrentHttpMethod, response);
        return response;
    }
    catch (Exception ex)
    {
        logScope.SetErrorMessage("{HttpMethod} Request {@Request} Failed {@Exception}", CurrentHttpMethod, request, ex);
    }
}
```

If ProcessRequest succeeded the result would be something like the following:
```
Starting "mFj41fuL7kWvOmIlNw4kg": GET Request FooDto { Bar: null }
Completed "mFj41fuL7kWvOmIlNw4kg" in 00:00:00.0838446 (83 ms): GET Response FooResponseDto { Bar: "Baz" }
```

If ProcessRequest threw then:
```
Starting "mFj41fuL7kWvOmIlNw4kg": GET Request FooDto { Bar: null }
Completed "mFj41fuL7kWvOmIlNw4kg" in 00:00:00.0838446 (83 ms): GET Request FooDto { Bar: null } Failed InvalidOperationException...
```

### DIFF
--- a/src/Serilog.Extras.Timing/Extras/Timing/ITimedScope.cs
+++ b/src/Serilog.Extras.Timing/Extras/Timing/ITimedScope.cs
@@ -1,0 +1,31 @@
+﻿using System;
+
+namespace Serilog.Extras.Timing
+{
+    /// <summary>
+    /// Disposable scope for the timing. Supports explicit completetion messages.
+    /// </summary>
+    public interface ITimedScope : IDisposable
+    {
+        /// <summary>
+        /// Sets an alternative completetion message.
+        /// </summary>
+        /// <param name="messageTemplate"></param>
+        /// <param name="propertyValues"></param>
+        void SetCompletetionMessage(string messageTemplate, params object[] propertyValues);
+
+        /// <summary>
+        /// Sets an alternative completetion message and sets the log level to Error.
+        /// </summary>
+        /// <param name="messageTemplate"></param>
+        /// <param name="propertyValues"></param>
+        void SetErrorMessage(string messageTemplate, params object[] propertyValues);
+
+        /// <summary>
+        /// Sets an alternative completetion message and sets the log level to Warning.
+        /// </summary>
+        /// <param name="messageTemplate"></param>
+        /// <param name="propertyValues"></param>
+        void SetWarningMessage(string messageTemplate, params object[] propertyValues);
+    }
+}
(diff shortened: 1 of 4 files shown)


nblumhardt:
@eamodio sorry about the impasse. @mivano would you like to field this one, or should I take a stab at it? I'm not really familiar with the package/API nor where it's heading. Cheers!

mivano:
I just added pull request #356 which might help although I do not think it solves the above issue completely. I do like the idea however, but we might be able to abstract it a bit and offer it as a process wrapper. So within a using block you start a process, report on the process and call finish when ready. 

So something like

```csharp
using (var process = logger.StartProcess("id etc")){

   try{
      process.Progress("message", percentage);
      process.Add(property);

      // All done
      process.Finish();

      // Or even cancelled
      process.Cancel();
   }
   catch(Exception ex){
      process.Failure(ex, "message");
   }

}

With the finish we assume correct completion and mark it as such in the logger. Otherwise the process was not completed. This way we can pass additional info to the logger, while still being in a scope of a certain process. The timing information we can add directly to it. 

Is this something? I know it is a bit different, but it evolves into a more rich process reported. Looking at your SetError function you might just want also to report a process of something. Let me know what your thoughts are.

nblumhardt:
@eamodio the challenges for us in keeping up with "extras" and so-on has led to a bit of a change in process; this project has now graduated to become the independent [SerilogMetrics](https://github.com/serilog-metrics/serilog-metrics).

I'll close this here but I'm sure your contributions would be appreciated over at the new repository. Cheers!

mivano:
I picked up the progress sample above in my new repository; https://github.com/serilog-metrics/serilog-metrics/pull/6

PR #346, opened 2014-12-24
pass context properties along to nlog

mharen:
Pass all the context properties along to nlog, which can then be included in the [nlog output](https://github.com/NLog/NLog/wiki/EventContext-layout-renderer), e.g. `${event-context:item=somePropertyItem}`.

### DIFF
--- a/src/Serilog.Sinks.NLog/Sinks/NLog/NLogSink.cs
+++ b/src/Serilog.Sinks.NLog/Sinks/NLog/NLogSink.cs
@@ -41,10 +41,9 @@ namespace Serilog.Sinks.NLog
             // pass along the event's properties to nlog
             foreach (var property in logEvent.Properties)
             {
-                // format simple scalar strings without wrapping quotes, which is more likely to be what nlog users expect
-                var format = (property.Value is ScalarValue && property.Value.GetType() == typeof(string))
-                    ? "l" // literal
-                    : null;
+                // format simple scalar strings as literals (without wrapping quotes), which is more likely to be what nlog users expect
+                var sv = property.Value as ScalarValue;
+                var format = (sv != null && sv.Value is string) ? "l" : null;
 
                 nlogEvent.Properties[property.Key] = property.Value.ToString(format, null);
             }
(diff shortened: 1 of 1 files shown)


src/Serilog.Sinks.NLog/Sinks/NLog/NLogSink.cs
```diff
@@ -29,34 +29,45 @@ public void Emit(LogEvent logEvent)
...
             var message = logEvent.RenderMessage(formatProvider);
             var exception = logEvent.Exception;
 
+            var nlogEvent = new LogEventInfo(level, loggerName, message)
+            {
+                Exception = exception
+            };
+
+            // pass along the event's properties to nlog
+            const string literalFormat = "l";
```
nblumhardt: `l` is only valid when the property value is a `ScalarValue` with inner `Value` of type `string` - but otherwise, this seems sane to me :+1:

nblumhardt:
Very nice, thanks!

I think a check of the type of the property value is needed before passing `l` as noted above.

Otherwise :+1: to merging this.

mharen:
Cool, I'll fix it up real quick.

Thanks,
Michael Haren

On Sun, Jan 4, 2015 at 4:44 PM, Nicholas Blumhardt <notifications@github.com

mharen:
Alright, @nblumhardt- should be good now. Let me know if you have any other feedback. Thanks!

src/Serilog.Sinks.NLog/Sinks/NLog/NLogSink.cs
```diff
@@ -29,34 +29,49 @@ public void Emit(LogEvent logEvent)
...
+            var nlogEvent = new LogEventInfo(level, loggerName, message)
+            {
+                Exception = exception
+            };
+
+            // pass along the event's properties to nlog
+            foreach (var property in logEvent.Properties)
+            {
+                // format simple scalar strings without wrapping quotes, which is more likely to be what nlog users expect
+                var format = (property.Value is ScalarValue && property.Value.GetType() == typeof(string))
```
nblumhardt: Should be more like:

```
var sv = property.Value as ScalarValue;
var format = sv != null && sv.Value is string ? "l" : null;
```

I.e. `property.Value.GetType()` will be returning `typeof(ScalarValue)` in the code as it stands. The value we need to check is a subproperty of `ScalarValue`, also called `Value`. Hope this helps :)

mharen: Makes sense, thanks!

nblumhardt:
Great - thanks!

PR #355, opened 2015-01-15
Preliminary support added for vNext

### LINKED ISSUE #353: "System.Linq": "4.0.0-beta-*" in Project.json
Serilog doesn't specify a version number for Linq in it's project json.  This prevents us from importing the Serilog nuget package into a fresh vnext project unless we force a linq version in the consumer project:

{
    "version": "1.0.0-*",
    "dependencies": {
        "Serilog": "1.4.118"
    },
    "frameworks": {
        "aspnetcore50": {
            "dependencies": {
                "System.Linq": "4.0.0-beta-22231"
            }
        }
}

Failure to add the aspnetcore50 dependency results in compilation issues.  I've not entirely got to grips with vnext yet but the current project.json for Serilog prevents it's use in vnext except against aspnetcore50.

Devbot-Net:
- Build script extended to work with vNext projects
    Usage:
        .\build.ps1 "1.4" "0" -notouch # Typical ms build
        .\build.ps1 "1.4" "0" -notouch -r CLR # vnext build (additional options avail)
- project.json now adds support for frameworks Aspnet50, Aspnetcore50 and net45. Where previously this was only aspnetcore50 causing problems when using in projects that targeted the other frameworks (fixes #353).
    - NB. K259/Portability not supported
- vNext Solution Added (add in additional projects as they are made vNext compatible)
- Renamed aspnetcore kproj for Serilog to reflect additional target frameworks


nblumhardt:
Thanks very much! I'll need a little time to look at this, I'll try to make some progress on this within a day or so. @khellang knows the build (and v.Next I'd say) better than I, so might have some input here :)

nblumhardt:
Also, @Devbot-Net - can you please point PRs at the `dev` branch? No need to retarget this one, but if you submit anything else that would be handy. Thanks!

Devbot-Net:
My bad, will do next time.  

We're going to be using this PR for a while so I'll let you go how it goes in the wild.

khellang:
@nblumhardt If we want peeps to send PRs against `dev`, we should set it as default branch :smile:

Devbot-Net:
To be fair, it should be common sense to PR dev I guess.  It's what most of us do with work I expect; at the very least I didn't have a good excuse for going to master.

I'm going to take a few mins seeing if I can get some of the other projects building through vnext (before my boss arrives...).  :)

nblumhardt:
@khellang already is :) ... just might have forked at an earlier time.

image

khellang:
Ah :+1:

Serilog-vNext.sln
```diff
@@ -0,0 +1,39 @@
...
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "assets", "assets", "{E1740E33-407A-42F2-AA7F-B8677BAAD5DB}"
+	ProjectSection(SolutionItems) = preProject
+		README.md = README.md
+		assets\Serilog.snk = assets\Serilog.snk
+	EndProjectSection
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{542E288D-8275-46B0-908B-E9346EB99336}"
+	ProjectSection(SolutionItems) = preProject
+		Build.ps1 = Build.ps1
+		sake.cmd = sake.cmd
```
khellang: Does this file exist anywhere?

Devbot-Net:
*scratches head*.  Not in the repo.

Will add now...

khellang:
This looks good to me. I take it this will build packages for vNext (`aspnet50`, `aspnetcore50`, `net45`). How does the `net45` package work compared to the existing package? Should we push both? Is it needed?

I'm thinking that we somehow should get the vNext binaries into the existing `Serilog` package. Any idea what the best solution for that is? Does NuGet support the new targets yet?

So many questions :stuck_out_tongue_winking_eye:

Devbot-Net:
So, updated my forked dev branch; github is telling me I need to commit against my forked master to update this PR; got a feeling that isn't the case.  Either way, the sln should be there now.

So, I've targetted Serilog at the aspnet's and net45 for now.  Unfortunately support for mobile platforms appears a little lacking in the vnext build process so I'm not able to make this completely portable yet.  (Totally possible that everything is supported and I was doing it wrong; will check out the PackageManager source again when I get a chance).

The net45 package is intended to be equivilant to the existing packages (taking into account the portability caveats above).  Once we can get portable working, the new build process I've added can completely replace the previous build system; the nuget packages created should be no different to how it is built now.

The nuget package version is currently not updated by the build; my current thinking is extending the build.ps1 slightly to replace the version in the project.json's before packaging them as kpm does not have a parameter in the same way nuget.exe does.

Nuget appears to support the new targets; if you unzip the nupkg's created during the vnext build you should find aspnet50 and aspnetcore50 folders under lib.  We could go with a slightly convulted process
- rename nupkg to zip
- unzip package
- copy those lib folders out
- include them in the standard build

That would potentially be a workaround until we can get portability working?  At which point we should be able to replace that process with a straight up kpm install to produce the complete packages?

NB: I'm no expert on this new build process yet, I just know how to do the basics.  Anyone got a friend on the vnext team? :)

khellang:
Hmm. I get this:

> The term 'kpm' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included,  verify that the path is correct and try again.

Devbot-Net:
Under what circumstance do you get that error?  (i.e. During appveyor build, when running locally, when just typing kpm into powershell/cmd?).

khellang:
Just ran it locally. No `KRE_HOME` etc. in the path (from before, at least)

Devbot-Net:
Ah, my mistake.  Seems I have that from a -g install of kre:

    C:\Program Files (x86)\KRE;%USERPROFILE%\.kre;C:\Program Files\KRE

If you change the Install-Kre function to the following, does it work?

    function Install-Kre()
    {
    	$krePath = Join-Path $env:UserProfile "\.kre"
    	$env:Kre_Home = $krePath
        $kvmPath = Join-Path $krePath "\bin\kvm.cmd"
        Write-Output "Using KRE-$runtime-$selectedKreArch.$selectedKreVersion"
        & $kvmPath "install $selectedKreVersion -$selectedKreArch -r $runtime"
        & $kvmPath "use $selectedKreVersion -$selectedKreArch -r $runtime"
    }

khellang:
Also. After I run the script, the CD is `/src/Serilog` :confused:

Devbot-Net:
We can also add $env:KRE_FEED to specify the vnext nuget source if people want to build against something other than KRE/dotnet (it's being renamed) beta1.  They did have some RC builds up last week but they have regressed it to beta3.

That should solve your issues @khellang?  For every pushd the is a popd; now that the script shouldn't error on the KRE_HOME issue it should hit the respective popd that was causing your second issue.

Devbot-Net:
Just noticed (in xunit) how to switch the Serilog project to Portable: ".NETPortable,Version=v4.5,Profile=Profile259"

I'll look at making that change in a couple days when I get a chance.  From there it should be much easier to add in other projects but there's no way I can single-handedly transition all existing projects, I just don't have the time I'm afraid.

nblumhardt:
Hey folks, the landscape has changed quite a lot since this work started out. Thankfully, much less is required to get this up and running today :-)

Thanks for all the effort, I'll open a new PR shortly!

NB

Issue #344, opened 2014-12-22
Split the sinks and extras into separate repositories


nblumhardt:
We've hit 30 sinks - it's no longer really feasible to maintain them all in the same repository, and since the project has stabilised, there's not much advantage to doing so either.

Splitting the project into multiple subprojects will make it easier to manage, and lets contributors with interest in a specific component focus their efforts 

Here's a proposal:

- [ ] Leave only the projects comprising the _Serilog_ NUPKG in the `serilog/serilog` repository
- [ ] For everything else, create a new
  * repository of the form `serilog/serilog-x-y`, for example `serilog/serilog-sinks-couchdb`,
  * _AppVeyor_ build for the package,
  * _README.md_ file in the repository root with more complete documentation
- [ ] Find a single "owner" for each package/repository, either
  * the original contributor of the package, if they're interested, or
  * someone from the core team
- [ ] Make the [Provided Sinks] wiki page an index into each sub-repository (move the docs but keep them discoverable)

I think we should also **turn off** the issue trackers in the sub-repositories and instead point users to the main repository's tracker via a link at the top of the README - otherwise issues with a cross-cutting element might not be handled optimally.

Thoughts?

nblumhardt:
Alright, we're in full swing! The first two:

 * https://github.com/serilog/serilog-sinks-applicationinsights
 * https://github.com/serilog/serilog-sinks-mongodb

I think the strategy is going to require some compromises, doing this as a "big bang" is not much fun. Moving sinks one-by-one as they need to rev seems to work well.

Here's the basic requirements for each move:

 - [ ] Create a new repository _seriog-sinks-sinkname_ in the https://github.com/serilog organisation (readme, VIsual Studio .gitignore, Apache 2.0 license)
 - [ ] Turn off the issue tracker in the new repository
 - [ ] Create a new team _Owners (Sink Name)_ and add the original contributor/interested parties
 - [ ] Grant the team write access to the repo
 - [ ] Remove the sink from the https://github.com/serilog/serilog repository and build script
 - [ ] Push/merge to master, making sure the sink is no longer published from the old location
 - [ ] Copy the project layout and build script from one of the existing independent repositories
 - [ ] Move the source/tests for the sink there
 - [ ] `Install-Package Serilog` and remove the dependencies from the NUSPEC file unless it's one of the customised ones (let the versions be picked up from `packages.config`)
 - [ ] Push!
 - [ ] Create a project in the Serilog organisation in _AppVeyor_
[...]

PR #356, opened 2015-01-17
Serilog.Extras.Timing improvements and extensions

mivano:
I extended the timing library with a health check and a meter measure. This is based on the work of metrics.net (which is a java port). The already existed gauge, timer and counter have been extended and can be configured with more options. Like changing the messages or adding additional properties.
I also removed the sealed keyword so they can be reused. 

Next to that, I added a test project and started test cases for the library. 

I did the development on the mac with Xamarin, so I hope this did not mess up the project in any way. There is an additional test project, all the other changes are in the Serilog.Extras.Timing project.


src/Serilog.Extras.Timing/Extras/Timing/HealthCheckResult.cs
```diff
@@ -0,0 +1,96 @@
...
+using System;
+using Serilog.Events;
+
+namespace Serilog.Extras.Timing
+{
+
+	/// <summary>
+	/// The result data of a health check.
+	/// </summary>
+	public class HealthCheckResult{
```
nblumhardt: Nitpick, but to avoid a bunch of churn/inconsistency when folks edit this using VS's conventions, can we place opening `{` on its own line? There's a _ReSharper_ settings file with the solution, if you like I can run an automatic reformat pre-merge?

Serilog.sln
```diff
@@ -1,25 +1,12 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
```
nblumhardt: Yikes - SLN file merge hell :*( ... any way we can add the project using VS so the solution file remains a bit less churny? I'm mostly concerned that a big SLN file chance will make the other PRs coming in harder to merge successfully.

src/Serilog.Extras.Timing/Extras/Timing/HealthCheckResult.cs
mivano: It is the Xamarin Studio coding convention. Starting to use more and more the mac for development, but then you get this. I will do a reformat on my Windows machine with VS.

src/Serilog.Extras.Timing/Extras/Timing/CounterMeasure.cs
```diff
@@ -24,7 +27,7 @@ sealed class CounterMeasure : ICounterMeasure
         readonly LogEventLevel _level;
         readonly string _template;
         readonly bool _directWrite;
-        static AtomicLong _value;
```
nblumhardt: Aha! :-) Sorry I missed this aspect of the PR on first glance!

Serilog.sln
mivano: With all those magic guids...yeah I can see that this is not handy. Will do a change in VS instead of Xamarin.

src/Serilog.Extras.Timing/Extras/Timing/CounterMeasure.cs
```diff
@@ -46,8 +49,10 @@ public CounterMeasure(ILogger logger, string name, string counts, LogEventLevel
             _value = new AtomicLong();
         }
 
-
-        public void Increment()
+		/// <summary>
+		/// Increments the counter.
+		/// </summary>
+        public virtual void Increment()
```
nblumhardt: Haven't read ahead; but to support higher perf scenarios the original would have been getting close to an order of magnitude faster - worth considering?

nblumhardt:
This is a really, really awesome project; exciting to see the update - sorry about losing track of it for nearly a month!

After sorting out this PR, what do you think about moving this to its own top-level repo e.g. serilog/serilog-extras-timing, or maybe even serilog/serilog-metrics (and reincarnate as _Serilog.Metrics_)?

Tried to add what feedback I can, but you know this one the best so let me know how you want to tackle it. Cheers!

mivano:
Might not be a bad idea to move it out of the big solution and into its own repository. Serilog-metrics looks to me a better name as it contains more than just some timing functions. Is this something you can set up and allow me to push the code to after doing some namespace changes?

nblumhardt:
Yes, I can do that. @khellang @merbla @SimonCropp any thoughts on kicking off another "top-level" project like this? Any conventions we want to set and/or avoid? Cheers!

khellang:
I think it sounds like a good plan. Maybe @mivano or someone else wants to take some ownership of it as well? :smile:

mivano:
Of course, then I take ownership of it.

nblumhardt:
@SimonCropp and I spoke briefly yesterday about the need for some general principles to help us decide when to create a new top-level project under /serilog vs. when to try spinning off more independent subprojects. I think there's a good discussion to have, following on from what we've started with #344.

There are some cool opportunities open to us as well - unfortunately I've been a bit slow in writing it up :-) I'll try to follow up here in the next day or two...

nblumhardt:
With this now moved to https://github.com/serilog-metrics/serilog-metrics I guess the remaining task here is to publish a "redirect" package?

nblumhardt:
Done! https://www.nuget.org/packages/Serilog.Extras.Timing/2.0.0 now depends on `SerilogMetrics`. It's listed as a 2.0 so "breaking" - hopefully the namespace change isn't too much of a headache. Cheers!

mivano:
Great, thanks.

Issue #357, opened 2015-01-17
PeriodicBatchingSink should flush on unhandled exception


konste:
PeriodicBatchingSink class subscribes to two events:
     AppDomain.CurrentDomain.DomainUnload += OnAppDomainUnloading;
     AppDomain.CurrentDomain.ProcessExit += OnAppDomainUnloading;

DomainUnload event is never raised in the default application domain.
ProcessExit event is only raised when the process exits normally, but when the process exits because of unhandled exception (which is the most interesting case from the log perspective) ProcessExit event is NOT raised. UnhandledException event is raised instead.

IMHO it would be very helpful if PeriodicBatchingSink subscribes to all three events.

Thank you!
Konstantin

nblumhardt:
Thanks for the suggestion. It's pretty common to subscribe to `UnhandledException` and set `e.Handled = true` as a "global catch block" in some kinds of apps. This would be incompatible with the way the sink is flushed currently, which is to join with the worker thread (and any synchronous flush in this method could cause trouble for some apps).

I don't think we can take this one on as a bug; it's an interesting improvement that can be made, but needs a bit of design work. Happy to discuss the ins-and-outs if you decide to take a shot at it, but will close this ticket for now.

Thanks!

konste:
I researched it a little further and it seems e.Handled = true you mentioned is something specific for WPF and Silverlight and relevant to Application.DispatchUnhandledException event.

For AppDomain.UnhandledException event there is no way to "catch" it and prevent process from exiting within 2 seconds if IsTerminating event handler parameter is true.

As far as i can tell, AppDomain.UnhandledException is what you get INSTEAD of ProcessExit in case of abnormal termination. You cannot guaranty that the sink would be flushed in this case, but at least it is better to try, than not to try and guaranty losing the most important pre-crash log events.

< sent from mobile device >

On Jan 17, 2015, at 2:33 AM, Nicholas Blumhardt <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for the suggestion. It's pretty common to subscribe to UnhandledException and set e.Handled = true as a "global catch block" in some kinds of apps. This would be incompatible with the way the sink is flushed currently, which is to join with the worker thread (and any synchronous flush in this method could cause trouble for some apps).

I don't think we can take this one on as a bug; it's an interesting improvement that can be made, but needs a bit of design work. Happy to discuss the ins-and-outs if you decide to take a shot at it, but will close this ticket for now.

Thanks!

-

[...]

PR #366, opened 2015-01-26
#357 PeriodicBatchingSink now flushes on unhandled exceptions

chuwik:
Context for this in issue #357 

Not very happy with the name "ComeFromInvalidAppDomainContext" for the method, but couldn't think of a better one!

Also tried to create a unit test, but it's to do so when the AppDomain and unhandled exceptions are at play. Any suggestions for how to do this?

### DIFF
--- a/src/Serilog.FullNetFx/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
+++ b/src/Serilog.FullNetFx/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
@@ -60,11 +60,21 @@ namespace Serilog.Sinks.PeriodicBatching
 
             AppDomain.CurrentDomain.DomainUnload += OnAppDomainUnloading;
             AppDomain.CurrentDomain.ProcessExit += OnAppDomainUnloading;
+            AppDomain.CurrentDomain.UnhandledException += OnAppDomainUnloading;
         }
 
         void OnAppDomainUnloading(object sender, EventArgs args)
         {
+            if (ComeFromInvalidAppDomainContext(args)) 
+                return;
+
             CloseAndFlush();
+        }
+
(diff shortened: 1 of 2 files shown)


nblumhardt:
Looks good. I think rather than give a name to that method, just inlining it would be as clear or clearer - how does that sound? Cheers!

chuwik:
Sounds good, inlined it and fixed the condition: now it does "if it is of type UnhandledExceptionEventArgs and it's not terminating => don't flush"; the first part of that conditional was wrong before and would trigger the return for EventArgs coming from other places.

src/Serilog.FullNetFx/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
```diff
@@ -79,6 +84,7 @@ void CloseAndFlush()
 
             AppDomain.CurrentDomain.DomainUnload -= OnAppDomainUnloading;
             AppDomain.CurrentDomain.ProcessExit -= OnAppDomainUnloading;
+            AppDomain.CurrentDomain.UnhandledException += OnAppDomainUnloading;
```
nblumhardt: Should be `-=`, not `+=` here

nblumhardt:
Great, looking good! Just one line comment noted above. There's also a corresponding .NET 4.0 implementation in _Serilog-net40.sln_ that needs to be in sync. Nothing else from me! :-)

src/Serilog.FullNetFx/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
chuwik: Right! Silly me!

chuwik:
All fixed now! Thanks!

PR #369, opened 2015-01-28
[Elasticsearch Sink] Log message should be named "message" instead of "r...

jensbengtsson:
...enderedMessage"

This change was made in the other elasticsearch json formatter, but not in the .net 4.0 one. So this should be changed here as well.


nblumhardt:
Thanks very much

Issue #30, opened 2013-05-02
Natively support JSON.NET dynamic objects (JObject)


ChristianWeyer:
On Twitter you mentioned that it should be fairly straightforward to implement via IDestructuringPolicy
;)

mcintyre321:
@kkozmic Any chance I could take a look at your ```IDestructuringPolicy ``` (the one that works with the wrapper)?

kkozmic:
Sorry, I don't have that code with me anymore.

nblumhardt:
#371 echoes this one, I think it's one for us to reopen, but might be a "2.0"

nblumhardt:
Build 1.4.168 removes the limitation that made this awkward; would love to see a _Serilog.Extras.JObject_ or similar that takes advantage of it :-)

Thanks @Jaben for nailing this!

nblumhardt:
https://github.com/destructurama/json-net

mcintyre321:
Nice!

CumpsD:
@nblumhardt that's awesome, thanks!

PR #374, opened 2015-02-03
Non-breaking improvement for issue #30

Jaben:
The IDestructuringPolicies are now run before IEnumerable cast -- allowing an opportunity for policies to support IEnumerable values. Added IsValueTypeDictionary function to simplify the lookup for the dictionary conversion code.


nblumhardt:
Great - high time we did this, kudos!

Might take me a few more days to get into gear and fully think though why I didn't do this in the first place, but at first glance it's looking good.

Jaben:
@nblumhardt I had the same thought... Based on the existing IDeconstructPolicy implementations -- I didn't see anything that would break. But it may break (and fix) various configuration implementations now that all objects, even IEnumerable ones, would being to be passed into their code.

PR #376, opened 2015-02-05
Wait for Loggly client to submit the log event.

tetious:
In the current version of loggly-csharp, client.Log returns a Task, so it needs to be awaited or .Wait()ed otherwise the log message will not succeed if the calling thread is aborted quickly.

https://github.com/neutmute/loggly-csharp/commit/e38b247caa5ce7a1a6c43608635fe4e9a82c9dd9

### DIFF
--- a/src/Serilog.Sinks.Loggly/LoggerConfigurationLogglyExtensions.cs
+++ b/src/Serilog.Sinks.Loggly/LoggerConfigurationLogglyExtensions.cs
@@ -30,17 +30,23 @@ namespace Serilog
         /// <param name="loggerConfiguration">The logger configuration.</param>
         /// <param name="restrictedToMinimumLevel">The minimum log event level required in order to write an event to the sink.</param>
         /// <param name="formatProvider">Supplies culture-specific formatting information, or null.</param>
+        /// <param name="batchPostingLimit">The maximum number of events to post in a single batch.</param>
+        /// <param name="period">The time to wait between checking for event batches.</param>
         /// <returns>Logger configuration, allowing configuration to continue.</returns>
         /// <exception cref="ArgumentNullException">A required parameter is null.</exception>
         public static LoggerConfiguration Loggly(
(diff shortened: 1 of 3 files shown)


nblumhardt:
Yikes, nicely spotted; thanks. The performance impact of the `Wait()` here could turn out to be pretty extreme for some current users - I wonder if we're better of converting the sink to a `PeriodicBatchingSink` which would then move the work to a background thread? @neutmute?

tetious:
Ahh! I was wondering if there was something like PeriodicBatchingSink, but am new to Serilog and didn't spot it. Given the impact of the Wait() in our experimentation, this is likely a good idea. 
It doesn't seem to arduous to convert. I will try to take a stab at it shortly.

tetious:
I've converted it to PeriodicBatchingSink.

This seems to work in our tests. Comments/concerns are welcome.

neutmute:
Certainly seems a better solution than wait(). 
I made the loggly-csharp call async so it wouldn't block while sending the data across the internet. 
it should not be async and spin up its own thread to send?

nblumhardt:
Ah I see - doesn't look like there's a batch interface to get events across to Loggly. I think this is an improvement on the un-awaited version.

Since it's a breaking change I think we should rev the major version number and move the sink out to a standalone build as we've been incrementally doing with other sinks now via #344 - I'll merge the PR and set up the new repo in one shot.

@tetious @neutmute would either/both of you be interested in being the nominal owners of the Loggly sink going forward? Cheers!

neutmute:
@nblumhardt I'm not actually using serilog in anger in any of my projects yet so will bow out of the race for now :)

nblumhardt:
@neutmute :) :+1:

tetious:
I'm certainly happy to help out with patches and such, but I don't know that I have the time or organization skills to manage anything except my sock drawer. ;)

PR #378, opened 2015-02-07
Provides a System.Diagnostics.TraceListener implementation that outputs to Serilog

vossad01:
Provides a [`System.Diagnostics.TraceListener`](https://msdn.microsoft.com/en-us/library/system.diagnostics.tracelistener%28v=vs.110%29.aspx) implementation that outputs to Serilog.  As far as existing Serilog functionality this is probably most similar to `Serilog.Extras.MSOwin.LoggerFactory`

This came about because, near as I could tell, WCF only logs to Microsoft's tracing infrastructure so if you want consolidated logging `TraceListener` is the only extension point.

This does not respect the `TraceListener.Filter` member, I am not sure if contractually it needs to.  I noted that [NLog's TraceListener](https://github.com/NLog/NLog/blob/00d53bc3b234917517e2fdc1735f8b4af2d5d8aa/src/NLog/NLogTraceListener.cs) also does respect `Filter` so it seems there is some precedent for this.

I have never authored a .nuspec; nor, to be honest, am I especially familiar with either Serilog or System.Diagnostics, so I mostly just tried to follow conventions and hopefully did all right.
___
To Do
- [x] Reduce string concatenation

### DIFF
--- a/src/Serilog.Extras.TraceListener/Extras/TraceListener/SerilogTraceListener.cs
+++ b/src/Serilog.Extras.TraceListener/Extras/TraceListener/SerilogTraceListener.cs
@@ -1,0 +1,195 @@
+﻿// Copyright 2015 Serilog Contributors
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using Serilog.Core.Pipeline;
+using Serilog.Events;
+
+namespace Serilog.Extras.TraceListener
+{
+    /// <summary>
+    ///     TraceListener implementation that directs all output to Serilog.
+    /// </summary>
+    public class SerilogTraceListener : System.Diagnostics.TraceListener
+    {
+        const LogEventLevel FailLevel = LogEventLevel.Fatal;
+        const string MessageMessageTemplate = "{TraceMessage:l}";
+        const string FailMessageTemplate = "Fail: " + MessageMessageTemplate;
+        const LogEventLevel DefaultLogLevel = LogEventLevel.Debug;
+        ILogger logger;
+
+        /// <summary>
+        ///     Creates a SerilogTraceListener that uses the logger from `Serilog.Log`
+        /// </summary>
+        /// <remarks>
+        ///     This is needed because TraceListeners are often configured through XML
+        ///     where there would be no opportunity for constructor injection
+        /// </remarks>
+        public SerilogTraceListener() : this(Log.Logger)
+        {
+        }
+
+        /// <summary>
+        ///     Creates a SerilogTraceListener that uses the specified logger
+        /// </summary>
+        public SerilogTraceListener(ILogger logger)
+        {
+            this.logger = logger.ForContext<SerilogTraceListener>();
+        }
+
+        public override bool IsThreadSafe
+        {
+            get { return true; }
+        }
+
+        public override void Write(string message)
+        {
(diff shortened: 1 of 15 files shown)


nblumhardt:
Thanks Adam, this looks great! The only thing I can spot that might need a tweak is the folder location of the project - it needs to live under `/src/` rather than in the root.

It might also be better to name:

```csharp
const string MessageMessageTemplate = "{message:l}"
```

as:

```csharp
const string MessageMessageTemplate = "{TraceMessage:l}"
```

since "message" is a bit overloaded throughout Serilog.

Cheers!

vossad01:
Thanks for taking a look!  Those items should be fixed up now.

src/Serilog.Extras.TraceListener/Extras/TraceListener/SerilogTraceListener.cs
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        public override void Write(string message)
+        {
+            logger.Write(DefaultLogLevel, MessageMessageTemplate, message);
+        }
+
+        public override void Write(string message, string category)
+        {
+            logger.Write(DefaultLogLevel, "{Category:l}: " + MessageMessageTemplate, category, message);
```
khellang: Just a general concern; how much do we care about concatenations like this? I'm sure in this specific case, by the time this runs, the compiler has optimized it to `"{Category:l}: {TraceMessage:l}"` (since it's a `const`), but still, I'm sure we have other concatenations like this lying around which causes allocations for each logging statement. It might seem like nothing, but it all adds up across big applications.

vossad01: If no one cares about the repetition I am happy to change it.  To be honest, I initially did write out all the strings and once I had everything working I refactored it to how you see it now to avoid any possible concerns with duplication.

I personally had less concern with this one (because it is `const` like you mention), than `WriteEvent` a little further down where the message string is built conditionally based on the parameters.

I wasn't sure if doing things for performance reasons would be considered "premature optimization"

nblumhardt: Const concatenations aside (I'm pretty sure the compiler will take care of those) it's probably best to avoid the allocations. The core of Serilog works pretty hard to avoid them - the sinks and "extras" don't so much, but a bit of optimization is always welcome :-)

vossad01:
To Do
- [ ] Reduce string concatenation

nblumhardt:
Wonderful, thanks. We'll try getting this into the next release. Cheers!

vossad01:
Great!  Thanks!

It is nice that AppVeyor lets artifacts be publicly downloaded.  I pulled the nupkgs and have been using it this week; configured with `activityLevel=Warning` this seems to capture the WCF exceptions I wanted it for.

nblumhardt:
Hi again,

Timing might be a bit unfortunate here, actually. We've had a lot of discussion recently about keeping the core Serilog project manageable, and maintainable without a hugely diverse skillset. What we ended up settling on was:

 * The serilog org will still accept sinks, but these will be in standalone repos (like https://github.com/serilog/serilog-sinks-mongodb)
 * The 'extras' projects will either be absorbed into Serilog itself (e.g. we'll do this with _Serilog.Extras.AppSettings_) or more likely, move them to separate organisations and encourage them to take an independent identity (https://github.com/serilog-metrics/serilog-metrics is the poster-child for this approach right now)

You can follow some discussion in #344.

Would you be interested in creating and maintaining a top-level `serilog-trace-listener` org and distribute this as e.g. _SerilogTraceListener_ (note the shift from a Serilog sub-namespace)? If so, we'd be delighted to help out.

Hope this makes sense, please let me know what you think! :-)

vossad01:
Thanks for the heads up.  There are a couple technical issues presented by this PR concerning the split:

 * Testing infrastructure provided by `Serilog.Tests` is utilized in the tests.  Would you be thinking of publishing a NuGet package of that?
 * The internal type `SilentLogger` from `Serilog.Core.Pipeline` is utilized.  Would the preference be for the new package to provide its own implementation, or should something be done to make it accessible?

I have one other concern that I have e-mailed you about.

nblumhardt:
Hi Adam - sorry about the delay on the ticket, followed up on our email thread and forgot about the open ticket here too.

`SilentLogger` should probably be duplicated, it's not useful enough to warrant exposure on the API IMHO.

A public test support library would be nice - it's more a question of time on that one. Duplicating some of the infrastructure is probably unavoidable in the short term.

Closing this one for the time being but feel free to jump back in if you need help with anything. Cheers!

PR #379, opened 2015-02-07
Corrects a minor spelling mistake in the description of the MSOwin nuget package

vossad01:
Corrects a minor spelling mistake in the description of the MSOwin nuget package.

I probably would have just ignored this since it is minor were it not for the fact this is the description that gets published on nuget.org and appears in package search results.


nblumhardt:
Thanks!

PR #382, opened 2015-02-10
Fixes support for using multiple CounterMeasure

steven-dawkins:
Using Serilog.Extras.Timing to create two counters they both share a value which is incremented together.

Code:
```csharp
var bookCounter = logger.CountOperation("books", "created books", false, LogEventLevel.Debug);
var authorCounter = logger.CountOperation("authors", "created authors", false, LogEventLevel.Debug);

bookCounter.Write();
authorCounter.Increment();
bookCounter.Write();
```

Output:

```
2015-02-09 22:07:56 [Debug] "books" count = 0 created books
2015-02-09 22:08:13 [Debug] "books" count = 1 created books
```

The fix is a one liner removing the static property on CounterMeasure _value

In the second commit I created a test project for Serilog.Extras.Timing with a test covering this case.


nblumhardt:
Yikes, nice catch - merging, but @mivano please yell if I've missed something obvious in this scenario. Cheers!

mivano:
Aih nice one indeed. BTW @nblumhardt I still have a large PR (https://github.com/serilog/serilog/pull/356) open for other improvements. Any change of having a look at that one? I might need to merge that one again.

Issue #386, opened 2015-02-14
Extras.AppSettings should expand environment variables when reading the configuration


CaioProiete:
When using Serilog in a Windows Client application, it is very common to want to save the log in the user's AppData folder, where we know we can write to.

I'd like to be able to do something like this.

    <appSettings>
      <add key="serilog:write-to:RollingFile.pathFormat" value="%LOCALAPPDATA%\MyApp\Log-{Date}.txt" />
    </appSettings>

And have the `%LOCALAPPDATA%` environment variable expanded.

---

I believe a workaround would be to add a new property via `Enrich`, but then I'd have this property in every log entry, which is not what I want... A custom enricher could help control that, but then it would still be called on every log call, which is not what I want either.

kberridge:
Just wanted to leave a note that I was trying to include %LOCALAPPDATA% in the pathFormat in the C# .WriteTo.RollingFile(...) call, which doesn't work!  I finally realized this only works from appSettings.  This works out just fine for me but I wanted to point out that I found the inconsistency surprising.

CaioProiete:
Interesting feedback @kberridge. Thanks.

I personally wouldn't bother using env variable expansion if I'm configuring the path via C# (i.e. I'd just use the Environment class to expand as I assemble the path), but I can see the appeal of having this working in C# too just for consistency, so :+1: .

PR #387, opened 2015-02-14
#386 Expand environment variables on Extras.AppSettings values

CaioProiete:
Details on issue #386.


nblumhardt:
Fantastic, thanks Caio!

PR #388, opened 2015-02-15
Newest elmah.io package

ThomasArdal:
Updated Serilog.Sinks.ElmahIO.nuspec to point to the newest stable package of elmah.io.


nblumhardt:
Thanks Thomas!

PR #389, opened 2015-02-17
Bug es uri

mookid8000:
Corrected spelling error in default localhost URI
Made the fact that the ES sink options are optional more obvious by making it an optional parameter

### DIFF
--- a/src/Serilog.Sinks.ElasticSearch/LoggerConfigurationElasticSearchExtensions.cs
+++ b/src/Serilog.Sinks.ElasticSearch/LoggerConfigurationElasticSearchExtensions.cs
@@ -175,7 +175,7 @@ namespace Serilog
         /// <param name="loggerSinkConfiguration"></param>
 	    /// <param name="options">Provides options specific to the Elasticsearch sink</param>
         /// <returns></returns>
-        public static LoggerConfiguration Elasticsearch(this LoggerSinkConfiguration loggerSinkConfiguration, ElasticsearchSinkOptions options)
+        public static LoggerConfiguration Elasticsearch(this LoggerSinkConfiguration loggerSinkConfiguration, ElasticsearchSinkOptions options = null)
         {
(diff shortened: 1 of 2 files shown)


khellang:
LGTM. Just one thing; how do we feel about optional parameters vs overloads? It can lead to [some versioning fubar](http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/), so it might not be the best for public methods.

nblumhardt:
Thanks!

@khellang yah, none of the options are particularly good. I went with optional parameters originally throughout Serilog because, despite versioning being trickier, overloads create a confusing API over the long-term.

The strategy for working around the versioning issue so far has used overloads temporarily. E.g. from Phil's example:

```csharp
// v1
  public static void Redirect(string url, string protocol = "http");
```

We then remove the defaults, obsolete the old method, rename its defaulted args and add a new one in V2:

```csharp
// this was v1
  [Obsolete(...)]
  public static void Redirect(string url, string obsoleteProtocol);

// v2
public static void Redirect(string url, string protocol = "http",   bool permanent = false);
```

It's imperfect, but in the usage pattern we encourage most callers will then recompile against v2 while still running against v1.

The only source breakage is:

```csharp
Foo.Redirect(a, b);
```

...for which there will be an obsoletion warning, and the user needs:

```csharp
Foo.Redirect(a, protocol: b);
```

...where the arg name disambiguates from the obsoleted method.

Note we only do this in the configuration APIs, never the logging API, so the number of affected LOC is generally only a few.

In the long run it's better than having to have the "too many confusing overload permutations" discussion and just obsoleting a bunch of them anyway.

Hope this makes some sense :-) (Any ideas for fixing the one wart on the scheme?)

khellang:
Makes perfect sense :smile: Sounds like a good approach. Regarding source breakage, I think that problem will be less of a big deal once VS 15 comes around with the ability to ship analyzers in NuGet packages. They'll make it even easier to quickly move people away from obsolete APIs by providing "fix all" code fixes to correct the API usage :smile:

PR #390, opened 2015-02-18
Added durability option to the ElasticSearch sink...

mookid8000:
...similar to how the Seq sink achieves reliable log forwarding.


mookid8000:
sorry - forgot to base work off of dev!

PR #397, opened 2015-03-01
Added Enrichers for Azure Environment

jbattermann:
Added a new Project that provides Azure RoleId, -Name and DeploymentId Enrichers. Relies on the Azure SDK (the current one is 2.5.0) as the underlying 'Microsoft.WindowsAzure.ServiceRuntime' assembly is not available as NuGet package.


nblumhardt:
Howdy! Nice PR, thanks.

Unfortunately it's come at a time when we're starting to scale back what we accept in the core `serilog` project/repository, and while we're still encouraging sinks to be submitted (albeit as separate repositories) we're aiming to move out all of the "extras". You can pick up some of the discussion via #344 (which I'd guess you've already seen via your Application Insights work).

@mivao has kicked off our flagship "community" project as https://github.com/serilog-metrics/serilog-metrics published to https://nuget.org/packages/serilogmetrics. This is the model we'd like to use for "extras" type things going forwards.

If you would consider self-publishing this as e.g. _SerilogAzureEnrichers_ under a similar model we'd love to help out. There's even a new "community" package logo published here: http://serilog.net/images/serilog-community-nuget.png

PR #401, opened 2015-03-06
Support IEnumerable<Uri> in AppSettingsReader

mdean:
This change adds support for IEnumerable<Uri> in the AppSettingsReader, allowing multiple Uri settings separated by a space.  We are using app settings with Elasticsearch and need to specify multiple nodes in the setting.  Please let me know if there is anything else I need to do.

### DIFF
--- a/src/Serilog.Extras.AppSettings/Extras/AppSettings/PrefixedAppSettingsReader.cs
+++ b/src/Serilog.Extras.AppSettings/Extras/AppSettings/PrefixedAppSettingsReader.cs
@@ -134,7 +134,8 @@ namespace Serilog.Extras.AppSettings
             var extendedTypeConversions = new Dictionary<Type, Func<string, object>>
             {
                 { typeof(Uri), s => new Uri(s) },
-                { typeof(TimeSpan), s => TimeSpan.Parse(s) }
+                { typeof(TimeSpan), s => TimeSpan.Parse(s) },
+                { typeof(IEnumerable<Uri>), s => s.Split(new [] { ' ' }, StringSplitOptions.RemoveEmptyEntries).Select(u => new Uri(u)) }
             };
 
             var convertor = extendedTypeConversions
--- a/test/Serilog.Extras.AppSettings.Tests/PrefixedAppSettingsReaderTests.cs
+++ b/test/Serilog.Extras.AppSettings.Tests/PrefixedAppSettingsReaderTests.cs
@@ -1,4 +1,8 @@
-﻿using System.Collections.Specialized;
+﻿using System;
+using System.Collections;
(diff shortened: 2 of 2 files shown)


nblumhardt:
Hi, thanks for the PR!

There are two reservations I have right now about this. The first is trivial - space is not always a typical separator in lists, so maybe considering CSV would be a better option?

The second is that it seems very case-specific right now; just supporting `IEnumerable<Uri>` and not, say, `IEnumerable<string>`, `int` etc. seems patchy.

I think this one is a bit of a recurring problem, it'd be great to figure out a more comprehensive approach we could rely on and extend to things like complex objects (the user credential object required by the SMTP sink came up recently).

Any thoughts?

mdean:
Thanks for the feedback!

Regarding the use of space, it wasn't my first choice, but my intent was to pick a delimiter that "should" never appear in a Uri.  That would be easy to change if there is a more suitable delimiter.

I see your point that this is pretty specific, and I was instinctively following the pattern implemented by the single Uri without considering other needs as I was unaware of them.  I think this is an interesting problem you bring up, mostly due to the potentially numerous requirements to parse and initialize the various types correctly.

mivano:
One thing @Mpdreamz and I discussed was using JSON instead. So have one method extension with a complex type (which might be having an interface like ISerilogSinkSettings) and have an appSettings key like

```xml
<add key="serilog:write-to:RollingFile:settings" value="{ pathFormat: '../logs/{{Date}}.json', retainedFileCountLimit: 10}" />
```

It wont solve the issue that you might want to initialize a different serializer like in the ES sink. But that might be fixed by also adding a SerializerType (string) property. 

This way you can still use the appSettings to configure Serilog from deployment tools like Octopus and also supply the more complex types. Alternatively; pass the settings string to the Sinks (implementing a certain interface) instead so they are responsible for interpreting this and setting it up.

mdean:
Using JSON makes perfect sense to take this further.  Should I close this pull request?

nblumhardt:
JSON does sound like a good way to go.

Perhaps the algorithm could be something like:

1. If the parameter (in the extension method) is of a known simple type (e.g. `string`, `int`, `DateTime`, ... perhaps `Uri`) then assume the `value='...'` attribute contents are literal values as we do today
2. Otherwise, assume the `value='...'` contents are JSON, and attempt to deserialize an instance of the parameter type from it

The JSON (de-)serializer would share formatting logic for known simple types like `DateTime` so that the syntax is consistent with (1).

I think figuring out this feature is an important part of stabilizing `<appSettings>` support for inclusion in the core. I'm going to create a new ticket with this info and the other requirements I have in mind :-)

Issue #395, opened 2015-02-25
Uppercasing of properties in Output Template does not work.


thinktainer:
When trying to uppercase the logging level, the whole message gets swallowed.

This will reproduce the behaviour:

    using System;
    using Serilog;

    namespace ConsoleApplication1
    {
        class Program
        {
            static void Main(string[] args)
            {
                var logger =
                    new LoggerConfiguration().WriteTo.Console(outputTemplate: "{Timestamp} {Level:u} {Message}").CreateLogger();
                logger.Information("Hello, serilog");
                Console.ReadKey();
            }
        }
    }

nblumhardt:
Interesting - must be a regression introduced when output templates were separated from message templates; thanks for the report!

the-dwyer:
I have submitted a pull request for a possible fix to this issue.

nblumhardt:
As per the discussion in #409, `Level` is not a string and therefore uppercase formatting does not work for it.

Uppercasing within message templates isn't currently supported - not sure we'll go down this path.

PR #409, opened 2015-03-19
Uppercasing of properties in Output Template does not work. #395

the-dwyer:
Hi,

I added three unit tests for this bug as well as a possible fix.

Let me know what you think.

Thanks,

Michael

### DIFF
--- a/src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
+++ b/src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
@@ -84,9 +84,10 @@ namespace Serilog.Formatting.Display
                 var sv = propertyValue as ScalarValue;
                 if (sv != null)
                 {
+                    var value = sv.Value == null ? "null" : sv.Value.ToString();
                     var overridden = new Dictionary<string, LogEventPropertyValue>
                     {
-                        { pt.PropertyName, new LiteralStringValue(sv.Value.ToString()) }
+                        { pt.PropertyName, new LiteralStringValue(value) }
                     };
 
                     token.Render(overridden, output, _formatProvider);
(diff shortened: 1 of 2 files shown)


src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
```diff
@@ -82,11 +82,11 @@ public void Format(LogEvent logEvent, TextWriter output)
                 // rendering and support some additional formats: 'u' for uppercase
                 // and 'w' for lowercase.
                 var sv = propertyValue as ScalarValue;
-                if (sv != null && sv.Value is string)
+                if (sv != null)
                 {
                     var overridden = new Dictionary<string, LogEventPropertyValue>
                     {
-                        { pt.PropertyName, new LiteralStringValue((string) sv.Value) }
+                        { pt.PropertyName, new LiteralStringValue(sv.Value.ToString()) }
```
nblumhardt: Is it possible for `Value` to be null here?

nblumhardt:
Thanks Michael! Added one note.

the-dwyer:
Hi,

I added a unit test for the case when the value of the ScalarValue is null as well as the code to make the unit test pass.  I looked in the Render method of ScalarValue and it uses "null" when the value is null so that is what I used as well.

Let me know what you think.

Thanks,

Michael

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
```diff
@@ -82,11 +82,12 @@ public void Format(LogEvent logEvent, TextWriter output)
                 // rendering and support some additional formats: 'u' for uppercase
                 // and 'w' for lowercase.
                 var sv = propertyValue as ScalarValue;
-                if (sv != null && sv.Value is string)
```
nblumhardt: I'm not sure I completely understand why this change works - previously we only used this rendering path for strings, but in the new version we'll end up rendering everything - integers, datetimes and so-on, down this path. Am I misunderstanding here? Thanks!

nblumhardt:
Aaah- sorry, I just re-read https://github.com/serilog/serilog/issues/395 and realised what's going on here.

`:u` and `:w` are not valid on the level because it is an `enum`, not a `string`, and those specifiers only work for strings.

Can we please pause this PR and go back to the issue? I believe this case is already covered but will need to take another look at it.

the-dwyer:
You bet!  I'll close this one and if I have another possible solution, I'll send in another pull request.

Michael

Issue #428, opened 2015-04-16
RollingFileSink to roll over on FileSize


thinktainer:
In our application it would be desirable to have the RollingFileSink as a reliable backup mechanism for when network logging fails. We would prefer to have the log file roll over when it reaches the file size limit rather than truncating (e.g.: log-{Date}-001.txt log-{Date}-002.txt). We are aware that this would make the disk usage unbounded, and would prefer to live with this tradeoff.
I have looked at the source code, and am curious if such a behaviour was ever planned when looking at https://github.com/serilog/serilog/blob/dev/src/Serilog.FullNetFx/Sinks/RollingFile/TemplatedPathRoller.cs#L71?
Is this a feature you would be open to consider including as a configurable option?

nblumhardt:
Hi @thinktainer - thanks for getting in touch!

Have you looked at simply passing `null` for the `fileSizeLimitBytes` argument of `WriteTo.RollingFile()`? This will permit unbounded disk usage, but not truncate files at any size.

https://github.com/serilog/serilog/blob/dev/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs#L179

Roll-on-size is something I'd be interested to look into, but I'm not certain the complexity tradeoff would be a great one right now (unsure how much this would churn). If you end up having a chance to investigate any insights here would be valuable.

thinktainer:
Hey Nicholas, thank you for getting back to me so fast.
We have considered growing the files much larger too, the problem with that is that the tooling for inspecting them at that size is less then ideal for our scenario; most of the times we need to log onto a deployed instance and inspect files there, so we would either need to split them with powershell or similar, or use a buffer based editor like vim.
I'm going to spike an implementation of this our end, and will get back in touch as soon as I have something a bit more tangible.
Another thing I have noticed is that the comment at: https://github.com/serilog/serilog/blob/dev/src/Serilog.FullNetFx/Sinks/RollingFile/RollingFileSink.cs#L49 seems to describe behaviour that I seem to not be able to get to work here (the files never increment the inc part). Can you elaborate what the intended usage for this behaviour is? Would you expect the files to roll over on sink initialisation in case they have grown too big, or depending on another condition?

nblumhardt:
[...]

PR #434, opened 2015-04-24
size rolling log file

thinktainer:
## This is a sketch of how i'd imagine the new behaviour to work.

- Only size rolling implemented, saving size + date for later, as I feel there needs to be more discussion about desired behaviour
- Doesn't yet follow your advice of factoring this out into a module, at which point the integration points would obviously need to change.

### DIFF
--- a/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions-net40.cs
+++ b/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions-net40.cs
@@ -25,6 +25,7 @@ using Serilog.Formatting.Raw;
 using Serilog.Sinks.DiagnosticTrace;
 using Serilog.Sinks.IOFile;
 using Serilog.Sinks.RollingFile;
+using Serilog.Sinks.RollingFile.DateOnly;
 using Serilog.Sinks.SystemConsole;
 
 namespace Serilog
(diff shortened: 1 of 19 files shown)


thinktainer:
I'm submitting this in the current state in order to hopefully get some feedback about implementation / test coverage, design. I hope I'm not too far off the mark, but want to avoid going on a tangent for too long.

nblumhardt:
Thanks for posting this, it looks like you've taken it a long way towards a solid implementation!

As mentioned in https://github.com/serilog/serilog/issues/428, I think this is a difficult one for us to take in the integrated form- there are a few breaking API (e.g. `class RollingFileSink` renamed)  and behavioural changes (must choose a maximum days retention) that would make this possible only with a major revision, and given the very wide usage of the sink it's uncertain whether we'd want to roll any significant change out without a lot of interest.

What do you think of the separate package option? It would essentially be a clean slate, which might also allow for some improvements in how the sink operates and is factored. We've started pushing more in this direction across the board, e.g. just recently we published a new colored console that's achieved a good number of downloads in its 20 days of existence: http://www.nuget.org/packages/serilog.sinks.literate

Cheers!
Nick

thinktainer:
I'd be totally up for that, would you be happy to set up the infrastructure required?

nblumhardt:
Yes, though the process for this currently is for you to set up and do the work in a repository of your own, then transfer ownership of the repo to `serilog` (and join it's "owners" team there).

Probably good to get some thoughts in from @serilog/core in case I'm out on a limb - folks, is this the best way for us all to go about providing a more sophisticated rolling file sink (e.g. with size rolling policies)?

merbla:
If there is nothing common or reusable in the policies I would say *YAS* (Yet Another Sink).  Perhaps `2.0` is the time for this sort of change to the core.

PR #436, opened 2015-04-30
Explicitly reference assemblies needed for use under DNX

nblumhardt:
I have been able to use the Serilog NUPKG unaltered in both .NET and CoreCLR based projects build with VS2015 RC, with the only extra requirement being some explicit assembly references (the portable surface area Serilog uses is referenced automatically in the project system for .NET 4.5).

This PR just adds those references via the NUSPEC. Still needs some testing before I can say this is complete - starting a new sink as a validation exercise :-)


nblumhardt:
No joy, still getting prompted to reference _System.Runtime_ when installing into a DNX 4.5.1 project. Will let this one bake a while.

Adding the references by hand is enough to get up and running, in case anyone's keen:

```json
    "frameworks": {
        "dnx451": {
            "dependencies": {
                "System.Runtime": ""
            }
        },
```

It may in fact turn out we need to add target-specific dependencies on these as if they're packages, rather than adding them as `<frameworkAssemblies>`??

nblumhardt:
This work's moving to `dev-v2`

PR #442, opened 2015-05-06
Fixes #441 - KeyValuePairSettings.ConvertToType does not work with Nullable<System.TimeSpan>

### LINKED ISSUE #441: KeyValuePairSettings.ConvertToType does not work with Nullable<System.TimeSpan>
It appears that type info should be updated with info from unwrapped type for correct result, or

    KeyValuePairSettings.ConvertToType("00:01:00", typeof(System.TimeSpan?));

throws

    An unhandled exception of type 'System.InvalidCastException' occurred in mscorlib.dll

    Additional information: Invalid cast from "System.String" to "System.TimeSpan".

coding4food:


nblumhardt:
Very nice - thanks for that.

PR #456, opened 2015-05-28
backport KeyValuePairSettings and AppSettings to net40

tsvayer:
KeyValuePairSettings implementation uses Reflection API.
Reflection API has breaking changes in .net45 (http://blogs.msdn.com/b/dotnet/archive/2012/08/28/evolving-the-reflection-api.aspx)
AppSettings depends on KeyValuePairSettings, so it is missing from .net40 build.
This commit is a backport to old Reflection API.
I have tried to share the same KeyValuePairSettings class by implementing Extensions to simulate TypeInfo from .net45, but unfortunately C# does not support Property Extensions.
Thus, KeyValuePairSettings-net40 was introduced.
It is still almost identical except 2 simple Property vs Function calls.


nblumhardt:
Thank you Vitaly, looks great.

Issue #460, opened 2015-06-01
Default configuration level is too high


klettier:
Hi,

When I define a 'restrictedToMinimumLevel' for my sink that is below the 'LoggerConfiguration' minimum level my events are not logged.
I created a simple test that fail because the last call to debug will never be emitted.

```
    [Test]
    public void LoggingShouldWorkWithOnlySinkLogEventLevel()
        {
            var events = new List<LogEvent>();
            var sink = new DelegatingSink(events.Add);

            var log = new LoggerConfiguration()
                .WriteTo.Sink(sink, LogEventLevel.Debug)
                .CreateLogger()
                .ForContext<LoggerTests>();

            log.Verbose("Suppressed");
            log.Information("Emitted");
            log.Debug("Emitted");

            Assert.AreEqual(2, events.Count);
            Assert.That(events.All(evt => evt.RenderMessage() == "Emitted"));
        }
```

nblumhardt:
Thanks for the report and PR; this one's by design however - the default level in Serilog's always been `Information`, and restriction for an individual sink only narrows down the events from there.

With a time machine I think going back and revising this decision, perhaps to set the minimum level as `LevelAlias.Minimum` like you've suggested, might be a good alternative to what we have today. Making this change now would be a big surprise for those dependent on the current default though, and I don't think we'd win a _huge_ amount out of it even via a 2.0/breaking release.

Thanks again, especially for the very nice PR (with test!) - much appreciated and very welcome any time you decide to pick up the keyboard :-) ... Cheers!

PR #461, opened 2015-06-01
Reduce configuration minimum level

klettier:
Issue https://github.com/serilog/serilog/issues/460


nblumhardt:
Discussion in #460.

PR #463, opened 2015-06-02
AppSettings: Update the unit test that checks environment variable expansion is supported

CaioProiete:
The test was being ignored because it required an App.config present with a key, so I made the changes required for the test to run - I have it running locally in TeamCity and it's passing, so I'd expect it to run fine in your CI... We'll see.


nblumhardt:
Great! Thanks Caio!

PR #464, opened 2015-06-02
Expose AppSettings configuration

CaioProiete:
I try to keep my Serilog settings in the App.config as much as possible as it facilitates deployment a lot (i.e. transforms by environment via Octopus), however, I have scenarios where I need to tweak some properties of specific sinks dynamically via code, so I thought I'd expose the properties somehow to have the ability to get into the pipeline and change the dictionary before it gets consumed by the different sinks.

For example, imagine I have the following setting in my App.config:
```
<add key="serilog:write-to:RollingFile.pathFormat" value="C:\logs\myapp-{SomeTag}-{Date}.log" />
```

And I'd like `{SomeTag}` to be replaced by something that is obtained at runtime, at the start of the app, when I'm configuring Serilog.

With this PR, I'm exposing the settings read from the App.config to the developer, so we can have a chance to modify the settings dictionary before Serilog processes it.

```
    .ReadFrom.AppSettings(settings =>
    {
        // Modify the dictionary to do replaces and/or add items and/or remove items etc.
    })
    .CreateLogger();
```

ps: This builds uppon PR #463

### DIFF
--- a/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs
+++ b/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs
@@ -24,7 +24,6 @@ using Serilog.Events;
 using Serilog.Formatting.Display;
 using Serilog.Formatting.Raw;
 using Serilog.Settings.AppSettings;
-using Serilog.Settings.KeyValuePairs;
 using Serilog.Sinks.DiagnosticTrace;
 using Serilog.Sinks.IOFile;
 using Serilog.Sinks.RollingFile;
@@ -266,13 +265,14 @@ namespace Serilog
         /// containing sinks, use <code>serilog:using</code>. To set the level use 
         /// <code>serilog:minimum-level</code>.
         /// </summary>
-        /// <param name="settingConfiguration">Logger setting configuration</param>
+        /// <param name="settingConfiguration">Logger setting configuration.</param>
+        /// <param name="exposeSettings">Action allowing changes to the Serilog settings read from the App.config.</param>
         /// <returns>An object allowing configuration to continue.</returns>
         public static LoggerConfiguration AppSettings(
-            this LoggerSettingsConfiguration settingConfiguration)
+            this LoggerSettingsConfiguration settingConfiguration, Action<Dictionary<string, string>> exposeSettings = null)
         {
             if (settingConfiguration == null) throw new ArgumentNullException("settingConfiguration");
-            return settingConfiguration.Settings(new AppSettingsSettings());
+            return settingConfiguration.Settings(new AppSettingsSettings(exposeSettings));
         }
     }
 }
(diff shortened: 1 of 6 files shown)


nblumhardt:
Thanks for the PR!

I'm still wrapping my head around how the various settings providers will compose; I'm not sure at this stage how the API is set to evolve, so I don't think now's the right time for us to take this one.

(Hoping it's not difficult for you to implement/plug this in as a custom `ILoggerSettings` from your own solution - if it is, let me know, very interested.)

Cheers!

CaioProiete:
No worries. I understand the uncertainty on the settings providers, especially now after the "monster update".

It is def. not "hard" to implement this in a custom `ILoggerSettings`, but at this point I can't see a way other than completely duplicating your `AppSettingsSettings` code and then add my changes.

More context on why I need such a thing (may be something to consider as you re-think settings providers):

I'd like to add the ID of the current process to the rolling file name, because my users run several instances of my app on the same VM.

```
<add key="serilog:write-to:RollingFile.pathFormat" value="C:\logs\myapp-{ProcessId}-{Date}.log" />
```

At first I thought about updating the rolling strategy, but didn't feel right - seemed too specific to my case given the process ID never changes whist the app is open (unlike the date). Not sure what you think.

Cheers!

nblumhardt:
Interesting - thanks for the follow-up. Definitely a valid scenario. Will give this some thought!

nblumhardt:
@CaioProiete what would you think of this hack?

```csharp
Environment.SetEnvironmentVariable("PROCESSID", currentPid.ToString());
```

Then:

```xml
<add key="serilog:write-to:RollingFile.pathFormat"
         value="C:\logs\myapp-%PROCESSID%-{Date}.log" />
```

? :-)

CaioProiete:
@nblumhardt Clever for sure :). But I'd be too afraid to use in Production, if I'm honest.

It works on my machine(tm), but I don't like the fear of knowing that I could get a `SecurityException` one day, on a user's machine, and have to figure out why (Windows app, hundreds of end-users).

Maybe I'm being overcautious, but between using this hack or determining the location dynamically in C#, I'd stick with the latter for the time being, until we have a better alternative.

Nice try, though! :)

nblumhardt:
Fair enough :-)

I don't think we'll move on this directly right now, thanks for the suggestion and effort on this though. Hope it's all going well!

artganify:
@nblumhardt

How about using enrichers instead of Environment? Or some other 'serilog
specific environment' instead of that weird .NET thing? Also, maybe not
only taking a string, but a string resolver (Func<string>) so users could
easily take care of exception handling on their own.

    .Enrich.Environment("processId", () => {
        try {
            return Util.GetProcessId();
        } catch {
            return string.Empty;
        }
    );

    .WriteTo.RollingFile("log_{ProcessId}.txt);

PR #467, opened 2015-06-05
Added NuGet Badge

merbla:
Added a little visibility when adding viewing the repo.  This should be extended for MyGet beta builds


nblumhardt:
:+1:

PR #471, opened 2015-06-16
Including logging level in Trace logger

ldemidov:
On Azure, there's a an option to log Trace events to a destination set
on Azure Portal (storage tables or files).    Since Serilog was not
including the event type in the trace log, trace messages generated by
Serilog are not being logged.  There's also a minimum logging level
setting in Azure that uses this event type to choose logging messages to
log.

To get this feature working properly, I updated the Trace sink to
include the event type in the Trace message based on the Serilog
LogEventLevel.


nblumhardt:
Thank you, looks like a good improvement.

Issue #479, opened 2015-06-30
Avoiding params array and boxing allocations when the level isn't enabled


justinvp:
Hi Nick!

I'm really loving Serilog! In addition to all the structured logging goodness, I like the simple usage that the static `Log` class provides:

```c#
int x = 1;
int y = 2;
int z = 3;

Log.Debug("The values are {X}, {Y}, and {Z}", x, y, z);
```

However, this results in 4 heap allocations (the params array allocation and 3 box allocations for the value types) even when the debug level isn't enabled. Such allocations aren't immediately obvious to most developers unless they're aware of the issue.

This can be avoided by checking `IsEnabled` before the call:

```c#
if (Log.IsEnabled(LogEventLevel.Debug))
    Log.Debug("The values are {X}, {Y}, and {Z}.", x, y, z);
```

But then I have to remember to always check `IsEnabled` before the actual call to avoid the allocations when the level isn't enabled, which isn't as nice as simply calling `Log.Debug`.

It'd be nice if there were some built-in generic overloads (up to ~3-4 parameters), similar to `String.Format`'s and `Console.WriteLine`'s arg0, arg1, and arg2 overloads, but generic. The implementation inside such overloads could first check `IsEnabled` and only allocate the array (and box any value types) if the level is enabled.

NLog provides generic overloads along these lines for this reason.

Something like:

```c#
public static class Log
{
    ...

    [MessageTemplateFormatMethod("messageTemplate")]
    public static void Debug<T0>(string messageTemplate, T0 propertyValue0)
    {
        if (IsEnabled(LogEventLevel.Debug)
            Logger.Debug(messageTemplate, new object[] { propertyValue0 });
    }

    [MessageTemplateFormatMethod("messageTemplate")]
    public static void Debug<T0, T1>(string messageTemplate, T0 propertyValue0, T1 propertyValue1)
    {
        if (IsEnabled(LogEventLevel.Debug)
            Logger.Debug(messageTemplate, new object[] { propertyValue0, propertyValue1 });
    }

    [MessageTemplateFormatMethod("messageTemplate")]
    public static void Debug<T0, T1, T2>(string messageTemplate, T0 propertyValue0, T1 propertyValue1, T2 propertyValue2)
    {
        if (IsEnabled(LogEventLevel.Debug)
            Logger.Debug(messageTemplate, new object[] { propertyValue0, propertyValue1, propertyValue2 });
    }

    // TODO add overloads for the other levels

    ...
}
```

[...]

PR #484, opened 2015-07-08
Add Log overloads that help avoid array and boxing allocations

justinvp:
This PR is the first part of #479 for Serilog 1.5.

Are you OK with the type parameter and parameter naming?

 1. I considered `TPropertyValue` but chose the shorter `T` to align with [`Action<T>`](https://msdn.microsoft.com/en-us/library/018hxwa8(v=vs.110).aspx)/[`Func<T, TResult>`](https://msdn.microsoft.com/en-us/library/bb549151(v=vs.110).aspx).
 2. For the index suffix on the type parameter and parameter names, I debated whether to start with `0` or `1` and went with `0` to align with the parameter names of [`String.Format`](https://msdn.microsoft.com/en-us/library/d9t40k6d(v=vs.110).aspx)'s overloads and because the `messageTemplate` supports specifying numeric property names that start at zero.
 3. For the overload that just has a single `propertyValue` parameter, I left the index off of the type parameter and parameter names, to align with [`Action<T>`](https://msdn.microsoft.com/en-us/library/018hxwa8(v=vs.110).aspx)/[`Func<T, TResult>`](https://msdn.microsoft.com/en-us/library/bb549151(v=vs.110).aspx), though, this is in conflict with [`String.Format`](https://msdn.microsoft.com/en-us/library/fht0f5be(v=vs.110).aspx), which keeps the `0` suffix on the parameter name for the single parameter case.

I don't have a strong preference either way on these.

Let me know if you'd like me to make any changes.

### DIFF
--- a/src/Serilog/Log.cs
+++ b/src/Serilog/Log.cs
@@ -108,12 +108,117 @@ namespace Serilog
         /// Write a log event with the specified level.
         /// </summary>
         /// <param name="level">The level of the event.</param>
-        /// <param name="messageTemplate"></param>
-        /// <param name="propertyValues"></param>
+        /// <param name="messageTemplate">Message template describing the event.</param>
+        /// <param name="propertyValue">Object positionally formatted into the message template.</param>
+        [MessageTemplateFormatMethod("messageTemplate")]
+        public static void Write<T>(LogEventLevel level, string messageTemplate, T propertyValue)
+        {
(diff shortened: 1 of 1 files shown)


nblumhardt:
Looks perfect - thanks!

khellang:
All those `MessageTemplateFormatMethod` attributes just reminded me that I need to get on with the R# plugin again. So much to do... *sigh*

Issue #487, opened 2015-07-11
Port Logentries Sink to PCL


ravensorb:
Once PeriodicBatching is ported to the PCL library it would be great to see Logentries sink ported as well.

nblumhardt:
Closing as dependent on #486 - once that work is done, PRs covering logentries and other currently full-fx sinks would be great.

hgjura:
I tried to add the Logentries sink to PCL library and it failed. I thought there was a PCL version of this.

Issue #486, opened 2015-07-11
Request: Move PeriodicBatching to PCL version of library


ravensorb:
Serilog is a fantastic logging framework -- unfortunately it is almost unusable in mobile applications as move of the sinks leverage PeriodicBatching and that is only provided in the Full client version.

Note: I am working on a port of it to full PCL -- only issue is the Timer object is not support in Profile259.  I think I have a possible solution and am testing now.

merbla:
@ravensorb there have been a few sinks that required a PCL which have run into this issue.  Most have implemented a mechanism particular to that sink.

e.g.
 - [Previous Issue (Closed)](https://github.com/serilog/serilog/pull/179) 
 - https://github.com/serilog/serilog-sinks-splunk/blob/master/src/Serilog.Sinks.Splunk/Sinks/Splunk/SplunkViaHttpSink.cs#L179

Be interested in what you come up with however, there are already a few variations on top of the existing PeriodicBatching that need to be considered.

ravensorb:
I just created the PR. Let me know what you think

nblumhardt:
Thanks for all the investigation Shawn!

To add to Matthew's list - another portable `Timer` replacement that's had a bit of testing: https://github.com/continuousit/seq-client-portable/blob/master/Seq.Client.Portable/Client/Portable/PortableTimer.cs

Definitely needs to happen! :-)

I don't think the timing is great for this right now - VS2015 support is coming soon and there are already a few changes in the works that need to line up there.

I think we should leave **this issue open**, but close the PR and additional (logentries) tickets for the time being. Once 2015 has shipped and the great project reoganisation completed, a lot should fall into place.

What do you think? Cheers!

ravensorb:
Nice, doesn't support the change method but still I like it - might be easier to maintain.

If you me to cancel the PR, that's is ok with me -- I just have an immediate need to support Serilog in a series of mobile PCL based applications so I will need to keep this fork going for now.

Would it be easier if I integrated this into a newer branch? I'd be willing to do another PR if that would help.

nblumhardt:
@ravensorb I think for now let's put this on hold while the core project work happens - making progress, hope to have something on the `dev-v2` branch shortly, but it will be a while before it has stabilised. Cheers!

ravensorb:
ok, I'll cancel it.

ravensorb:
Any update on this? I'd like to stop needing to maintain a private build :smile:

nblumhardt:
[...]

PR #488, opened 2015-07-12
Moved PeriodicBatching to PCL project and updated build process

ravensorb:
This PR includes a port of the PeriodicBatching Sink to the PCL project. In addition to that, I migrated the build process to use psake (this can be dropped -- I just find it easier to work with).

Note: To move PeriodicBatching to PCL, a replacement for the Timer object was needed (I targeted Profile259 and the primary profile).  To replace this, I opted for a implementation based on the Task object.  All test cases pass and I was able to port the Serilog Sing to PCL with success.  That said, if anyone has a better alternative, I welcome the suggestions

Note: This is related to Issue #487


ravensorb:
Ok, I just realized this project uses AppVeyor for building.  Based on that, I'll roll back my changes for the build process -- sorry about that.

merbla:
Related to #486

PR #497, opened 2015-07-24
Changed to use branch specific badges

merbla:
Updating read me to get ready for build/nuget badges for pre-release.  (v2 etc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,12 @@
-Serilog [![Build status](https://ci.appveyor.com/api/projects/status/b9rm3l7kduryjgcj?svg=true)](https://ci.appveyor.com/project/serilog/serilog) [![NuGet Version](http://img.shields.io/nuget/v/Serilog.svg?style=flat)](https://www.nuget.org/packages/Serilog/) 
-=================================================================================================================================================
+#Serilog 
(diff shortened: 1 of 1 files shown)


nblumhardt:
Hey mate, what do you think about just updating the badge to point to master only? Not sure we're ready enough on the v2 branch to make publicising its status worthwhile, I'm still on the fence about a lot of it :-) ...?

merbla:
Do you think dev is worthwhile (along with master)?  Thinking of people quickly seeing the PR status of something recently moved to dev.  Also was thinking on how we would go about surfacing the dev NuGet package.

nblumhardt:
LGTM mate.

RE the dev package, perhaps we should sort out that MyGet feed; will dig up the last thread on it... :+1:

nblumhardt:
After merging, looked a bit funny not having the summary of the repo purpose ahead of the table, so gave it a slight tweak on the way in (couldn't resist :-))

merbla:
+1 for style :satisfied:

merbla:
Was looking at MyGet, now trying understand the benefit of using that vs alpha/beta NuGet from AppVeyor on dev branch.

Thoughts??

nblumhardt:
AppVeyor's definitely a welcome simplification if we can make it work, +1 there

Issue #485, opened 2015-07-09
Change logger/sink configuration at runtime?


artganify:
Is there a straightforward way to change the logger configuration (or sink configuration) at runtime?

I generally set up my main logger with a rolling file sink to print out critical errors (especially at startup) as this works in most scenarios. Independently of what type of application I have (service, website, console), I only need to ensure the file system is write-able and I'm good to go.

    // dummy
    var logfilepath = GetLogFilePath(Assembly.GetEntryAssembly());
    Log.Logger = new LoggerConfiguration()
                            .WriteTo.RollingFile(logfilepath)
                            .CreateLogger();
    Log.Debug("It works!");

However, I also would like to log into other sources that are more prone to fail at startup (e.g. databases when wrong connection strings are given). So what I would like to achieve is the following:

1. Set up a simple, stable logging sink for first startup
2. Set up e.g. database, log everything related to it
3. If e.g. database is up and running, add database sink to serilog.

The only way I see to accomplish this is:

1. Create a logger configuration
2. Do stuff
3. Create a new logger configuration and overwrite the existing one.

Dummy code:

    var logfilepath = GetLogFilePath(Assembly.GetEntryAssembly());
    Log.Logger = new LoggerConfiguration()
                            .WriteTo.RollingFile(logfilepath)
                            .CreateLogger();
    Log.Debug("It works!");

    Log.Debug("Setting up database with {ConnectionString}", connectionString);

    Log.Logger = new LoggerConfiguration()
                                .WriteTo.RollingFile(logfilepath)
                                .WriteTo.SomeDatabaseSink(connectionString)
                                .CreateLogger();

Is there a cleaner, more straightforward approach to change the logger configuration other than overwriting it completely?

nblumhardt:
Here's one option that's debatably better:

```csharp
// <snip>
Log.Debug("Setting up database with {ConnectionString}", connectionString);

Log.Logger = new LoggerConfiguration()
                            .WriteTo.Sink((ILogEventSink)Log.Logger)
                            .WriteTo.SomeDatabaseSink(connectionString)
                            .CreateLogger();
```

Instead of letting the old logger drop, the new one just redirects to it.

It's not ideal, but if you set up enrichment etc. on the second (rather than the first) logger it might cut a bit of duplication/noise out. It also has the property of being less likely to cause file locking issues.

:+1: or :-1: ?

IanYates:
I have a similar issue and hadn't thought of @nblumhardt 's solution.

[...]

PR #499, opened 2015-07-29
More pliable LoggerConfiguration

trevorreeves:
In trying to add some fluent builder patterns on top of LoggerConfiguration I have come across the limitation that sinks and filters etc cannot be removed from the configuration object once they have been added.  Is there a fundamental reason why this would be a bad thing to do?

Assuming an answer of 'no', this branch contains 2 commits; 

- the first is a simple one to change the collections inside LoggerConfiguration to be protected rather than private.  This would allow specializations of LoggerConfiguration to add the features they want.

- the second are some simple methods that my current use case would find useful - to reset the configuration object to an empty state, or to empty the individual collections.

The first commit is very harmless at least, and would open it up for many customizations.

### DIFF
--- a/src/Serilog/LoggerConfiguration.cs
+++ b/src/Serilog/LoggerConfiguration.cs
@@ -140,6 +140,56 @@ namespace Serilog
         }
 
         /// <summary>
+        /// Resets the configuration object to its original default state.
+        /// </summary>
+        public LoggerConfiguration Reset()
+        {
+            ClearSinks();
+            ClearEnrichers();
+            ClearFilters();
+            ClearDestructureConfig();
+            _minimumLevel = LogEventLevel.Information;
+            return this;
+        }
+
+        /// <summary>
+        /// Removes all configured sinks.
+        /// </summary>
+        public LoggerConfiguration ClearSinks()
+        {
+            _logEventSinks.Clear();
+            return this;
+        }
+
+        /// <summary>
+        /// Removes all configured enrichers.
+        /// </summary>
+        public LoggerConfiguration ClearEnrichers()
+        {
+            _enrichers.Clear();
+            return this;
+        }
+
+        /// <summary>
+        /// Removes all configured filters.
+        /// </summary>
+        public LoggerConfiguration ClearFilters()
+        {
+            _filters.Clear();
+            return this;
+        }
+
+        /// <summary>
+        /// Removes all configured destructuring customizations.
+        /// </summary>
+        public LoggerConfiguration ClearDestructureConfig()
+        {
+            _additionalScalarTypes.Clear();
+            _additionalDestructuringPolicies.Clear();
+            return this;
+        }
+
+        /// <summary>
         /// Create a logger using the configured sinks, enrichers and minimum level.
         /// </summary>
         /// <returns>The logger.</returns>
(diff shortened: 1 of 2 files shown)


nblumhardt:
Thanks for the PR - can you please send a bit more info about your scenario? Keen to support it, but would rather avoid loosening up LoggerConfiguration if we can (especially if there is an alternative implementation possible).

LoggerConfiguration also takes ownership of several disposable things passed in. Not sure about adding any APIs that would suggest it is reusable, since at least with most sinks, they need to be owned by a single logger instance only.

May just need a bit more context. Thanks again for jumping into it!

trevorreeves:
Yeah no worries.  Thanks for the reply.  Will try and get across my scenario here, (failing that I'll create a proper sample app if needed). And maybe you can point me in a better direction.

My scenario isn't that I want to modify a LoggerConfiguration in between creating multiple loggers, but it's that a single configuration instance will be manipulated by different actors during startup, who may want to remove things from the configuration, or not, before it reaches its final state.  After startup, when in its final state, the configuration object will be used to create loggers.

I have been writing a config builder API for DNX, to support named loggers - roughly to meet the needs I mentioned here (https://github.com/serilog/serilog-framework-configuration/issues/1).  I've ended up with a builder component leveraging MS' OptionsModel.  With that in mind, below is a simple example of using the plain MS component, with the LoggerConfiguration employed as an options class: 

```c#
public class Startup
{
        public void ConfigureServices(IServiceCollection services)
        {
             services.Configure<LoggerConfiguration>(config => config
                  .MinimumLevel.Error()
                  .WriteTo.Console());

            // then later on during the ConfigureServices phase...

            services.Configure<LoggerConfiguration>(config => config
                 .Reset() // we could either choose to reset the configuration, or not.
                 .WriteTo.File("test.log"));
       }
}
```

The restriction imposed on us here by the options model, is that we don't have the freedom to create a new configuration object when we want to 'reset', we can only modify the existing one (we only get an Action<LoggerConfiguration> to play with). 

Reflecting on this more, I think you'd be justified in saying this isn't the problem of the LoggerConfiguration, it's that of the builder pattern around it.

----

Finally, I just wanted to clarify some points you mentioned:

> LoggerConfiguration also takes ownership of several disposable things passed in. Not sure about adding any APIs that would suggest it is reusable, since at least with most     sinks, they need to be owned by a single logger instance only.

So only one sink instance should only be owned by one ILogger?
```c#
var config = new LoggerConfiguration().WriteTo.File("test.log");
var l1 = config.CreateLogger();
var l2 = config.CreateLogger(); // a bad thing?
```

artganify:
Hi @trevorreeves 

I actually had a similar question a while back. There related issue is #485, in which @nblumhardt suggests for altering the logging configuration at runtime:

    // <snip>
    Log.Debug("Setting up database with {ConnectionString}", connectionString);

                            .WriteTo.Sink((ILogEventSink)Log.Logger)
                            .WriteTo.SomeDatabaseSink(connectionString)
                            .CreateLogger(); 

It's not really obvious and pretty and it seems manipulating the configuration at runtime in various ways is indeed something wanted (many other loggers feature that - for instance *log4net* with their logging hierarchy) but I then sticked with that and it worked fine.

nblumhardt:
@trevorreeves RE `// a bad thing?` - yes, definitely not safe. I'd expect `LoggerConfiguration` to throw, here, but need to review the code to be sure :-)

Closing this for the time being, but would love to have some input on `serilog-framework-configuration` - the discussion might be good to carry over there.

@artganify replying to your other post, sorry about the delay.

PR #500, opened 2015-07-29
Continue logging when Exception Rendering fails

richardlawley:
Under certain circumstances, calling .ToString() on an exception can itself throw an exception.  In this situation, Serilog does not log the message or exception.  This occurrence is logged to SelfLog, but this is still tricky to find.

This PR introduces error handling at the point where the exception is rendered into the output.  If .ToString() fails, then a message combining the exception type and message is used.  If this fails, a constant is used.  Importantly, the log message is still written.

I've added test coverage of this functionality.  The Serial attribute is added for people using ncrunch, as testing SelfLog is not thread-safe.

For information only, the real-world scenario which triggered this is from a program where a dependency had been removed.  I think the library may have also been doing some dynamic loading.  When trying to use the functionality exposed in the library, a FileNotFoundException was generated.  This exception was also generated when trying to view the StackTrace of the exception, which in turn was used by ToString.

### DIFF
--- a/src/Serilog/Formatting/Display/OutputProperties-net40.cs
+++ b/src/Serilog/Formatting/Display/OutputProperties-net40.cs
@@ -15,6 +15,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Serilog.Debugging;
 using Serilog.Events;
 
 namespace Serilog.Formatting.Display
@@ -51,6 +52,11 @@ namespace Serilog.Formatting.Display
         public const string ExceptionPropertyName = "Exception";
 
         /// <summary>
+        /// Message which will be rendered in place of an exception if the original exception could not be rendered
+        /// </summary>
+        public const string FailedToRenderExceptionMessage = "-Could Not Render Exception-";
+
+        /// <summary>
         /// Create properties from the provided log event.
         /// </summary>
         /// <param name="logEvent">The log event.</param>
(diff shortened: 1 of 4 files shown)


nblumhardt:
Thanks for the thorough PR.

I still need some time to think about this - it's definitely an improvement, but `Exception` is possibly not a special case. Just about any log event's parameter can fail in `ToString()`, so having (keeping?) consistency there seems like something to explore too.

Just for interest's sake, is there a particular exception type that's prone to doing this?

Cheers!

richardlawley:
The exception I encountered this on was a FileNotFoundException being triggered by a dependency of my application not having been deployed to the output directory.  It's possible this was made more complicated by some use of dynamic loading within the affected library.  I'm well aware this is an edge case and most people probably won't see it, but it did take me quite a bit of time to figure out what was happening given that there was no indication from the logging of an error.

Of the explicit parameters in that function, I think `Exception` is a special case as the others are all simple types and should not be able to cause exceptions, other than potentially MessagePropertyName, which I would've expected should be handled within the `LogEventPropertyMessageValue` class.

artganify:
@nblumhardt Exceptions are indeed some special case because first developers rely hevily on them for debug and tracing purposes and they are the most straightforward way to leverage exceptional situations and second) even Serilog offers additional methods to pass in exceptions directly, independently of the rest of the formatted message.

So:

1. Developers should fix their code that an exception never throws another exception at best. They should be simple property bags which contain only information that is valid during the state the program behaves in an unexpected manner (and not some services which pretty print debug messages).

2. Serilog should log exceptions, no matter what. Having exceptions swallowed is the worst I can think of for debugging and tracing them down in the debug pipeline of a framework can be extremely time consuming.

What do you guys think?

nblumhardt:
I don't think my comment above was clear, sorry:

> I still need some time to think about this - it's definitely an improvement, but Exception is possibly not a special case

As in, if we're going to fix this then perhaps we can consider making a general `SafeToString()`-type function that could be used:

1. Here for exceptions
2. In other sinks/formatters (e.g. when rendering exceptions to JSON)
3. (Possibly) when converting other objects to scalar values in place of `ToString()`

The reason for avoiding special cases is to make sure users' intuition is applicable - i.e. if Serilog handles a misbehaving `ToString()` via the self-log in some places, but by rendering a special message in others, it'll be easy for people to make mistakes.

nblumhardt:
@richardlawley any thoughts on this? Cheers! :-)

richardlawley:
I think doing it on the Exception would be an improvement on now, and I don't think this would constitute inconsistent behaviour.  Exceptions are a special case because they're not always generated by the developer - if you explicitly log something where ToString breaks, at least it's something you've normally had control over.

If you did want to fix this globally, would you manage to get all occurrences?  Would it affect third party sinks which would then behave inconsistently, or would you manage to bury it all within the implementation?  If the latter then I'd say it was worth doing, if not then I would say No.

nblumhardt:
Guys, I think we're going to leave this for now. There are some valid considerations both ways - what's tipping this towards closing is that as written, we'll only end up handling this when writing to log files as text; the very common JSON sinks, as well as colored console sinks, won't be covered by this change.

In the long run I think this might still be interesting to try, but needs to go back to the proposal stage to see if we can come up with a broader implementation.

Thanks again for the PR @richardlawley and all the input, much appreciated.

PR #509, opened 2015-08-17
Fix build

khellang:
This fixes the build by adding the NuGet.config. This file is added automatically when using the beta6 "file > new"-templates.

### DIFF
--- a/src/Serilog/Project.json
+++ b/src/Serilog/Project.json
@@ -8,7 +8,6 @@
     "frameworks": {
         "net40": {
             "compilationOptions": {
-                "define": [ "NET40" ],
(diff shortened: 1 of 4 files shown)


src/Serilog/Project.json
```diff
@@ -8,7 +8,6 @@
     "frameworks": {
         "net40": {
             "compilationOptions": {
-                "define": [ "NET40" ],
```
khellang: These are added automatically by DNX, i.e. for `net45`, a `NET45` preprocessor directive is added.

nblumhardt:
:+1:

PR #510, opened 2015-08-17
Added initial build script for DNX-based builds

khellang:
This should build, pack and test everything using both regular CLR and core CLR :grin:


khellang:
The build will probably fail until #509 is pulled in :smile:

khellang:
Alright, I kicked the tires on this again. Let's see if we get the build going :smile:

khellang:
Argh! :angry: 

> error CS0656: Missing compiler required member 'Microsoft.CSharp.RuntimeBinder.Binder.GetMember'

Does anyone know what this means? It's for the portable target.

nblumhardt:
Thanks for picking this one up, the DNX support needs a lot of love.

No ideas regarding the build error, had some similar trouble with .NET 4 targeting on the beta5 tools and gave up for the time being. Hoping your luck is better ;-)

akoeplinger:
@khellang your error is caused by a missing dependency on "System.Dynamic.Runtime". Also, I just tried building it on Linux and you need to rename `Project.json` -> `project.json`

khellang:
@akoeplinger You're awesome, as always! :metal::grin::metal: Thank you so much :+1:

khellang:
@nblumhardt Alright, now the projects are building successfully and packages are produced at https://ci.appveyor.com/project/serilog/serilog/build/283/artifacts.

nblumhardt:
Champion :-)

Issue #505, opened 2015-08-12
Support LoggingLevelSwitch on a per-sink basis


nblumhardt:
Sink configuration methods accept a `restrictedToMinimumLevel` parameter of type `LogEventLevel`. Augmenting this or replacing it with `LoggingLevelSwitch` would permit individual sinks to be manipulated at runtime. May assist with #485 and similar issues.

caseymarcallen:
@nblumhardt We are looking at the Serilog.Sinks.Seq sink and have modified it to accept a `LoggingLevelSwitch` instead of a `LogEventLevel` so the min level that is returned from the HTTP call to the Seq server can be exposed to the rest of the application rather than being hidden inside the internal `_minimumAcceptedLevel` variable. 

Our specific use case is that we are exposing a route in our MVC website so that web clients can log from javascript. We then use the website and the Seq Sink to forward this information on to the internal Seq Server. We do however want to expose that min level back to the javascript so that they only post the log events over the network back to the server if the level is greater than the min level as configured by the Seq Server API Key. Hence we want access to that `mininumAcceptedLevel` value that the sink if using.

I was about to submit a pull request against the Seq Sink, then noticed this thread here. What are your thoughts on the `SeqSink` using a `LoggingLevelSwitch` internally instead of a `LogEventLevel` and allowing the switch to be passed in via the `SeqLoggerConfigurationExtensions.Seq()` extension method instead of the `LogEventLevel`. It's not a Serilog global solution like you are proposing here, but it suits the Seq Sink purpose.

nblumhardt:
Hi Casey, thanks for the info. Good to validate there's more than one scenario for this.

To avoid too much version compatibility juggling, I'll tackle the core Serilog change ASAP, probably in the next few days. Some help propagating this out through the various sinks would be great, if you're still keen to jump in once the foundation is laid!

caseymarcallen:
Sounds good. Happy to help where I can.

nblumhardt:
The foundations are there in #511 - a little stalled on deciding how it should go out (1.5, 1.6, 2.0, ...) - thoughts appreciated on that one.

caseymarcallen:
@nblumhardt I've implemented the functionality in the Seq Sink and while testing various scenarios noticed that if you are logging with an API key to Seq and have the min level set to something against the Seq API Key, if you then delete that key Seq will continue to report back to the Seq Sink the previous API key min level. I would have expected it to start reporting back null. Bug in Seq or by design?

[...]

PR #511, opened 2015-08-17
Add a `levelSwitch` parameter to all sink configuration methods

nblumhardt:
An implementation for #505. Supports configuration like:

```csharp
var fileLevel = new LoggingLevelSwitch(LogEventLevel.Warning);

  .WriteTo.File("...", levelSwitch: fileLevel)
  .CreateLogger();

// Later...
fileLevel.MinimumLevel = LogEventLevel.Debug;

log.Information("This will be written to the file");
```

This is a **binary breaking change** for user code that calls sink configuration methods, but not for third-party sink assemblies. I don't think it justifies a "2.0" version bump, but open to ideas.

### DIFF
--- a/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs
+++ b/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs
@@ -13,18 +13,17 @@
 // limitations under the License.
 
 using System;
-using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Threading;
 using Serilog.Configuration;
+using Serilog.Core;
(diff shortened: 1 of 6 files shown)


khellang:
Maybe this could be re-targeted to dev-v2? Do you have a bunch of breaking changes pending? Maybe it's time to break stuff? :facepunch:

nblumhardt:
@khellang I have resisted the temptation so far to start piling stuff onto `dev-v2` because of the uncertain RTM date for the DNX tooling. In the grand scheme of things though, it does sound much wiser to do _one_ breaking release, even if that means a prolonged (six month?) bake time. Will re-target :-) :facepunch:

nblumhardt:
Moved to #515.

PR #512, opened 2015-08-17
Run tests on DNX

khellang:
This PR

- Adds Serilog.Tests back into the solution (using project.json)
- Migrates tests from NUnit to xUnit (since xUnit runs on CoreCLR as well)
- Adds a `test` command, so you can do `dnx . test`

There are a couple of issues, though;

- There are 4 failing tests. These are just skipped for now:
  - `Serilog.Extras.AppSettings.Tests.AppSettingsTests.EnvironmentVariableExpansionIsApplied`
  - `Serilog.Extras.AppSettings.Tests.KeyValuePairSettingsTests.FindsConfigurationAssemblies`
  - `Serilog.Tests.Context.LogContextTests.DoesNotPreventCrossDomainCalls`
  - `Serilog.Tests.Sinks.PeriodicBatching.PeriodicBatchingSinkTests.WhenAnEventIsEnqueuedItIsWrittenToABatch_OnTimer`
- xUnit has no `Assert.Inconclusive`, which is used two places in `Serilog.Tests.Context.LogContextTests`

This builds on top of #510, so that should be merged first, if you want to accept this.


test/Serilog.Tests/Context/LogContextTests.cs
```diff
@@ -143,7 +144,7 @@ public void MultipleNestedPropertiesOverrideLessNestedOnes()
         // since user property types may not be serializable.
         // Fails if the Serilog assemblies cannot be loaded in the
         // remote domain. See also LogContext.Suspend()
-        [Test]
+        [Fact(Skip = "Fails on DNX451")]
```
nblumhardt: By "DNX451" do you mean "DNXCORE50"? I was under the impression 451 and .NET 4.5 were roughly equivalent. I am probably a bit behind :-)

test/Serilog.Tests/Settings/AppSettingsTests.cs
```diff
@@ -1,18 +1,18 @@
...
 using Serilog.Events;
 using Serilog.Tests.Support;
 
 namespace Serilog.Extras.AppSettings.Tests
 {
-    [TestFixture]
     public class AppSettingsTests
     {
-        [Test]
+        [Fact(Skip = "Fails on DNX451")]
```
nblumhardt: Is it possible to raise a ticket listing the skipped tests and what needs to happen to get them back to happy?

nblumhardt:
Epic PR! :-) Love your work. Will need a little more time to run over it and review, will try to turn around as quickly as possible.

test/Serilog.Tests/Context/LogContextTests.cs
khellang: Yes, but there could be several reasons for why this fails now. It could be the migration to xUnit (I know they run tests in parallel), or the fact that it runs through DNX.

test/Serilog.Tests/Settings/AppSettingsTests.cs
khellang: I raised #513 to address this :smile:

Issue #507, opened 2015-08-16
Seq sink - dynamically increasing log level not processed until event transmitted


nblumhardt:
It's possible to change the Seq sink's logging level on the fly from the server. This works fine for changes between lower levels (`Information` <-> `Debug`) but if the level is set to a high value (`Fatal`) then a change to a lower level (`Information`) won't take effect until an event matching the previous level is processed - which in the case of `Fatal` or `Error` might be practically never.

The batching sink implementation may need a way to periodically "wake up" the sink to check its current level (doing this with a thread may complicate shutdown).

PR #514, opened 2015-08-19
PeriodicBatchingSink hook for processing on ticks even when the batch is empty

nblumhardt:
The Seq sink needs an opportunity to do periodic processing, regardless of whether any events have been emitted (#507). This PR adds `OnEmptyBatch()` to `PeriodicBatchingSink` to enable this to occur on the same timer thread used for batch processing, so that additional dispose/shut-down code isn't required.

I'm not certain whether other sinks can make use of this at present, but the general scenario (_query a remote service for current logging level_) sounds like one that will appear again, especially after #511.

### DIFF
--- a/src/Serilog.FullNetFx/Sinks/PeriodicBatching/PeriodicBatchingSink-net40.cs
+++ b/src/Serilog.FullNetFx/Sinks/PeriodicBatching/PeriodicBatchingSink-net40.cs
@@ -132,7 +132,10 @@ namespace Serilog.Sinks.PeriodicBatching
                     }
 
                     if (_waitingBatch.Count == 0)
+                    {
+                        OnEmptyBatch();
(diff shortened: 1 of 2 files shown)


nblumhardt:
@serilog/core anyone with an opportunity to review? :-D

merbla:
Looks good, is there any merit (possibly v2) having a similar ability to do periodic work at the `Logger` level? I am thinking for diagnostic purposes to view statistics when multiple sinks/levels are in use.

- **Sink A** (110 *Info*, 110 *Warning*, 3 *Error*)
- **Sink B** (10 *Warning*, 3 *Error*)

nblumhardt:
@merbla thanks. At the logger level there isn't a thread to do the work on, so might be tough.

PR #518, opened 2015-08-26
Added Write overload with timestamp to ILogger

dougmulley:
This verload is useful for preserving timestamps when merging logs from another source into Serilog sinks.

### DIFF
--- a/src/Serilog/Core/Pipeline/Logger.cs
+++ b/src/Serilog/Core/Pipeline/Logger.cs
@@ -123,6 +123,12 @@ namespace Serilog.Core.Pipeline
         [MessageTemplateFormatMethod("messageTemplate")]
         public void Write(LogEventLevel level, Exception exception, string messageTemplate, params object[] propertyValues)
         {
+            Write(DateTimeOffset.Now, level, exception, messageTemplate, propertyValues);
+        }
+
+        [MessageTemplateFormatMethod("messageTemplate")]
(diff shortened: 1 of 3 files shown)


nblumhardt:
Hi Doug, thanks for the PR :-)

This should already be handled by `void Write(LogEvent logEvent)` so I don't think we'll want to make this change to `ILogger`. It's not a fantastically straightforward API, but you can construct a `LogEvent` with a specific timestamp and pass it through.

We may need to do some work on making it easier to construct a `LogEvent`, keen to hear how it goes in your scenario!

Cheers,
Nick

dougmulley:
Building the LogEvent was my first approach. The challenge I ran into was that the types for mapping "object[] propertyValues" to an IEnumerable<LogEventProperty> (i.e. PropertyValueConverter, destructuring policies) are internal. Would you be open to making some of those types public? If those types were public, I could probably create an extension method matching the overload in this pull request.

nblumhardt:
I think that's the general functionality we need.

Just exploring the scenario a bit further, does the data you're merging in use old-style `{0}` format strings? If so, it may be possible to do something like:

```csharp
var scalars = new HashSet<Type> { typeof(int), ... };
var properties = propertyValues.Select((v, i) =>
  new LogEventProperty(i.ToString(), new ScalarValue(
    v == null || scalars.Contains(v.GetType) ? v : v.ToString())
).ToArray();
```

dougmulley:
Thanks for that tip. I'm working through an approach like that to get the properties built correctly. I should be able to get it to work with the public interface you currently have exposed. 

Thanks, again, for the guidance.

Issue #536, opened 2015-09-21
ReadFrom.AppSettings() that takes NameValueCollection


JuanjoFuchs:
```AppSettingsSettings.Configure(LoggerConfiguration loggerConfiguration)``` is accessing ```ConfigurationManager.AppSettings``` directly.

It would be helpful for projects that do not have ```ConfigurationManager``` available and want to use an **app.config** file like any other web or console project (http://stackoverflow.com/a/5191101/1277401), taking advantage of all the ```AppSettingsSettings``` features.

**Comment by nblumhardt** [comment]
Thanks for the issue and PR.

Just looking over your changes, it appears that `.ReadFrom.KeyValuePairs()` with the secondary configuration passed through should cover this scenario. What are the limitations there?

Many thanks!
Nick

**Comment by JuanjoFuchs** [comment]
The limitations I've found in ```.ReadFrom.KeyValuePairs()``` are: 

- It expects keys to not have ```serilog:``` as a prefix.
- It doesn't expand environment variables.
- You need to add explicit ```using:``` keys to use serilog's built-in sinks.

In other words, ```.ReadFrom.KeyValuePairs()``` expects a set of keys totally different from what you would use for ```.ReadFrom.AppSettings()``` and I am using the same ```app.config``` file, it's just that sometimes it's not available through ```ConfigurationManager.AppSettings```. 

So, as a user, I thought it would be more transparent if I could have the same set of features with the same set of inputs, but being able to chose how to load them from file. But then again, maybe there's another way to achieve this.

**Comment by nblumhardt** [comment]
Thanks for the follow-up, makes sense. Is there a way to express the API more in terms of configuration files rather than `IDictionary<,>`? The dictionary makes the overlap with `KeyValuePairs` a little confusing - if the overload of `AppSettings()` took a secondary configuration file name as an argument the purpose of the API might be clearer. What do you think?

Just for the benefit of anyone searching for an interim solution (and so I could see it in all its ugly glory myself :-)) I've included the workaround version below:

```csharp
var settings = // Open the dictionary here
settings.Add("serilog:using:_BuiltInSinks", typeof(AppSettingsSettings).Assembly.FullName);

Log.Logger = new LoggerConfiguration()
  .ReadFrom.KeyValuePairs(settings
    .Where(kvp => kvp.StartsWith("serilog:")
    .ToDictionary(kvp => kvp.Substring(8), Environment.ExpandEnvironmentVariables))
  .CreateLogger();
```

[...]

PR #537, opened 2015-09-21
ReadFrom.AppSettings() that takes key value pair collection

JuanjoFuchs:
Fixes serilog/serilog#536 - Allow ReadFrom.AppSettings to receive a list of key value pairs and process them like any other App.config appSetting.


nblumhardt:
Just noticed this is still open but the linked issue is closed. Thanks for driving this @JuanjoFuchs - maybe one to pick up freshly at https://github.com/serilog/serilog-settings-appsettings if you still have an interest in this scenario. All the best!

PR #540, opened 2015-09-22
Fixes #532 - adds `WriteTo.Logger()` accepting `ILogger`

### LINKED ISSUE #532: WriteTo.Logger() that takes an ILogger
This method is missing because although loggers do implement ILogEventSink, they're disposable too, so disposing a "new" logger would inadvertently dispose the "old" one.

It should be easy however to implement this method in such a way that the logger is adapted into a sink that does not propagate disposal.

nblumhardt:
Also now ensures sub-loggers are disposed when appropriate, which should not be a breaking change but will fix flushing of sub-loggers when the parent is shut down.


khellang:
:+1:

PR #550, opened 2015-10-25
Add support for multiple sets of AppSettings using custom prefixes

### LINKED ISSUE #430: Support multiple sets of AppSettings using a custom prefix.
At the moment AppSettings are read using `serilog:` as a prefix in the appsettings config section:
``` xml
<add key="serilog:using" value="Serilog.Sinks.Elasticsearch" />
<add key="serilog:write-to:Elasticsearch.serverAddress" value="http://localhost:9200" />
```
It would be useful to allow multiple sets of configuration to be read by supporting the use of custom prefixes, e.g.:
``` xml
<add key="serilog1:using" value="Serilog.Sinks.Seq" />
<add key="serilog1:write-to:Seq.serverUrl" value="http://localhost:5341/" />
<add key="serilog2:using" value="Serilog.Sinks.Elasticsearch" />
<add key="serilog2:write-to:Elasticsearch.serverAddress" value="http://localhost:9200" />
```
which could then be read as follows:
``` c#
var logger1 = new LoggerConfiguration()
  .ReadFrom.AppSettings("serilog1")
  .CreateLogger();
var logger2 = new LoggerConfiguration()
  .ReadFrom.AppSettings("serilog2")
  .CreateLogger()

### PR DESCRIPTION
**Pull request by caioproiete**
Add support for multiple sets of AppSettings using custom prefixes. Resolves #430

### DIFF
--- a/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs
+++ b/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs
@@ -279,12 +279,13 @@ namespace Serilog
         /// <code>serilog:minimum-level</code>.
         /// </summary>
         /// <param name="settingConfiguration">Logger setting configuration</param>
+        /// <param name="settingPrefix">Optional prefix to use when reading keys in appSettings</param>
         /// <returns>An object allowing configuration to continue.</returns>
         public static LoggerConfiguration AppSettings(
-            this LoggerSettingsConfiguration settingConfiguration)
+            this LoggerSettingsConfiguration settingConfiguration, string settingPrefix = null)
         {
             if (settingConfiguration == null) throw new ArgumentNullException("settingConfiguration");
-            return settingConfiguration.Settings(new AppSettingsSettings());
+            return settingConfiguration.Settings(new AppSettingsSettings(settingPrefix));
         }
     }
 }
--- a/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs
+++ b/src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs
@@ -279,10 +279,27 @@ namespace Serilog
         /// <code>serilog:minimum-level</code>.
         /// </summary>
         /// <param name="settingConfiguration">Logger setting configuration</param>
+        /// <returns>An object allowing configuration to continue.</returns>
+        public static LoggerConfiguration AppSettings(
+            this LoggerSettingsConfiguration settingConfiguration)
+        {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by caioproiete** [comment]
Thanks @danielwertheim, that's a good point, and I thought about it when I was writing the code for this PR.

My rational for not including the `:` by default, is that I'm **concatenating** a prefix that can be whatever you want, so I should respect that.

So if your prefix is daniel-`, with a dash instead of a colon, it will respect that and will work.
e.g. 
```
var log1 = new LoggerConfiguration()
    .ReadFrom.AppSettings("daniel-")
    .CreateLogger();

<add key="daniel-serilog:minimum-level" value="Warning" />
```

Note that the implementation keeps the "serilog" prefix, and just concatenate the custom prefix in front of it, whatever it is.

---

If I was actually **replacing** the default `serilog` prefix with something else, then I would totally agree that you should not have to specify the `:` 

e.g.
```
var log1 = new LoggerConfiguration()
    .ReadFrom.AppSettings("daniel")
    .CreateLogger();

<add key="daniel:minimum-level" value="Warning" />
```

**Comment by caioproiete** [comment]
Updated unit test to work with #562  and reverted inserting the custom prefix in front of `serilog:` in AppSettings. e.g. `customprefix:serilog:directivename`

This should be good to go.

**Comment by caioproiete** [comment]
Also amended `LoggerConfigurationFullNetFxExtensions` to do same validations as `dev-v2` branch

File: src/Serilog.FullNetFx/LoggerConfigurationFullNetFxExtensions.cs [hunk 1]
```diff
@@ -267,12 +267,19 @@ public static class LoggerConfigurationFullNetFxExtensions
         /// <code>serilog:minimum-level</code>.
         /// </summary>
         /// <param name="settingConfiguration">Logger setting configuration</param>
+        /// <param name="settingPrefix">Optional prefix to use when reading keys in appSettings</param>
         /// <returns>An object allowing configuration to continue.</returns>
         public static LoggerConfiguration AppSettings(
-            this LoggerSettingsConfiguration settingConfiguration)
+            this LoggerSettingsConfiguration settingConfiguration, string settingPrefix = null)
```
nblumhardt: For v1.5, can we use an overload and non-default param so that the change is non-breaking? Otherwise all looks great, thanks for taking the time to push this one forwards.

caioproiete:
@nblumhardt: Good catch. I remember we agreed on going with the overload for v1.5, and forgot to add it.

I've just updated the PR now.

nblumhardt:
Great, thanks!

PR #555, opened 2015-10-28
Added event sources (potential feature)

artganify:
One of the big 'strenghts' of semantic/structured logging frameworks such as SLAB is the ability to create predefined event sources for the different business units or stages of an application. With event sources, log outputs can easily be unified by providing strongly typed log invocation methods rather than ambigious log messages written under the developers current discretion. The current PR contains event sources implemented as simple adapters for an underlying logging instance (since this is just a concept). They could easily be extended to true semantic sources like in SLAB :smiley:

Would this be an useful feature for Serilog?

### DIFF
--- a/Serilog.sln.DotSettings
+++ b/Serilog.sln.DotSettings
@@ -123,7 +123,8 @@
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_OBJECT_AND_COLLECTION_INITIALIZER_STYLE/@EntryValue">CHOP_ALWAYS</s:String>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/JavaScriptCodeFormatting/ALIGN_MULTIPLE_DECLARATION/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/JavaScriptCodeFormatting/JavaScriptFormatOther/ALIGN_MULTIPLE_DECLARATION/@EntryValue">True</s:Boolean>
-	<s:String x:Key="/Default/CodeStyle/CSharpFileLayoutPatterns/Pattern/@EntryValue">&lt;Patterns xmlns="urn:schemas-jetbrains-com:member-reordering-patterns"&gt;&#xD;
+	<s:String x:Key="/Default/CodeStyle/CSharpFileLayoutPatterns/Pattern/@EntryValue">&lt;?xml version="1.0" encoding="utf-16"?&gt;&#xD;
+&lt;Patterns xmlns="urn:schemas-jetbrains-com:member-reordering-patterns"&gt;&#xD;
   &lt;TypePattern DisplayName="COM interfaces or structs"&gt;&#xD;
     &lt;TypePattern.Match&gt;&#xD;
       &lt;Or&gt;&#xD;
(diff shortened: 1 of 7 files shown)


artganify:
Additional approaches or related features to consider would be:

* Named static event sources that could be retrieved from a global repository
* Changing the access modifier from `Serilog.Core.Pipeline.Logger` to be just *public* rather than *sealed* so people intending to create an *event source* could provide their own derived logger rather than building one up using the `LoggerConfiguration` (this would require access to `MessageTemplateProcessor` tho)

nblumhardt:
Thanks for the PR! It's a bit tricky to grasp the scenario without an example - could you possibly illustrate what code using this would look like?

artganify:
Sure! A requirement we're often having at our company is logging to specific, preconfigured sinks - not only based on a certain filter such as a minimum log level or the severity of an event, but completely different log configurations. Log messages are expected to have a certain format or contain certain properties. MSPNP's Semantic Logging Application Block (SLAB) is a logging framework which deeply integrates that idea using *event sources*.

The following example shows an *event source* only of interest for a developer. It prepares a logger for the application startup with simple and local sinks which are guaranteed to be fail-safe in most cases.

    public class ApplicationHostSource : EventSource {

        public void Startup() {
            // prepare a predefined startup log event,
            // containing e.g. environment information
        }

        public void Shutdown() {
            // prepare a predefined shutdown log event,
            // containing e.g. environment information
        }

        protected override void OnBuildUp(LoggerConfiguration configuration) {
            configuration.WriteTo
                .ColoredConsole()
                .WriteTo.File("startup.txt");
        }
    }

The following *event source* denotes the logging configuration for business transactions using e.g. repositories/uow or sagas.

    public class BusinessTransactionSource : EventSource {

        public void StartTransaction(IAuditor auditor, int transactionId) {
	    // describe the auditor and UOW
            // e.g. user and affected aggregates	
        }

        public void EntityChanged<TEntity>(int transactionId, TEntity entity) {
            // describe the previous and current state e.g.
            // by fetching information from the DALs object tracking
        }

        public void EndTransaction(int transactionId) {
            // describe information about the result of the transaction
	}

        protected override void OnBuildUp(LoggerConfiguration configuration) {
            // business transactions are usually of interest for a business manager
            // they will be written to a mongodb table which can be inspected
            // using a dedicated interface of the application
            configuration.WriteTo
                .MongoDB("mongo://mymongodb/log");
        }
    }

General usage in a infrastructure DAL implementation (given that we're working in the car seller business):

    public class DAL {
	
        private readonly BusinessTransactionSource _eventSource = new BusinessTransactionSource();

        public void RegisterNewCar(Car carDescription) {
            // initialize UOW
            _eventSource.StartTransaction(logisticsTeam, 3423);

            // write to underlying storage
            _eventSource.EntityAdded(carDescription);

            // commit UOW
            _eventSource.EndTransaction(3423);
        }
    }

Right, there might be other ways to introduce the concept of *event sources* to Serilog with the existing stack, e.g. using secondary loggers, custom wrapper classes and/or extensions for `ILogger`. I just thought of a simple, obvious and straightforward way to do this familiar for people that previously worked with SLAB or ETW. :smile:

nblumhardt:
Aha! Gotcha :-)

It's a great concept. I think it's something better explored with a add-on than included in the core at this point, though I'm sure you'd get a lot of input from other folks around the Serilog community.

We could list it [here](https://github.com/serilog/serilog/wiki/Community-Projects) if you decide to put something out there?

Cheers!

artganify:
Great idea. Will do. :)

Issue #535, opened 2015-09-21
.NET Core


MatthiasJost:
Is the .NET Core migration already on the roadmap?

khellang:
@MatthiasJost It's already sitting on the [dev-v2 branch](https://github.com/serilog/serilog/tree/dev-v2) :smile: I'm not sure what @nblumhardt has in mind when it comes to shipping it, though.

MatthiasJost:
@khellang OK cool. I am still waiting for other libraries to be migrated. It looks like I need to wait a bit. The current version is ASP.NET 6 Beta 7. I guess it would be nice to have a "pre-release" package soon. I do not know when the release is expected.

merbla:
@MatthiasJost if you are looking at ASP.NET, have you checked out the [Serilog repo/package](https://github.com/serilog/serilog-framework-logging) that will fit in with Microsoft.Framework.Logging

MatthiasJost:
That looks very interesting. Saw that Framework.Logging once. Probably very interesting for me. Thank you @merbla.

merbla:
@MatthiasJost an updated set of [packages](https://www.nuget.org/packages/Serilog.Framework.Logging/) are available.  Can this issue be closed?

birarich:
I closed the issue...

But still we have to wait for serilog to run on dnx core 5.o?

nblumhardt:
@birarich Serilog already runs on .NET Core 5.0 - just install the _Serilog_ package and you're good to go!

Its the (externally-shipped) sinks that need updating - the `dev-v2` branch is working towards providing all the APIs that the sinks use on all platforms that support them, so you will likely need to recompile any sinks you need from source.

Notwithstanding any major changes from .NET/ASP.NET (!!!) it is probably time for us to get the pre-release v2 package tidied up and pushed to NuGet so the sinks can start to be re-targeted too. Let's leave this issue open as a way to track it:

 - [ ] `PeriodicBatchingSink` for .NET Core (needs an issue#)
 - [ ] `ILogger` generic overloads and public `Logger` abstract base class with template method for implementors
 - [ ] Check/finalize v2 prerelease build numbering
 - [ ] Publish from _AppVeyor_

Anything else?

birarich:
When I use serilog to write log into file it gives me error saying u cannot
write in file... as file method is obsolute.....

nblumhardt:
Thanks - these sinks are not yet supported on core. Added an item to the list above :-)

[...]

PR #557, opened 2015-10-31
.NET Core/v2 updates

nblumhardt:
Addresses items from #535 :

 - Reenable ignored tests
 - `dotnet` TFM
 - Update tooling to `beta8`
 - `PeriodicBatchingSink`
 - Merged other changes from serilog/serilog, cleanup, fixes

Sorry about the sprawling commit, in hindsight there could be better times to update the copyright headers ;-)


nblumhardt:
Struggling with CI, will hope to have this in a mergeable state tomorrow.

caioproiete:
Hey @nblumhardt, any chance on revisiting this: "Disallow ':' in custom setting prefixes so that ambiguities like the possibility of using "serilog:write-to" as a prefix are not allowed." ?

I understand the reason for disallowing, but think the trade-off is more expensive than the benefit, and would prefer to keep the risk of using a wrong prefix instead of losing the ability to use `:` in the prefix.

The `:` is pretty common on config files when people want to introduce a prefix to make appSettings keys more unique (not just a Serilog thing).

``` 
<appSettings>
  <add key="myApp:baseUri" value="https://myapp:86" />
  <!-- lots of other settings using the prefix "myApp:" -->

  <add key="myApp:serilog:minimum-level" value="Verbose" />

  <add key="myApp:serilog:write-to:RollingFile.restrictedToMinimumLevel" value="Verbose" />
  <add key="myApp:serilog:write-to:RollingFile.pathFormat" value="..\log\myApp-{Date}.log" />

  <add key="myApp:serilog:using" value="Serilog.Sinks.Seq" />
  <add key="myApp:serilog:write-to:Seq.restrictedToMinimumLevel" value="Verbose" />
  <add key="myApp:serilog:write-to:Seq.serverUrl" value="http://localhost:5341" />
  <add key="myApp:serilog:write-to:Seq.bufferBaseFilename" value="..\log\seq\myApp.seq.buffered.log" />
</appSettings>
```

Not a big deal, but I'd like to be able to continue doing the above, where everything can share the same prefix `myApp:`.

nblumhardt:
Hey Caio - definitely able to revisit. I'm cautious of losing the (future) ability to validate config like this:

```xml
  <add key="myApp:serilog:minimum-level" value="Verbose" />
  <add key="myApp:other:serilog:minimum-level" value="Verbose" />
```

and the possible bugs people will raise when trying to use nested custom prefixes.

Since the API doesn't know the complete set of keys, it seems like it would be hard to make the implementation smart enough to deal with these cases.

For the benefit of slightly more naming flexibility it doesn't seem like it's worth introducing the ambiguity.

If the goal is just to prefix `serilog:` with something else, perhaps we could make a slight tweak to the API? I.e. instead of accepting a prefix that replaces `serilog:`, we could accept a prefix that is added to it?

Under this scheme your example would look like:

```csharp
.WriteTo.AppSettings(prefix: "myApp")
```

and the default value of the `prefix` argument would be `null`. We'd then be able to disallow `:` and `serilog` as custom prefixes to prevent ambiguity.

Hope this makes sense. What do you think?

caioproiete:
Thanks @nblumhardt.

Yes, I like that. In fact, that is what [my first implementation](https://github.com/caioproiete/serilog/commit/ca0e74117457ee4d7187488894d79a351047228f#diff-b5d4dad5af238ea116667a8b184e6dbdR23) was doing before we went with the idea of making the parameter have a default value (`serilog`).

I guess the next question now is wether to make the `:` mandatory before `serilog` or not.
```
.WriteTo.AppSettings(prefix: "myApp")
```
vs
```
.WriteTo.AppSettings(prefix: "myApp:")
```

I def. prefer the first one, but that means we could never have something like this (as your PR), as you would always need the `:` before `serilog`
```
<add key="myApp-serilog:minimum-level" value="Verbose" />
```

It is not too bad, but it is a little bit confusing that if you want a custom prefix, you must end with ":", otherwise, start with "serilog", which is why I went with the second approach on [my first implementation](https://github.com/caioproiete/serilog/commit/ca0e74117457ee4d7187488894d79a351047228f#diff-b5d4dad5af238ea116667a8b184e6dbdR23)

nblumhardt:
Just commited this version: `.WriteTo.AppSettings(prefix: "myApp")`, seems to work nicely (though it ain't my day for builds!). Hopefully the best of both worlds........... Feedback appreciated :-)

khellang:
Nice @nblumhardt! :shipit:

nblumhardt:
Thanks - the process of working through this gave me even more admiration for your determination in getting the first cut done! Still lots of rough edges in the tooling it seems (though much better than six months ago :-))

I've started to think we need to have a more explicit discussion on here about our compatibility goals; I made a note in the corresponding ticket, but still have some thoughts to gather.

In a nutshell, it will be _very_ hard to ship an optimal v2 with any binary compatibility whatsoever. But, if we're not targeting any binary compatibility, we should be much more aggressive in what we improve - since we'll probably never have another opportunity.

The big concerns for me are `ILogger`, `Log` and `LogContext`, i.e. the types used by shared library code. Organisations will have a lot of this, and it's a pain to update. Configuration code is probably not such a big deal, as it can be easily updated app-by-app.

PR #562, opened 2015-11-02
use silent logger when creating a logger and there are no sinks in the configuration

### LINKED ISSUE #560: An unconfigured logger should be "off"
It's surprising that instead of an "off" logger, the following:

```csharp
new LoggerConfiguration().CreateLogger()
```

returns an `Information`-level logger with no sinks. The performance of this is not comparable to a truly disabled logger.

If no sinks are configured for a logger configuration, an instance of the (internal) `NullLogger` class should be returned from `CreateLogger()`.

horsdal:
Fixes #560 

I could not find the `NullLogger` mentioned in the issue, but I figured the `SilentLogger` was the one to use instead.

### DIFF
--- a/src/Serilog/LoggerConfiguration.cs
+++ b/src/Serilog/LoggerConfiguration.cs
@@ -130,6 +130,9 @@ namespace Serilog
                     disposable.Dispose();
             };
 
+            if (!_logEventSinks.Any())
+                return new SilentLogger();
+
             var sink = new SafeAggregateSink(_logEventSinks);
             
             if (_filters.Any())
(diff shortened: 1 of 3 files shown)


src/Serilog/LoggerConfiguration.cs
```diff
@@ -130,6 +130,9 @@ public ILogger CreateLogger()
                     disposable.Dispose();
             };
 
+            if (!_logEventSinks.Any())
```
merbla: Can this can be moved prior to the dispose setup? e.g. ` Action dispose = () =>`

PR #563, opened 2015-11-03
Remove redundant SafeAggregateSink from filtered sink pipeline

jim-short:


nblumhardt:
Hi Jim, thanks for the PR. The `SafeAggregateSink` at this point provides exception suppression that the filtering sink does not. Thanks again for taking the time to send this though - much appreciated.

jim-short:
Hi Nicholas, thanks for the feedback and good point. Is it worth considering suppressing in Logger.Dispatch?

nblumhardt:
Hi Jim - this is interesting, it looks like we should. Do these propagate currently?

https://github.com/serilog/serilog/wiki/Reliability#enrichers indicates that exceptions from enrichers should be suppressed.

jim-short:
Hi Nicholas, I see the enricher exceptions being logged. The motivation behind the PR was that a SafeAggregateSink with a single sink was being inserted into the logging pipeline in LoggerConfiguration.CreateLogger for the sole purpose of suppressing exeptions from the pipeline, rather than putting a try..catch around the _sink.Emit in Logger.Dispatch which would make it clear that no exceptions could be thrown by the Dispatch method. The benefit of the LoggerConfiguration.CreateLogger inserting the exception handler into the pipeline does mean that developers can create a Logger without that handler if IO errors were important to the caller for some reason. It is not a big issue by any means, just a niggle really, so thanks for taking a look.

Issue #566, opened 2015-11-04
More obvious close/flush facility when using static Log class


nblumhardt:
To close and flush the static log class in v1 looks like:

```csharp
(Log.Logger as IDisposable)?.Dispose();
```

It's terse - but that's about all that can be said about it.

It also fails to reassign `Log.Logger` to an "off" or non-disposed logger instance in case further logging does occur.

This ticket proposes:

```csharp
Log.CloseAndFlush();
```

Internally this would make the dispose call as well as reassign the logger, along the lines of:

```csharp
var toClose = Logger;
Logger = new NullLogger();
(Logger as IDisposable)?.Dispose();
```

(Atomic CAS to substitute the logger instances would be a plus.)

nblumhardt:
This has been implemented via the PR and is now in `dev` as `Log.CloseAndFlush()`.

PR #568, opened 2015-11-06
#566 - Try and set the log source if it's specified

johnduhart:
This implements #566. I wrote Unit tests for this, however I was not able to run them. I'm assuming this has to do with the outstanding #557


nblumhardt:
Nicely done, thanks John!

Issue #567, opened 2015-11-05
Make JsonFormatter.WriteLiteral virtual


inosik:
I'm currently writing a sink for [LiteDB](http://www.litedb.org/) and need to write a custom `JsonFormatter` class, which would write dates and some other types in a [special format](https://github.com/mbdavid/LiteDB/wiki/Data-Structure#json-extended).

I adapted the implementation of the [`MongoDBJsonFormatter` class][mongodb], and the `Timestamp`-Property of my log events is written correctly now.

The problem is, that dates in the `Properties` sub object cannot be written in that format, as in the [`JsonFormatter.WriteLiteral` method][writeliteral] there is already a writer for `DateTime`. The next possible extension point is the [`JsonFormatter.WriteLiteralValue` method][writeliteralvalue], where I don't get into for `DateTime`, as it already [has been handled][too-late].

One (probably) possible solution would be to override the implementations for [`LogEventPropertyValue` classes][valueclasses] and handle them in the `WriteJsonProperty` method, but I don't think that this is a good approach.

The `MongoDBJsonFormatter` suffers the same problem. It writes the `Timestamp` in the databases format, but dates in `Properties` get written as strings.

[mongodb]: https://github.com/serilog/serilog-sinks-mongodb/blob/master/src/Serilog.Sinks.MongoDB/Sinks/MongoDB/MongoDBJsonFormatter.cs
[writeliteral]: https://github.com/serilog/serilog/blob/dev/src/Serilog/Formatting/Json/JsonFormatter.cs#L327
[writeliteralvalue]: https://github.com/serilog/serilog/blob/dev/src/Serilog/Formatting/Json/JsonFormatter.cs#L322
[too-late]: https://github.com/serilog/serilog/blob/dev/src/Serilog/Formatting/Json/JsonFormatter.cs#L339
[valueclasses]: https://github.com/serilog/serilog/blob/dev/src/Serilog/Formatting/Json/JsonFormatter.cs#L91-L94

nblumhardt:
Thanks for the report and the detailed analysis Ilja. Extensibility of `JsonFormatter` has been a challenge for some time.

I think `WriteLiteral()` could change as this class evolves; would a method like:

```csharp
protected void AddWriter(Type type, Action<object, TextWriter> write)
{
    _literalWriters[type] = (v, q, w) => write(v, w);
}
```

be another option? This could be called in the constructor of your formatter to replace the default writers for `DateTime` and `DateTimeOffset`.

[...]

PR #572, opened 2015-11-09
Allow to add and override custom writers

inosik:
This is my proposed fix for #567. If you want me to change something, just tell me. Otherwise, :shipit:


nblumhardt:
Thanks, LGTM.

Issue #574, opened 2015-11-10
Support fixed-length level output for OCD goodness


iamkoch:
I spotted and subsequently tweeted Nick regarding seeing INF, DBG and other shortened levels in output logs to be told that it's only in certain sinks at the moment and that bringing this into the main codebase might be nice...

...so that's what I did. PR coming. Happy for feedback re approach.

I've opted for a dumber inclusion of "Lvl" as a template variable instead of allowing Level:3 / Level:4 or similar as I felt it had the smallest impact on the codebase and limited the wishlist for custom words or text, different interpretations of whether it should be DEB, DEBG, DBUG etc.

nblumhardt:
Completed via the linked PR :tada:

PR #575, opened 2015-11-10
Add support for fixed-length levels

iamkoch:
Addresses #574 (which I raised)

### DIFF
--- a/src/Serilog/Events/FixedLengthLogEventLevel.cs
+++ b/src/Serilog/Events/FixedLengthLogEventLevel.cs
@@ -1,0 +1,38 @@
+﻿namespace Serilog.Events
+{
+    /// <summary>
+    /// A fixed-length representation of <see cref="LogEventLevel"/>
+    /// </summary>
+    public enum FixedLengthLogEventLevel
+    {
+        /// <summary>
+        /// Shortened version of <see cref="LogEventLevel.Verbose"/>
+        /// </summary>
+        VRB,
+
+        /// <summary>
+        /// Shortened version of <see cref="LogEventLevel.Debug"/>
+        /// </summary>
+        DBG,
+
+        /// <summary>
+        /// Shortened version of <see cref="LogEventLevel.Information"/>
+        /// </summary>
+        INF,
+
+        /// <summary>
+        /// Shortened version of <see cref="LogEventLevel.Warning"/>
+        /// </summary>
+        WRN,
+
+        /// <summary>
+        /// Shortened version of <see cref="LogEventLevel.Error"/>
+        /// </summary>
+        ERR,
+
+        /// <summary>
+        /// Shortened version of <see cref="LogEventLevel.Fatal"/>
+        /// </summary>
+        FTL
+    }
+}
--- a/src/Serilog/Events/LogEvent-net40.cs
+++ b/src/Serilog/Events/LogEvent-net40.cs
@@ -67,6 +67,14 @@ namespace Serilog.Events
         }
 
         /// <summary>
+        /// A shortened, fixed-length representation of the event level.
+        /// </summary>
+        public FixedLengthLogEventLevel FixedLengthLevel
+        {
+            get { return (FixedLengthLogEventLevel) (int) _level; }
+        }
+
+        /// <summary>
         /// The message template describing the event.
         /// </summary>
         public MessageTemplate MessageTemplate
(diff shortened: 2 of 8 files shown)


nblumhardt:
Thanks Antony, looking good. I think this one is a candidate for v2, which `dev` is about to become. I will circle back to it shortly once the target branch is ready.

iamkoch:
Great stuff! Was actually very easy to implement. Will keep an eye out for any other issues that I can tackle that are a bit meatier.

iamkoch:
Hi @nblumhardt 

Any update on this one being included in the release?

Thanks
Ant

nblumhardt:
Thanks for the note - still planning to come back around to this one but just getting the v2 fundamentals/feature parity sorted out.

Looking back over this, perhaps using a width specifier would be nicer in the long run? I.e:

`"{Level,4}"

I'm not sure about the impact you mention however - I need a chance to look into the trade-offs. Any thoughts?

iamkoch:
I'll take a look at specifying the width. I think my comment was based on the relative ease of just introducing a new token (Lvl) as opposed to providing intelligence around the existing one (Level,3)

I'll spike it out on a separate branch and let you know what it looks like.

nblumhardt:
That sounds great. It'll be a very popular feature I think, so worth exploring the options :-)

iamkoch:
Hi Nick

Finally got round to this after some dotnet core issues on my machine. Should have something in the next few days.

iamkoch:
The existing behaviour around the "{Name},{Alignment}" seems to have a different intent to the one mentioned above with {Level,3}, namely that the writer will pad the output accordingly should the length of the message be shorter than the provided alignment value.

This means that in its current guise, something like {Name,10} would make `"Mike"` become, depending on direction,  `"      Mike"`. It doesn't, however, cause {Name,3} to turn "Mike" into "Mik" or similar. I'm not sure whether this makes sense, though, as if an alignment value is provided, e.g. 10, the output string can be greater than 10, never less than 10, but not exactly 10 unless by coincidence.

With the above in mind, I propose either:

1. Change the the alignment functionality to cause the property to be precisely that length, i.e. substring when greater than the provided value
2. Use a different mechanism than {Level,3} as it has different behaviour by virtue of trimming down a longer value as it's a 'special' case
3. Stick with my previous fixed-width, only 3 character option of {Lvl}

I think opting for 1 is dependent on the original intent when the feature was first added. 3 is the original feature I developed, however it's prescriptive and inflexible for other lengths. 2 is an option, but it means coming up with some other syntax that's akin to {Level,3} but isn't {Level,3} because it means something else.

Thoughts?

nblumhardt:
Hey, thanks for picking this back up!

I know there's a slight warping of semantics here, but it should be possible to just special-case the property name `Level` here:

https://github.com/serilog/serilog/blob/dev/src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs#L81

I.e. if the token's property name is exactly `"Level"`, instead of letting the token render itself, just call another routine passing the width (and possibly the `"u"` uppercase or `"w"` lowercase format specifiers) through.

The routine could special case each width, i.e. if the width is `3`, then get the text from a dictionary mapping levels to `Vrb`, `Dbg`, etc.

For any width that doesn't require special treatment (e.g. for width `1`, or width `7`, the full level names could just be trimmed or padded.

Ideally:

`{Level,3}` should produce `Dbg`
`{Level,3:u}` should produce `DBG`
`{Level,3:w}` should produce `dbg`

...and so on. I think there are pretty typical abbreviations for 2, 3, 4 and maybe 5 chars. Here's what the literate sink uses: https://github.com/serilog/serilog-sinks-literate/blob/dev/src/Serilog.Sinks.Literate/Sinks/Literate/LiterateConsoleSink.cs#L60 - may be some better precedent in Node.js, log4j and so-on, haven't looked widely.

How's this sound? Sorry about the brain-dump format :-)

iamkoch:
Heya

This sounds fine as you're OK with it.

I've been toying around with some concepts, however I was unsure whether to make this a more generic solution allowing for other core properties to apply the same logic, or to just hard code it explicitly to level.

I figured I'd use lengths 1-4 and 5 and above would trim or pad appropriately. 7 would return "Debug  " for example.

I'll finish up based on the above and create a new PR, closing this one as I do.

Thanks for the feedback.
Ant

iamkoch:
Closing as new PR raised

PR #581, opened 2015-11-13
Ability to set the setting delimiter to make the it possible to config serilog via ServiceConfiguration.Cloud.cscfg

johnkattenhorn:
It would be great honour if this PR was accepted but the main reason for adding the PR for other to see what I've done and see if there's any room for improvement.

### DIFF
--- a/GitVersionConfig.yaml
+++ b/GitVersionConfig.yaml
@@ -1,1 +1,3 @@
+assembly-versioning-scheme: MajorMinorPatch
+mode: ContinuousDeployment
 branches: {}
--- a/GitVersionConfig.yaml
+++ b/GitVersionConfig.yaml
@@ -1,0 +1,1 @@
+branches: {}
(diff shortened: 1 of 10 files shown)


nblumhardt:
Hi John, thanks for sending this. Sorry about the slow response, a bit swamped these last few days. I'll take a look this evening. Cheers!

johnkattenhorn:
I've found it's actually not a great deal of use on its own because I had completely overlooked the fact that the AppSettings will still go after the ConfigurationManager.AppSettings (I've got too used to using CloudConfigurationManager.GetSetting).

I've checked and there isn't an equivalent of getting all the settings in the CloudManager that I could find and it would've be a good idea either as that ties Serilog to Azure Configuration Manager even though it falls back to app.config.

The two settings we really wanted was logging level and apiKey for seq, logging level we can code for and Octopus takes care during deployment of setting the apiKey at least initially so that will have to do.

it was a good training exercise though :-)

nblumhardt:
Thanks for the follow-up. Sounds like configuration-in-code is the way to go. Cheers!

Issue #593, opened 2015-11-17
Package factoring


nblumhardt:
The "new .NET" has switched to much finer-grained packages as the norm. _Serilog.nupkg_ is currently the "kitchen sink"; since v2 is breaking now is the time to consider a finer-grained factoring.

* _Serilog.Abstractions_: `ILogger`, `Log`, `LogContext`, `LogEvent`, `LogEventLevel`, the various property/value classes; everything potentially used by client application code
* _Serilog_: `LoggerConfiguration`, the pipeline; depends on _Serilog.Abstractions_
* Sink packages for: file, rolling file, console, colored console, trace, observable, periodic-batching
* Enricher packages: environment, ...
* Packages for AppSettings and so-on

The advantage of doing this is reducing churn (no need to version core assemblies as sinks evolve, etc.) and prevention of bloat, especially dependency bloat as will be the case with v2 as it stands currently.

There's a lot of work represented in this, but better to have the discussion while there's an opportunity to make the change.

merbla:
+1.

The file, rolling file, console etc. lived in core for convenience.  With the next version of .Net it seems the approach is to break these down. 

There is a fine line between 1000s of packages and decoupled design.

We need to be wary over the long term of Package **Bloat** vs Package **Explosion** :boom: however I think the benefits are there.

khellang:
I like this. Especially the Serilog/Serilog.Abstractions split. Most libraries in an application only care about the `ILogger` and friends :+1:

caioproiete:
:+1:

merbla:
Might have gone overboard but is this the idea?
[XML Gist for Draw.IO](https://gist.github.com/merbla/5961d474ec90cba3df9b)
![](https://raw.githubusercontent.com/merbla/images/master/serilog/SerilogV2.png)

khellang:
I'd move the types listed below in Serilog.Abstractions and flip things so enrichers, configuration and sink packages depend on that instead.

- `ILogEventSink`
- `PeriodicBatchingSink`
- `ILoggerSettings`
- `ILogEventEnricher`
- `ILogEventPropertyFactory`
- `LogEventProperty`
- `LogEventPropertyValue`

I might've forgotten some...

merbla:
Yep thanks @khellang.  Was initially thinking about packages rather than the actual class structure.

Updated...

![](https://raw.githubusercontent.com/merbla/images/master/serilog/Serilog%20v2.0.1.png)

[...]

PR #602, opened 2015-12-06
Patching Factoring - New Abstractions Project

merbla:
Initial PR related to #593 for the following 

- Test AppVeyor Build
- Gather feed back from @serilog/core

Not ready to merge quite yet.


merbla:
mmm is there a race condition in the tests?  @khellang how did you fix the last one?  just ran again?

[404](https://ci.appveyor.com/project/serilog/serilog/build/404) - :fire:
[403](https://ci.appveyor.com/project/serilog/serilog/build/403) - :ok_hand:
[402](https://ci.appveyor.com/project/serilog/serilog/build/402) - :fire:
[401](https://ci.appveyor.com/project/serilog/serilog/build/401) - :fire:

khellang:
I think @nblumhardt got it passing by just running it again. It's flaky to say the least.

nblumhardt:
I think we need another `ConfigureAwait(false)` in `PeriodicBatchingSink` somewhere; trying to find a chance to check out that theory.... :-)

merbla:
ping @serilog/core.. review time pls

I think `src/Serilog.Abstractions/project.json` may get lighter once other cruft is moved in/out.

nblumhardt:
Looks great @merbla - feels like the right direction :+1:

* `MessageTemplate.PostionalProperties` and `.NamedProperties` - exposing these turns the otherwise immutable `MessageTemplate` type into a subtly-mutable one; any alternatives here?
* `IMessageTemplateParser` - any way we can avoid exposing this/moving it to abstractions?

Cheers! :-)

merbla:
Will have a look

src/Serilog.Abstractions/ILogger.cs
```diff
@@ -48,7 +48,7 @@ public interface ILogger
         /// </summary>
         /// <returns>A logger that will enrich log events as specified.</returns>
         ILogger ForContext(string propertyName, object value, bool destructureObjects = false);
-
+
```
khellang: :fire:

src/Serilog/project.json
```diff
@@ -6,69 +6,72 @@
...
+        "System.Diagnostics.TraceSource": "4.0.0-beta-23516",
+        "System.IO": "4.0.11-beta-23516",
+        "System.IO.FileSystem": "4.0.1-beta-23516",
+        "System.Linq": "4.0.1-beta-23516",
+        "System.Text.RegularExpressions": "4.0.11-beta-23516",
+        "System.Threading.Thread": "4.0.0-beta-23516"
+      }
     }
+  },
+  "dependencies": {
```
khellang: Can we move this above the `frameworks`? It's a bit hidden down here.

khellang:
LGTM, just a couple of nits :shipit:

merbla:
@nblumhardt @khellang thoughts on this approach?  

- Aiming to keep implementation internal etc.
- Is `IMessageTemplate` the way to go?

nblumhardt:
Hey @merbla I'm not sure I'm on the same page, it's a bit tricky to piece together from the diff, sorry - guessing this is obvious but:

 * why do we need `IMessageTemplate`?
 * why does the message template parsing infrastructure need to go into Abstractions?

Sorry about the possibly clueless questions, thanks for pushing forward on this!

merbla:
No worries @nblumhardt, really just feeling this one out.  I am sure there is more to come.

I have not gone for wholesale changes however we may need a few in that we have some interfaces that have dependencies on objects that rely on some internals of core

e.g. Currently the following is the landscape.

Ignoring `IMessageTemplate`, `LogEvent` has dependencies on interfaces and implementation that really should be in **Core**.  Similarly there are some interfaces that depend on `LogEvent` making this split troublesome.

I will have another think and ping you back when I have more.

merbla:
@nblumhardt thoughts on an `ILogEvent`? Seems a little pointless as I can't see the need for other implementations.

merbla:
Getting closer...

nblumhardt:
Aha! I'm on the same page now @merbla - thanks for persevering.

I like where things have landed, but I am wondering if:

a) `SelfLog` belongs in Abstractions anyway, since most sinks use it, and
b) Instead of an interface, `IMessageTemplate` could be a class, called `MessageTemplate` and the optimised/caching version in core could be a subclass called `CachedMessageTemplate` and this would only be used by the few parts of the code requiring the optimised properties. E.g.:

```csharp
public class MessageTemplate
{
    readonly string _text;
    readonly MessageTemplateToken[] _tokens;

    public MessageTemplate(IEnumerable<MessageTemplateToken> tokens)
        : this(string.Join("", tokens), tokens)
    {
    }

    public MessageTemplate(string text, IEnumerable<MessageTemplateToken> tokens)
    {
        if (text == null) throw new ArgumentNullException(nameof(text));
        if (tokens == null) throw new ArgumentNullException(nameof(tokens));

        _text = text;
        _tokens = tokens.ToArray();
    }

    public string Text
    {
        get { return _text; }
    }

    public override string ToString()
    {
        return Text;
    }

    public IEnumerable<MessageTemplateToken> Tokens
    {
        get { return _tokens; }
    }

    public string Render(IPropertyDictionary properties, IFormatProvider formatProvider = null)
    {
        var writer = new StringWriter(formatProvider);
        Render(properties, writer, formatProvider);
        return writer.ToString();
    }

    public void Render(IPropertyDictionary properties, TextWriter output, IFormatProvider formatProvider = null)
    {
        foreach (var token in _tokens)
        {
            token.Render(properties, output, formatProvider);
        }
    }
}

class CachedMessageTemplate : MessageTemplate
{
    readonly PropertyToken[] _positionalProperties;
    readonly PropertyToken[] _namedProperties;

    public CachedMessageTemplate(string text, IEnumerable<MessageTemplateToken> tokens)
        : base(text, tokens)
    {
        var propertyTokens = tokens.OfType<PropertyToken>().ToArray();
        if (propertyTokens.Length != 0)
        {
            var allPositional = true;
            var anyPositional = false;
            foreach (var propertyToken in propertyTokens)
            {
                if (propertyToken.IsPositional)
                    anyPositional = true;
                else
                    allPositional = false;
            }

            if (allPositional)
            {
                _positionalProperties = propertyTokens;
            }
            else
            {
                if (anyPositional)
                    SelfLog.WriteLine("Message template is malformed: {0}", text);

                _namedProperties = propertyTokens;
            }
        }
    }

    public PropertyToken[] NamedProperties
    {
        get { return _namedProperties; }
    }

    public PropertyToken[] PositionalProperties
    {
        get { return _positionalProperties; }
    }
}
```

This would totally avoid the expensive interface/virtual dispatch, and potentially allow `CachedMessageTemplate` to remain internal (not sure but it's a possibility).

What do you think?

nblumhardt:
@merbla nice diagrams BTW - what creates those? :-)

merbla:
@nblumhardt will bang it out and let you know.

nblumhardt:
@merbla I have a bit of time today, wondering if I might take on merging this and try out the `MessageTemplate`/`CachedMessageTemplate` transform in the process; what do you think? :-) Cheers!

merbla:
Go for it!

nblumhardt:
Have some time to spare, will pick this up and see if I can fix the build issue "on the way in" :-)

nblumhardt:
This could be a really sucky time to come out with a comment like this ;-) ... but now after having dug into the changeset I'm wondering how much we're actually going to win when the dust settles.

To support the full scope of what client code and sinks will want to use, we'll at least need to bring across:

 * `Log` and `LogContext` and their dependencies (used by client code)
 * `TextToken` and `PropertyToken` (used by the console sinks and others)
 * `JsonFormatter` (used by many sinks)
 * `OutputProperties` and `MessageTemplateTextFormatter` (hazy recollection of sinks needing this)
 * `SelfLog` (used by many sinks)

At this point _Abstractions_ may not be the best name - there's a lot of implementation code mixed in there as well.

After doing that, the only things left in the _Serilog_ package are `LoggerConfiguration`, the core pipeline (changes rarely), enrichers, destructuring code, app settings support and the sinks.

Once we move the sinks, app settings support, and potentially the enrichers to their own packages, there won't be much to justify shipping _Serilog_ and _Serilog.Abstractions_ separately.

We're also left with some awkward structuring, like the complete infrastructure behind message templates being in _Abstractions_, but the parser itself sitting lonely in core.

Would I be crazy to suggest we put this PR on hold, and look at pulling out the sinks etc. as a first step, after all?

merbla:
:hear_no_evil:  I think you are right.

Put this one on ice and start on the sinks and enrichers.

nblumhardt:
@merbla I'll close this one for now since a re-take is probably going to be "from scratch" given all the work you've done on factoring since this PR was opened.

PR #608, opened 2015-12-13
Attempt to fix intermittent hang during CoreCLR test runs

nblumhardt:
`Task.Wait()` in this instance may cause marshalling back onto the originating synchronisation context, leading to a deadlock. If this successfully resolves the intermittent test failure we should consider moving this change back to 1.5 as well.

See also: http://stackoverflow.com/questions/28305968/use-task-run-in-synchronous-method-to-avoid-deadlock-waiting-on-async-method

CC @merbla @khellang

### DIFF
--- a/src/Serilog/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
+++ b/src/Serilog/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
@@ -21,6 +21,7 @@ using System.Threading.Tasks;
 using Serilog.Core;
 using Serilog.Debugging;
 using Serilog.Events;
+using System.Threading;
 
 #if !NO_TIMER
 using System.Threading;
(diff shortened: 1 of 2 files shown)


nblumhardt:
It would appear that's the same failure. I will scour the code for any more of these...

nblumhardt:
Rebase/squash. I'm not 100% confident either of these help, but I've also run three consecutive builds without seeing the issue; may be time to merge and move on, @serilog/core ?

nblumhardt:
Any thoughts on this one folks?

khellang:
If this sorts things out, it LGTM! :shipit:

khellang:
What's the process to backport this to 1.x? Just cherry-pick it into master?

nblumhardt:
I think so. It's hard to be sure if it is a problem with Serilog, .NET on CoreCLR or with the test runner, but some defensiveness couldn't hurt :+1:

Issue #609, opened 2015-12-14
Support for PeriodicBatching in Profile 259


ravensorb:
Any plans to support PeriodicBatching in Profile 259?

nblumhardt:
Howdy - no current plans, but the work done in `dev` for v2 should make this easier. I think the availability of `ConcurrentQueue<T>` might be the main blocker.

ravensorb:
Any thoughts to breaking out the interfaces from the actual implementation so that a basic library could be included in any project and then the actual implementation be included in the apps?  That would allow you to effectively remove the platform dependency issue all together?

merbla:
hey @ravensorb some of the work as apart of #593 will hopefully help this as well.

ravensorb:
Looks great!  Anything I can do to help?

nblumhardt:
Currently getting the v2 ship-shape is priority 1 - if there's anything in https://github.com/serilog/serilog/labels/v2 that takes your fancy please let us know! :-) :+1:

nblumhardt:
2.x should now have this covered.

Issue #589, opened 2015-11-17
Enable strong naming for CoreCLR builds


nblumhardt:
Resolve split identity problems - ideally existing code dependent on ILogger will still run on CLR (requires strong name) but SN is a second-class citizen on CoreCLR. We don't want the same assembly to be present on the same (compatible) platforms with both strong and non-strong names. From #535.

nblumhardt:
I think the right approach to this is to strong-name the CoreCLR binaries, and either disable or rewrite each of the tests that relies on InternalsVisbleTo.

PR #616, opened 2015-12-21
Replaced Profile259 with dotnet5.1

khellang:
This replaces the old `Profile259` with `dotnet5.1`, which is the way forward.

`dotnet5.1` is the lowest System.Runtime-based TFM, roughly equivalent to `Profile259`, and should support everything back to WP Silverlight 8.0. See the [Standard Platform document](https://github.com/dotnet/corefx/blob/50c66ffcf62be7683a756df14a2dbc75720ae263/Documentation/project-docs/standard-platform.md#mapping-the-net-platform-standard-to-platforms) for more details. BTW, `dotnet5.1` will be `netstandard1.0` (there's an off-by-one error) and that's what it's called in the document.

I also (mostly) moved away from #ifdef'ing on platform and moved to features instead. I think this makes it clearer why things are #ifdef'ed and what should be moved to separate packages. It would be really nice if Serilog core would just be `dotnet5.1` only. The difference is mostly `System.Diagnostics.Process` and `System.IO.FileSystem`, which aren't available until `dotnet5.4`. There's also the whole `AppDomain`/`Timer` stuff in the `PeriodicBatchingSink` :disappointed: 

This should also unlock the following on the `Profile259` platforms:

 - `DiagnosticTraceSink`
 - `ConsoleSink`
 - `ColoredConsoleSink`
 - `ThreadIdEnricher`

### DIFF
--- a/src/Serilog/Enrichers/EnvironmentUserNameEnricher.cs
+++ b/src/Serilog/Enrichers/EnvironmentUserNameEnricher.cs
@@ -1,18 +1,18 @@
 ﻿// Copyright 2013-2015 Serilog Contributors
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !PROFILE259
+#if !DOTNET5_1
 
 using System;
 using Serilog.Core;
--- a/src/Serilog/Properties/InternalsVisibleTo.cs
+++ b/src/Serilog/Properties/InternalsVisibleTo.cs
@@ -1,6 +1,6 @@
 ﻿using System.Runtime.CompilerServices;
 
-#if DOTNET5_4
+#if DOTNET5_4 || DOTNET5_1
 [assembly: InternalsVisibleTo("Serilog.Tests")]
 #else
 [assembly: InternalsVisibleTo("Serilog.Tests, PublicKey=" +
--- a/src/Serilog/Context/ImmutableStack.cs
+++ b/src/Serilog/Context/ImmutableStack.cs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !PROFILE259 && !DOTNET5_4
+#if LOGCONTEXT
 using System;
 using System.Collections.Generic;
 using System.Runtime.Serialization;
--- a/src/Serilog/Context/LogContext.cs
+++ b/src/Serilog/Context/LogContext.cs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !PROFILE259 && !DOTNET5_4
+#if LOGCONTEXT
 using System;
 using System.Runtime.Remoting.Messaging;
 using Serilog.Core;
--- a/src/Serilog/Enrichers/LogContextEnricher.cs
+++ b/src/Serilog/Enrichers/LogContextEnricher.cs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !PROFILE259 && !DOTNET5_4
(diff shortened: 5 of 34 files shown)


khellang:
Who knows - maybe this'll fix #609?

khellang:
Ah, no it won't. It's missing `ConcurrentQueue` :disappointed:

nblumhardt:
Wow, the `#ifdef` changes make this _much_ nicer. :+1: 

Finally expecting a solid day to review PRs and help with v2 this weekend, will take a deeper look then unless I'm beaten to it, but looks great.

src/Serilog/Enrichers/EnvironmentUserNameEnricher.cs
```diff
@@ -1,18 +1,18 @@
...
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !PROFILE259
+#if !DOTNET5_1
```
khellang: I couldn't come up with a good "feature" name for this. It's basically some `Environment` methods/properties that are missing in `dotnet5.1`...

src/Serilog/Properties/InternalsVisibleTo.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.Runtime.CompilerServices;
 
-#if DOTNET5_4
+#if DOTNET5_4 || DOTNET5_1
```
khellang: `#if NO_SIGNING`?

nblumhardt: I'm thinking we'll ditch the unsigned version as part of #589 - probably not worth spending time on right now...

kolektiv:
Is it expected that a Profile259 PCL would be able to consume this? Trying beta450 with a Profile259 PCL (in F#) NuGet claims that there are no compatible resources in the package, so despite what the platform document claims, it doesn't appear to work "at the moment". Is this known/expected? F# is in a tricky place as it will lag quite a bit behind, and PCL will be the state of play there for a while longer I suspect, but it would be a shame if newer versions of libraries like Serilog stopped being accessible (although of course that's a valid choice for the Serilog team to make).

Just trying to get my head around it all!

kolektiv:
For the record, NuGet error:

> Could not install package 'Serilog 2.0.0-beta-450'. You are trying to install this package into a project that targets '.NETPortable,Version=v4.5,Profile=Profile259', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

khellang:
> Is it expected that a Profile259 PCL would be able to consume this?

Yes :smile:

> it doesn't appear to work "at the moment"

Yeah, it definitely looks that way. Maybe we need to add it back for now?

> F# is in a tricky place as it will lag quite a bit behind

I don't think that matters much, since this is a NuGet concern.

kolektiv:
In light of the answer to the first bit, yeah I'm not worried about F# as there's a story that "should" work. I guess it's now a question of working out which bit of NuGet to kick. I'll try it with Paket as well actually, just to see if that works. That's the normal approach anyway for most of F# I suspect.

kolektiv:
In case he happens to be interested, Paket doesn't yet support this scenario either (mentioning @forki just to cheer him up), but I'm inclined to think it's probably not in scope for Paket yet.

khellang:
So, just spoke to @davidfowl. Apparently `dotnet5.x` can't be referenced from PCLs (as we've discovered), but `netstandard1.x` *will*. I'm not sure if it's worth the hassle to add Profile259 back. 2.0 will most probably ship with `netstandard` anyway... @nblumhardt, what do you think?

forki:
Please create an issue on paket with details.

kolektiv:
I think it can remain away from Paket right now as it sounds like a moving target which isn't well defined yet. Paket will need to support `netstandard1.x` at some point, but it doesn't sound like that's yet. I can create a long term issue to track that if you'd like?

On the Serilog side, it sounds reasonable to assume that 2.0 will be usable from F#, once netstandard support is more mature and better defined, but for now it's probably best to stick with Serilog 1.5x in the knowledge that it'll be ok "some day". Presumably the APIs aren't going to change massively from 1.5x -> 2.0?

khellang:
> Presumably the APIs aren't going to change massively from 1.5x -> 2.0?

That's right. At least for the commonly used "application APIs". The package factoring will be different, though.

The v2 issues can be tracked at https://github.com/serilog/serilog/labels/v2 :smile:

Issue #601, opened 2015-12-03
use c#6


SimonCropp:
interested in a PR that updates the code to c#6?

khellang:
I say yay :+1:

nblumhardt:
:+1: too, nice to have it in one commit rather than patchy updates as we touch each file.

khellang:
Looks like this is partly done; https://github.com/serilog/serilog/commit/34a1aaeb336b23c898a9caa662a1b37fd2229bde

nblumhardt:
I think most of this is covered in v2 now.

PR #617, opened 2015-12-22
Kill some trailing whitespace, update to C# 6

khellang:
What the title says. Closes (most of?) #601. Was cruising by anyway.

### DIFF
--- a/src/Serilog/Context/LogContext.cs
+++ b/src/Serilog/Context/LogContext.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright 2013-2015 Serilog Contributors
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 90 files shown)


src/Serilog/Context/LogContext.cs
```diff
@@ -94,8 +95,7 @@ public static IDisposable PushProperties(params ILogEventEnricher[] properties)
             var stack = GetOrCreateEnricherStack();
             var bookmark = new ContextStackBookmark(stack);
 
-            foreach (var prop in properties)
-                stack = stack.Push(prop);
+            stack = properties.Aggregate(stack, (current, prop) => current.Push(prop));
```
nblumhardt: Not sure this is idiomatic use of `Aggregate` here  - should be equivalent but maybe the original was clearer?

src/Serilog/Settings/AppSettings/AppSettingsSettings.cs
```diff
@@ -29,7 +29,6 @@ class AppSettingsSettings : ILoggerSettings
         public AppSettingsSettings(string settingPrefix = null)
         {
             _settingPrefix = settingPrefix == null ? "serilog:" : $"{settingPrefix}:serilog:";
-            System.Console.WriteLine(_settingPrefix);
```
nblumhardt: Eek! Nice catch.

src/Serilog/Context/LogContext.cs
khellang: Huh, I guess R# suggest this and I just Alt+Enter'd my way through :stuck_out_tongue_closed_eyes: I'll revert...

nblumhardt:
Looks great! Definitely think this will avoid an endless stream of small updates in other changes.

Just one comment in there, otherwise :shipit: :-)

khellang:
Alright. Rebased and removed :smile:

nblumhardt:
:+1:

Issue #618, opened 2015-12-23
anonymous types with vb.net


richardversluis:
Hi,

Great library :). 

I have a question about using anonymous types in vb.net with serilog. When I do the following:

```vb.net
Dim someInfo = New With {.Bla = "bla", .Bla2 = "bla2"}
log.Information("{@SomeInfo}", someInfo)
```

The output is:

```
12/23/2015 11:48:10 VB$AnonymousType_1`2 { Bla: "bla", Bla2: "bla2" }
```

Like the example on the serilog.net homepage I wasn't expecting the type `VB$AnonymousType_1`2` to be logged, because it's anonymous. It this assumption wrong?

nblumhardt:
Interesting - I wasn't aware VB has a different convention for naming anonymous types than C# does.

The check that makes the C# version work is at: https://github.com/serilog/serilog/blob/dev/src/Serilog/Parameters/PropertyValueConverter.cs#L158

There are various more comprehensive ways of checking, that would also work for VB, here: http://stackoverflow.com/questions/2483023/how-to-test-if-a-type-is-anonymous

This seems like a fix we can make, but needs some care to make sure things stay as quick as possible.

nblumhardt:
This looks like a good implementation of the check: https://github.com/NancyFx/Nancy/blob/master/src/Nancy/ViewEngines/Extensions.cs#L29

Pvlerick:
I'm happy to grab this if no-one is on it already.

nblumhardt:
Closed in `dev` via #622.

richardversluis:
:+1:

PR #622, opened 2016-01-06
Added function to detect anonymous types and appropriate tests

Pvlerick:
Adressing #618 

I used some of the code pointed in the defect discussion, added a function to detect anonymous types based on the type name and some tests.

I didn't add a proper test for VB - a real test using a anonymous type coming from a VB library - as this is a bit tedious. This could be done with [Roslyn](https://github.com/dotnet/roslyn) though but I don't think it's worth the hassle.

I nevertheless did test the new dll in a console app written in VB and it works.

### DIFF
--- a/src/Serilog/Parameters/PropertyValueConverter.cs
+++ b/src/Serilog/Parameters/PropertyValueConverter.cs
@@ -25,6 +25,8 @@ using Serilog.Policies;
 
 #if NET40
 using Serilog.Platform;
+#else
+using System.Runtime.CompilerServices;
 #endif
 
 namespace Serilog.Parameters
@@ -154,9 +156,12 @@ namespace Serilog.Parameters
 
             if (destructuring == Destructuring.Destructure)
             {
-                var typeTag = value.GetType().Name;
-                if (typeTag.Length <= 0 || !char.IsLetter(typeTag[0]))
+                var type = value.GetType();
+                var typeTag = type.Name;
+                if (typeTag.Length <= 0 || IsCompilerGeneratedType(type))
+                {
                     typeTag = null;
+                }
 
                 return new StructureValue(GetProperties(value, limiter), typeTag);
             }
@@ -213,6 +218,20 @@ namespace Serilog.Parameters
                 }
                 yield return new LogEventProperty(prop.Name, recursive.CreatePropertyValue(propValue, true));
             }
+        }
+
+#if !NET40
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+#endif
+        internal static bool IsCompilerGeneratedType(Type type)
+        {
+            var typeInfo = type.GetTypeInfo();
+            var typeName = type.Name;
+
+            //C# Anonymous types always start with "<>" and VB's start with "VB$"
+            return typeInfo.IsGenericType && typeInfo.IsSealed && typeInfo.IsNotPublic && type.Namespace == null
+                && (typeName[0] == '<'
+                    || (typeName.Length > 2 && typeName[0] == 'V' && typeName[1] == 'B' && typeName[2] == '$'));
         }
     }
 }
--- a/test/Serilog.Tests/Parameters/PropertyValueConverterTests.cs
+++ b/test/Serilog.Tests/Parameters/PropertyValueConverterTests.cs
@@ -232,5 +232,15 @@ namespace Serilog.Tests.Parameters
             var item = pv.Properties.Single();
             Assert.Equal("Item", item.Name);
         }
+
+        [Fact]
+        public void CSharpAnonymousTypesAreRecognizedWhenDestructuring()
+        {
+            var o = new { Foo = "Bar" };
+            var result = _converter.CreatePropertyValue(o, true);
+            Assert.Equal(typeof(StructureValue), result.GetType());
+            var structuredValue = (StructureValue)result;
+            Assert.Equal(null, structuredValue.TypeTag);
+        }
     }
 }
--- a/src/Serilog/Platform/TypeInfo-net40.cs
+++ b/src/Serilog/Platform/TypeInfo-net40.cs
@@ -55,6 +55,8 @@ namespace Serilog.Platform
         public IEnumerable<PropertyInfo> DeclaredProperties => Type.GetProperties();
 
         public bool IsAssignableFrom(TypeInfo targetType) => Type.IsAssignableFrom(targetType.Type);
+
+        public bool IsNotPublic => Type.IsNotPublic;
     }
 }
 #endif


test/Serilog.Tests/Parameters/PropertyValueConverterTests.cs
```diff
@@ -232,5 +232,30 @@ public void ItemPropertiesNotAreIgnoredWhenDestructuring()
...
+            Assert.Equal(typeof(StructureValue), result.GetType());
+            var structuredValue = (StructureValue)result;
+            Assert.Equal(null, structuredValue.TypeTag);
+        }
+
+        [Fact]
+        public void CSharpAnonymousTypeNameIsRecognized()
+        {
+            var typeName = "<>f__AnonymousType0`1";
+            //var typeName = "VB$AnonymousType_0`2";
```
khellang: :fire:

Pvlerick: Oops, will fix :+1:

Pvlerick:
I changed some of tests into a ``Theory`` which is more appropriate.

khellang:
:shipit:

nblumhardt:
I'm a bit wary of the performance impact of calling `StartsWith()` twice for every destructured parameter of every log event. The old indexer code was deliberately simplistic to avoid the hit (profiler-directed optimisation I think).

Nancy does it like this to keep the checks ordered cheapest-first:

```csharp
return type.IsGenericType
       && (type.Attributes & TypeAttributes.NotPublic) == TypeAttributes.NotPublic
       && (type.Name.StartsWith("<>", StringComparison.OrdinalIgnoreCase) || type.Name.StartsWith("VB$", StringComparison.OrdinalIgnoreCase));
```

(I've knocked out a couple of checks from the Nancy version that we can probably skip.)

This will keep the cost down to a couple of flag checks unless the type is both generic and non-public, which seems like a small number of cases.

Thoughts?

Pvlerick:
After some [micro benchmarking](http://yoda.arachsys.com/csharp/benchmark.html) I am very disappointed at how slow ``StartsWith`` is, about 5 times slower over 100000000 iterations.

Turns out that the fastes approach is to compare the first character with '<', using something like ``s[0] == '<'`` which performs slightly faster than char.IsLetter over 100000000 iterations.

I think that would be the best choice for anything C# related as it seems to be the preferred starting character for compiler generated type ([anonymous types and others](https://github.com/dotnet/roslyn/blob/5a97d296d3dedbe6cc7c580835da32adb3252d23/src/Compilers/CSharp/Portable/Symbols/Synthesized/GeneratedNames.cs#L59), I didn't find closures that are named ``c__DisplayClass`` but I suppose they are somewhere in there).

When it comes to VB, I didn't find anything in the compiler's source yet, but I'll keep looking when I have time. Unfortunately, it seems that it will start "VB" so we might to have to find another trick to make it fast.

khellang:
A while ago I did the exact same benchmarks for both `StartsWith` and `EndsWith`: https://gist.github.com/khellang/e1d684336ec9b98e4832 :smile:

nblumhardt:
Checking that the string starts with `'V'` and then falling back to the full `StartsWith()` check if true is probably sufficient - I don't think there can be _that_ many types starting with _V_ ;-) ... I'm guessing that `IsGenericType` and the attribute check also narrow the field a lot first?

@khellang handy benchmark!

Pvlerick:
Indeed, checking that the type is generic (and non public, as mentioned above) is hopefully also a cheap way to filter down what reaches the ``StartsWith``

I'll still inspect the compiler a bit more to see if there is no cheap way to detect a VB anonymous type.

Pvlerick:
[The constant that defines anonymous VB types prefix](https://github.com/dotnet/roslyn/blob/a4e375b95953e471660e9686a46893c97db70b0e/src/Compilers/VisualBasic/Portable/Symbols/SynthesizedSymbols/GeneratedNames.vb#L267) is ``"VB$Anonymous"`` and other examples can be found [here](Friend Const HoistedUserVariablePrefix As String = "$VB$Local_").

It seems safe to assume that compiler generated types in VB will start with ``"VB$"``, I will check performances to see what would be the best options.

On another front, I see that the Type.IsPublic is not available on _"Serilog..Net Platform 5.1"_ nor _''Serilog..Net Platform 5.4"_ according to compiler warnings.
I can see that compiler preprocessing directives are heavily used, but I can't figure out which one should be used to exclude these.
On the same line, I could use a ``[MethodImpl(MethodImplOptions.AggressiveInlining)]`` attribute on the method I created, but the compiler reports a warning on .NET 4.0, so if I understand correctly I should add it using a ``#if !NET40`` directive?

nblumhardt:
If `Type.IsPublic` is unavailable, there's probably a blessed "alternative" API. I'd guess:

```
(type.Attributes & TypeAttributes.NotPublic) == TypeAttributes.NotPublic
```

will get us what we want.

khellang:
You could also just use the `TypeInfo` version. I think that should be available, i.e. `.GetTypeInfo().IsPublic` :smile:

Pvlerick:
@nblumhardt ``Type.Attributes`` is not supported on _Serilog..Net Platform 5.1_ nor _'Serilog..Net Platform 5.4_

@khellang  ``TypeInfo.IsPublic`` and a lot of others are not supported in _Serilog..Net Framework 4_

However, a check that could be made is ``GetType().Namespace == null`` because [anonymous types are always in the global namespace](https://blogs.msdn.microsoft.com/ericlippert/2010/12/20/why-are-anonymous-types-generic/). By the same token, they are always generic and sealed, so I could use ``TypeInfo.IsSealed`` and ``TypeInfo.IsGenericType`` that are both supported, which seems to me quite a good filter before doing the expensive name check.

BTW, any idea why there is a check for ``typeTag.Length <= 0`` [on line 158 in PropertyValueConverter.cs](https://github.com/serilog/serilog/blob/d97f2b4f6fbd2afe42f3ccebfeef1e431eda239b/src/Serilog/Parameters/PropertyValueConverter.cs#L158)? I'm a bit puzzled by this check, I can't think of a case where ``GetType().Name`` would return an empty string.

khellang:
@Pvlerick I think that's why we have [this](https://github.com/serilog/serilog/blob/d97f2b4f6fbd2afe42f3ccebfeef1e431eda239b/src/Serilog/Platform/TypeInfo-net40.cs) and [this](https://github.com/serilog/serilog/blob/d97f2b4f6fbd2afe42f3ccebfeef1e431eda239b/src/Serilog/Platform/TypeExtensions-net40.cs), although the latter should probably be changed to just return back `Type`.

Pvlerick:
@khellang a backport of TypeInfo, interesting :smile: 

Oh then, I'll try different approach and see which one is faster.

khellang:
@Pvlerick Yes, ideally we'd just keep that extension method and return a `Type` instead. That should work for most cases, I think.

Pvlerick:
@khellang you mean in a conditional directive?

khellang:
Yeah, instead of returning (and maintaining) our own `TypeInfo` type, we should just return the old `Type`. It has most of the members we use from `TypeInfo`.

Pvlerick:
@khellang ok, thanks for clarifying :+1:

Pvlerick:
@khellang I simply backporting the ``IsNotPublic`` which is sufficient for the moment. I'd rather be doing the changes in ``TypeInfo`` in a separate PR just in case it comes with it own set of issues :smile:

src/Serilog/Parameters/PropertyValueConverter.cs
```diff
@@ -154,9 +156,16 @@ LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructur
 
             if (destructuring == Destructuring.Destructure)
             {
-                var typeTag = value.GetType().Name;
-                if (typeTag.Length <= 0 || !char.IsLetter(typeTag[0]))
+                var type = value.GetType();
+                var typeTag = type.Name;
+                var typeInfo = type.GetTypeInfo();
+                // Check if the type is an anonymous type
+                if (typeTag.Length <= 0
```
nblumhardt: I'd be tempted to move all the rest of the condition here into the `IsCompilerGeneratedTypeName` function, since it is going to be inlined anyway - otherwise it's a bit tricky for example to trace that the `Length > 0` precondition that function has is met without skipping back up to the call site.

nblumhardt:
This is looking great, thanks for sticking with it @Pvlerick :-) ... Just one tiny comment, let me know what you think, then let's ship it!

khellang:
I'm thinking it's time to start talking about removing support for .NET 4. It's becoming a burden and even MS doesn't support anything less than .NET 4.5.2 now...

Pvlerick:
@nblumhardt I thought about the same, but I wanted to keep the function appart to be able to test the VB anonymous type name too; on the other hand we are quite confident that it is working now. Ideally a test using a VB built assembly should be added...

@khellang just my two cents, from what I see in the code base it is indeed a burden but don't you think this will stop some people from using Serilog if they are stuck on older versions? I was still stuck with .NET 4.0 until a few months ago :sob:

khellang:
It won't prevent anyone. They can keep using 1.x :smile: It'll just be another incentive for people to upgrade (as if running on an unsupported platform wasn't enough already :wink:). The world has to move forward. I think this (2.0) might be time to do just that.

nblumhardt:
@khellang I keep going back to this too :-). Given that we don't actually even _test_ .NET 4 builds, I think we're effectively "not supporting" it currently. But, I've also heard from a lot of people who have benefited from 4.0 compatibility.

Perhaps how we present it is the key here? Continuing to support .NET 4.0 through Serilog 1.5 seems like a good compromise - putting a nice version/platform compatibility table on the README would just about cover it now. v2 would then become 4.5.2+...

But, we've digressed - probably deserves its own thread rather than being tangled up in this PR. I'll open a new ticket.

nblumhardt:
@Pvlerick thank you again for the PR!

Issue #619, opened 2015-12-23
PeriodicBatchingSink.Emit produces too many contentions


DmitryNaumov:
During my investigations of why simple foreach loop with some logging produces log entries with millisecond-sized gaps between each. I suspected thread context switches to be the reason. High contention rate was also noted by performance counters and performance profiler:

DA0039: (Average)Contention Rate / sec = 17.94; A very high rate of .NET Lock contentions is occurring.

I took a look at sink implementation and imo we can do better to avoid _possible_ contention on every write to log. I _do_ understand that if's are fast, but still if thread encounters occupied Monitor, then it's moved from RUNNING state to end of the wait list of the threads with same priority (I do simplify things here). So simple operation like write to log or even do an increment eventually may take milliseconds!

I didn't dive deeply in the logic behind checking _unloading & _started, but I think Monitor Enter/Exit can be easily replaced with Interlocked.CompareExchange operations which do not affect thread scheduling (thread won't go to WAIT state).

        public void Emit(LogEvent logEvent)
        {
            if (logEvent == null) throw new ArgumentNullException(nameof(logEvent));

            lock (_stateLock)
            {
                if (_unloading) return;
                if (!_started)
                {
                    // Special handling to try to get the first event across as quickly
                    // as possible to show we're alive!
                    _started = true;
                    SetTimer(TimeSpan.Zero);
                }
            }

            _queue.Enqueue(logEvent);
        }

nblumhardt:
Thanks Dmitry. As far as I can tell this would be a good change, but it'd need some careful planning and testing no doubt.

DmitryNaumov:
Nicholas,
if you don't mind I can try to create test to show the problem.

nblumhardt:
Thanks for the offer. I should have been more clear in that last comment, sorry - I was thinking ahead to determining whether we'd introduce any regressions by making the change. I'm confident the locking will have the impact you've described when there's heavy contention :+1:

I'm open to making the change, but the bar for maintaining correctness is pretty high here - it'll be a fairly challenging change to be thoroughly confident about. At the moment there's a lot of work in getting v2 out the door that would take precedence, but if you want to tackle it in the meantime please feel free.

DmitryNaumov:
Nicholas,
Given the following changes, do you think they can have some bad consequences?

        private volatile bool _unloading;
        private int _started;

        public void Emit(LogEvent logEvent)
        {
            if (logEvent == null) throw new ArgumentNullException(nameof(logEvent));

[...]

PR #623, opened 2016-01-07
Replace state transitions under lock with Interlocked operations

DmitryNaumov:
This PR is about reducing contention on PeriodicBatchingSink.Emit call by replacing Monitor with Interlocked calls. (#619)

One tricky thing I left as in original code is in CloseAndFlush method where sink won't be unsubscribed from AppDomain events and timer won't be disposed if sink wasn't actually started (no calls to Emit made). Former clearly not an issue and the latter one also shouldn't cause problems for BCL Timer and portable one.


merbla:
Hey @DmitryNaumov, thanks for the PR.  There is some work over @ #593.  Will try to pull this in when the sink is moved to its own space.  There are a few open issues relating to `PeriodicBatchingSink`.

src/Serilog/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
```diff
@@ -216,11 +213,8 @@ void OnTick()
                     while (_queue.TryDequeue(out evt)) { }
                 }
 
-                lock (_stateLock)
-                {
-                    if (!_unloading)
-                        SetTimer(_status.NextInterval);
-                }
+                if (_state == Started)
```
nblumhardt: I think this is racy - the thread responsible for closing/unloading may have switched state to Unloading between the volatile read and the SetTimer call on the next line, meaning we may end up trying to restart the timer after the unload sequence has run.

src/Serilog/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
```diff
@@ -41,19 +41,21 @@ namespace Serilog.Sinks.PeriodicBatching
...
 
-        readonly object _stateLock = new object();
 #if NO_TIMER
         readonly PortableTimer _timer;
 #else
         readonly Timer _timer;
 #endif
-        bool _unloading;
-        bool _started;
+        volatile int _state;
```
nblumhardt: I think using `Interlocked` along with `volatile` will generate a compiler warning (IIRC). Do we get one in the usages below? If so and we're happy it is benign we should `#pragma` it out.

DmitryNaumov: I saw no compiler warning here, but I double check it. Any way, it may be removed due to the following comment.

src/Serilog/Sinks/PeriodicBatching/PeriodicBatchingSink.cs
DmitryNaumov: I agree it is tricky place here, I spent whole day refreshing ECMA memory model and diffs between lock vs volatile vs Volatile.Read vs Thread.VolatileRead.

(I wrote a lot trying to back up this line, but then I understand something and remove it)

So, kind of "racing" can happen, even when checking state under lock. I mean other thread can be 1ms late setting unloading and then we have timer scheduled for one more tick.

And give that this line is situated in OnTick method, which is executed in another (timer's) thread and does'nt concern us in terms of contention, here we can use Thread.VolatileRead as strongest fence to read latest value.

This also gives possibility to remove volatile keyword from _state and avoid _possible_ compiler warning mentioned in comment above.

nblumhardt:
Just following the comment on the outdated diff above:

> So, kind of "racing" can happen, even when checking state under lock. I mean other thread can be 1ms late setting state to Unloading and then we have timer scheduled for one more tick.

I'm not sure about this case, the current implementation is careful that `_started` and `_unloading` always match the timer state by dealing with both only under the lock. If we have a bug here currently it'd be good to track down.

I really appreciate the work you've done on this one, I think it's an important improvement. Just looking at the complexity around the 100% lock-free version, perhaps a quicker and easier alternative we can consider is simply relaxing the guarantee surrounding enqueued-but-never-flushed events?

I believe it's close to an earlier version you proposed:

```csharp
public void Emit(LogEvent logEvent)
{
    if (logEvent == null) throw new ArgumentNullException(nameof(logEvent));

    lock (_stateLock)
    {
        if (_unloading) return;
        if (!_started)
        {
            // Special handling to try to get the first event across as quickly
            // as possible to show we're alive!
            _started = true;
            SetTimer(TimeSpan.Zero);
        }
    }

    _queue.Enqueue(logEvent);
}
```

Would become:

```csharp
public void Emit(LogEvent logEvent)
{
    if (logEvent == null) throw new ArgumentNullException(nameof(logEvent));

    if (_unloading)
        return;

    if (!_started)
    {
        lock (_stateLock)
        {
            if (_unloading) return;
            if (!_started)
            {
                // Special handling to try to get the first event across as quickly
                // as possible to show we're alive!
                _started = true;
                SetTimer(TimeSpan.Zero);
            }
        }
    }

    _queue.Enqueue(logEvent);
}
```

The two reads outside the lock are just optimisations:

 * If the read of `_unloading` is stale then we'll potentially enqueue an event that won't be flushed
 * If the read of `_started` is stale we'll take a lock (causing a memory barrier anyway) and make the check again

(As you're probably aware, double-checked locking like this is famously broken in Java but works on the CLR.)

The invariants around the timer and two Boolean flags would remain unchanged. The all-queued-events-are-flushed invariant would be relaxed.

I think relaxing the latter invariant is harmless because if it was maintained the event would be lost anyway. Keeping a reference alive that would otherwise be GC'd is the side-effect, but once the sink itself is disposed it's most likely going to be a candidate for GC itself anyway, releasing the reference.

Once `_started` is set to `true` in the calling thread (always happens eventually because of the fence implied by `lock`)  then the lock will not be taken again, so contention will not occur on `Emit()` calls. Using locks in the other methods/worker thread is fine because there's little contention there.

What do you think?

nblumhardt:
To guarantee that the first event gets into the first batch, these lines:

```csharp
                _started = true;
                SetTimer(TimeSpan.Zero);
```

Could duplicate the enqueue call and short-circuit:

```csharp
                _queue.Enqueue(logEvent);
                _started = true;
                SetTimer(TimeSpan.Zero);
                return;
```

(Alternatively the queueing could happen earlier, but chances of enqueueing events that are never flushed would increase.)

DmitryNaumov:
@nblumhardt I agree that clearness is important code metric and lock-free implementations have a problem with it. My goal was to reduce contention in most cases, while edge cases like not started or unloading are not so important. So if you're fine with relaxed guarantee, I revert it back to simplified version and at the same time pull changes from master to remove conflicts.

nblumhardt:
@DmitryNaumov that sounds good to me.

It's a critical piece of code so @serilog/core please take a look and weigh in if you can see any gotchas Dmitry and I have missed :-)

DmitryNaumov:
Did what we discussed, can't wait to see it in prod :-)

nblumhardt:
Awesome, can't wait either! :shipit:

PR #626, opened 2016-01-10
Use round-trippable floating point formatting in JSON

nblumhardt:
Use "R" round-trip specifier when formatting floating-point numbers to JSON. Without this specifier, floating point numbers logged as JSON cannot be parsed using the standard `Double` and `Single` parsing in the framework, and break libraries such as JSON.NET that rely on these.

See: http://www.hanselman.com/blog/WhyYouCantDoubleParseDoubleMaxValueToStringOrSystemOverloadExceptionsWhenUsingDoubleParse.aspx

(Fixes https://github.com/continuousit/seq-releases/issues/358)


nblumhardt:
(Earlier test failure on this was due to our old friend the intermittent CoreCLR periodic timer failure. This is ready to merge.)

merbla:
:shipit:

PR #629, opened 2016-01-11
Rolling File Sink - Expand environment variables

joshka:
An open source WinForms app I use (DaxQuery) uses Serilog for logging, but makes the user configure the config file to point to a valid logging directory out of the box. This change allows filenames like `%TEMP%\DaxQueryLog.txt` or `%ALLUSERSPROFILE%\DaxQuery\Log.txt` to be specified directly in the config instead of hard coding an invalid folder or having to update this configuration in the app's installer.

Note: I made a few changes to the DotSettings file to align with the practice of not including private modifiers on class members that seems to be aligned with the practice I noticed in the code I changed. Feel free to throw this change out if it's not aligned with this project's style guidelines.

An extension of this PR would also provide a similar feature to the standard non rolling file sink, but the filename generation was not yet pulled out of the class. Perhaps an up-for-grabs issue in the tracker for someone who needs it?

### DIFF
--- a/Serilog.sln.DotSettings
+++ b/Serilog.sln.DotSettings
@@ -123,7 +123,8 @@
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_OBJECT_AND_COLLECTION_INITIALIZER_STYLE/@EntryValue">CHOP_ALWAYS</s:String>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/JavaScriptCodeFormatting/ALIGN_MULTIPLE_DECLARATION/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/JavaScriptCodeFormatting/JavaScriptFormatOther/ALIGN_MULTIPLE_DECLARATION/@EntryValue">True</s:Boolean>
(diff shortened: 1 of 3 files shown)


nblumhardt:
Hi Josh - thanks for the PR.

Wouldn't it be easier just to add `Environment.ExpandEnvironmentVariables(x)` to the WinForms app? (This is why we haven't rolled path expansion through the rest of the code-based configuration API).

joshka:
I thought the same initially,
But I went this way as you can fairly reasonably assume that the logging config is often done only in the app.config, not with code.

-- 
Josh McKinney

nblumhardt:
Thanks for the follow-up. I'm not sure I'm on the same page - if it's done in config, why not include the environment variable in there? `ReadFrom.AppSettings()` will do the expansion.

joshka:
Aha - I see where you're coming from now. I didn't realize the environment variable expansion was happening automatically under the hood. I looked at the docs for the rolling file and app settings to see if it was supported and didn't see anything obvious. Perhaps dropping this PR and putting a quick note in the https://github.com/serilog/serilog/wiki/AppSettings wiki page about environment variables is the right 'fix' for this instead of the proposed change.

nblumhardt:
Thanks Josh, that sounds like a plan, will do now :+1:

PR #636, opened 2016-01-16
Enable event enrichers from config

johncrn:
G'day there,

I ran into a circumstance where I wanted to enable the ThreadId enricher through config rather than code, but found it wasn't supported. This PR allows event enrichers implementing the ILogEventEnricher interface to be toggled on from the config file. E.g.:

    <add key="serilog:enrich:with:ThreadId"/>
    <add key="serilog:enrich:with:MachineName"/>
    <add key="serilog:enrich:with:ProcessId"/>
    <add key="serilog:enrich:with:EnvironmentUserName"/>
    <add key="serilog:enrich:with:LogContext"/>

Some caveats:
* Event enricher types are expected to have parameterless constructors.
* As the LogContext enricher wasn't public the reflection code scans all declared types in the config assemblies rather than just exported types, so it may pick up types not geared for public consumption. Not sure why LogContext isn't public but I didn't want to touch it. :)
* Case sensitive, but looks like that's in keeping with the code base.
* I'm not thrilled with the config key syntax. Looks a bit strange side by side the existing with-parameter values.

Hope this is helpful to someone!


nblumhardt:
Awesome! I've wanted this feature a few times as well. The syntax seems fine to me; for consistency there's an alternative we should probably also consider too:

When `write-to` is processed, we search for extension methods on `LoggerSinkConfiguration`, i.e. `write-to:File` actually calls `WriteTo.File()` under the hood. Not a typical approach, but keeps the mechanisms from drifting apart.

In the case of `enrich:with` could we do the same thing? This would mean `enrich:with:LogContext` would  be written `enrich:FromLogContext` mapping to `Enrich.FromLogContext()`. Machine names etc. would be represented as `enrich:WithMachineName`.

```xml
  <add key="serilog:write-to:File" />
  <add key="serilog:enrich:FromLogContext" />
```

`==`

```csharp
  .WriteTo.File()
  .Enrich.FromLogContext()
  .CreateLogger();
```

What do you think?

johncrn:
Ah yep, that makes sense. I'll update the PR later on.

johncrn:
Ok this now looks like:

    <add key="serilog:enrich:WithThreadId"/>
    <add key="serilog:enrich:WithMachineName"/>
    <add key="serilog:enrich:WithProcessId"/>
    <add key="serilog:enrich:WithEnvironmentUserName"/>
    <add key="serilog:enrich:FromLogContext"/>

Unlike the sink config the enrichers don't support supplying arguments via config, meaning enrichers with non default params cannot be applied. None of the supplied enrichers have params so I didn't think it a big deal.

Also the implementation in `KeyValuePairSettings` could be a little more DRY as it has similarities to the sink method calls. Just let me know if you want that tidied up a bit more.

johncrn:
Build failed with this error:
```
Error message:
Assert.Equal() Failure

Expected: 1

Actual:   0
Error stack trace:
   at Serilog.Tests.Sinks.PeriodicBatching.PeriodicBatchingSinkTests.WhenAnEventIsEnqueuedItIsWrittenToABatch_OnTimer() in C:\projects\serilog\test\Serilog.Tests\Sinks\PeriodicBatching\PeriodicBatchingSinkTests.cs:line 89
```
On this test:

        [Fact]
        public void WhenAnEventIsEnqueuedItIsWrittenToABatch_OnTimer()
        {
            var pbs = new InMemoryPeriodicBatchingSink(2, TinyWait, TimeSpan.Zero);
            var evt = Some.InformationEvent();
            pbs.Emit(evt);
            Thread.Sleep(TinyWait + TinyWait);
            pbs.Stop();
            Assert.Equal(1, pbs.Batches.Count);
            Assert.False(pbs.WasCalledAfterDisposal);
        }

Perhaps an AppVeyor issue, or is it something more subtle?

johncrn:
Did a little refactoring in `KeyValuePairSettings`. Build seems good now.

nblumhardt:
Thanks @johncrn - unfortunately the test failure is an intermittent one we're yet to resolve, it's only present on CoreCLR/v2. Nothing to worry about right now. Will check out the changes ASAP :-) Cheers!

nblumhardt:
Looks awesome, love this change. Thanks @johncrn!

Issue #635, opened 2016-01-14
Remove .NET 4.0 compatibility from 2.0 packages


nblumhardt:
Conditional platform support is starting to take its toll on the readability and malleability of the codebase. With .NET 4.0 having reached end-of-life it may be time to shed some of this in our 2.0 release.

It's possible, through Serilog 1.5, to continue supporting all .NET versions from 4.0 forwards, without needing to make the 2.0 codebase carry this burden directly.

This ticket has two parts:

* Update Serilog to drop .NET 4.0
* Update the README to include prominent version compatibility info

Comments, for/against? // @khellang 

From #618

colin-young:
One advantage would be for LogContext since AsyncLocal was introduced with 4.5. I've got a mess of conditional statements in there right now... And they're nested. Ugh.

Issue #588, opened 2015-11-17
Port LogContext to .NET Core


nblumhardt:
From #535. This may use AsyncLocal<T>.

nblumhardt:
Ready to attempt on `dev` if anyone is interested.

colin-young:
What's the status on this? Still up for grabs? I might be able to make an attempt at it, especially seeing as I'd kind of like to make use of it.

nblumhardt:
Absolutely! I'll knock off the `up-for-grabs` label in anticipation :-) ... Let us know if you need a hand with any of it.

colin-young:
Fork and pull request the usual way to do things here? I'm sure I'll be back with questions.

colin-young:
I'm presuming from the initial comment here that the intention was to replace CallContext with AsyncLocal<T> for .Net 4.5 and up?

nblumhardt:
I think so - `AsyncLocal` looks like the right API going forwards :+1:

colin-young:
So I've got things building now, with `net452`, `dotnet5.1` and `dotnet5.4`. My biggest concern, which I'm just about to start investigating is the unit tests. VS Test Explorer gives me 160 tests passed, 1 skipped. Build.ps1 gives me a total of 76 tests, 5 failing, 0 skipped, but the the console output clearly shows it is skipping `DoesNotPrevenCrossDomainCalls` because it "fails when run from Build.ps1."

Those numbers clearly don't agree with each other, so I have very little confidence in the current state of my fork.

Figured I'd ask here in case this is something that's sort of expected. In the meantime, I'm going to see if I can make sense of things.

khellang:
@colin-young It depends on the targets that you're running the tests under. Some of them are #ifdef'ed for `NET45` etc., so the numbers don't necessarily add up :smile:

colin-young:
@khellang Yup. Eventually figured that out when I decided to scroll back far enough in the console output to see the DNX 4.5.2 tests. I think you mean #ifdef'd as features, not platform :)

I was more concerned about the failure from the command line and having it passing in VS. The failure was a LogContext implementation error by me (surprise!) that I've corrected and now all tests are passing... however, I'm still concerned about the fact that VS somehow made that one pass with the error.

I'm going to do a bit of testing of the new-to-core LogContext in my test project, and then see if I can't figure out some way to do some async unit testing to make sure it really works.

adamhathcock:
:+1:

PR #640, opened 2016-01-19
Initial attempt at LogContext support in .Net Core

colin-young:
For #588 

* Removed .Net 4.0 (hoping for #635 )
* Changed from .Net 4.5.1 to 4.5.2, but left 4.5 as-is
* Added ASYNCLOCAL and REMOTING feature pivots
* Removed remoting support from .Net Core
* Replaced CallContext with AsyncLocal<T> when ASYNCLOCAL feature is available

I have not tested in a remoting situation, and have not reviewed the unit tests to see how they are covering that. All unit tests do pass (except maybe `DoesNotPreventCrossDomainCalls` which I haven't run yet).

Looking for some early feedback.

Thanks.


src/Serilog/project.json
```diff
@@ -31,7 +19,7 @@
         "dotnet5.1": {
             "compilationOptions": {
                 "keyFile": "../../assets/Serilog.snk",
-                "define": [ "NO_APPDOMAIN" ]
+                "define": [ "ASYNCLOCAL", "NO_APPDOMAIN" ]
```
nblumhardt: Is this missing a `"LOGCONTEXT"`? If so, is LogContext now supported on all target plaforms? It'd be great to get rid of the conditional if we can.

colin-young: No, unfortunately it is not. It shouldn't have the `ASYNCLOCAL` either, but that was harmless because of the missing `LOGCONTEXT`. Understanding https://github.com/dotnet/corefx/blob/c747d929093681a475559f804043cb5ac5470a49/Documentation/project-docs/standard-platform.md is apparently completely beyond me, because my interpretation is that `dotnet5.1` should run on _.Net Framework 4.5_, but attempting to compile that way results in errors about CallContext not being available (it came with 4.5 AFAIK), `dotnet5.2` which I understood to mean _.Net Framework 4.5.1_ and up results in even more errors...

I will remove the ASYNCLOCAL and I cleaned up another bit of #if !NET45 nonsense, although I've still got a bit #if !DOTNET5_1 around the entire `EnvironmentUserNameEnricher`

colin-young:
This is a mess now. I'm going to close this PR and try again.

PR #648, opened 2016-01-26
Enable LogContext for .Net Core.

colin-young:
For issue #588 

Removed .Net 4.0 (hoping for #635 )
Changed from .Net 4.5.1 to 4.5.2, but left 4.5 as-is
Added ASYNCLOCAL and REMOTING feature pivots
Removed remoting support from .Net Core
Replaced CallContext with AsyncLocal when ASYNCLOCAL feature is available
I have not tested in a remoting situation, and have not reviewed the unit tests to see how they are covering that. All unit tests do pass (except maybe DoesNotPreventCrossDomainCalls which I haven't run yet).

Looking for some early feedback.

Thanks.

### DIFF
--- a/src/Serilog/Context/ImmutableStack.cs
+++ b/src/Serilog/Context/ImmutableStack.cs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if LOGCONTEXT
 using System;
 using System.Collections.Generic;
 #if REMOTING
@@ -82,4 +81,3 @@ namespace Serilog.Context
 
     }
 }
-#endif
--- a/src/Serilog/Context/ImmutableStack.cs
+++ b/src/Serilog/Context/ImmutableStack.cs
@@ -37,6 +37,10 @@ namespace Serilog.Context
         public ImmutableStack(SerializationInfo info, StreamingContext context)
         {
         }
+
+        void ISerializable.GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+        }
 #endif
 
         ImmutableStack()
@@ -76,11 +80,6 @@ namespace Serilog.Context
 
         public T Top => _top;
 
-#if REMOTING
-        void ISerializable.GetObjectData(SerializationInfo info, StreamingContext context)
-        {
-        }
-#endif
     }
 }
 #endif
--- a/src/Serilog/Context/ImmutableStack.cs
+++ b/src/Serilog/Context/ImmutableStack.cs
@@ -15,7 +15,9 @@
 #if LOGCONTEXT
 using System;
 using System.Collections.Generic;
+#if REMOTING
 using System.Runtime.Serialization;
+#endif
 
 namespace Serilog.Context
 {
@@ -33,10 +35,6 @@ namespace Serilog.Context
 
 #if REMOTING
         public ImmutableStack(SerializationInfo info, StreamingContext context)
-        {
-        }
-#else
-        public ImmutableStack(StreamingContext context)
         {
         }
 #endif
--- a/src/Serilog/Context/ImmutableStack.cs
+++ b/src/Serilog/Context/ImmutableStack.cs
@@ -21,15 +21,25 @@ namespace Serilog.Context
 {
     // Needed because of the shallow-copying behaviour of
     // LogicalCallContext.
+#if REMOTING
     [Serializable]
     class ImmutableStack<T> : IEnumerable<T>, ISerializable
+#else
+    class ImmutableStack<T> : IEnumerable<T>
+#endif
     {
         readonly ImmutableStack<T> _under;
         readonly T _top;
 
+#if REMOTING
         public ImmutableStack(SerializationInfo info, StreamingContext context)
         {
         }
+#else
+        public ImmutableStack(StreamingContext context)
+        {
+        }
+#endif
 
         ImmutableStack()
         {
@@ -68,9 +78,11 @@ namespace Serilog.Context
 
         public T Top => _top;
 
+#if REMOTING
         void ISerializable.GetObjectData(SerializationInfo info, StreamingContext context)
         {
         }
+#endif
     }
 }
 #endif
--- a/src/Serilog/Context/LogContext.cs
+++ b/src/Serilog/Context/LogContext.cs
@@ -12,9 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
 #if LOGCONTEXT
 using System;
+#if REMOTING
 using System.Runtime.Remoting.Messaging;
+#endif
+#if ASYNCLOCAL
+using System.Collections.Generic;
+using System.Threading;
+#endif
 using Serilog.Core;
 using Serilog.Core.Enrichers;
 using Serilog.Events;
@@ -42,11 +49,19 @@ namespace Serilog.Context
     /// </code>
     /// </example>
     /// <remarks>The scope of the context is the current logical thread, using
-    /// <see cref="CallContext.LogicalGetData"/> (and so is
-    /// preserved across async/await calls).</remarks>
+#if ASYNCLOCAL
+    /// <seealso cref="AsyncLocal{T}"/>
+#else
+    /// <seealso cref="CallContext"/>
+#endif
+    /// (and so is preserved across async/await calls).</remarks>
     public static class LogContext
     {
+#if ASYNCLOCAL
+        static readonly AsyncLocal<ImmutableStack<ILogEventEnricher>> data = new AsyncLocal<ImmutableStack<ILogEventEnricher>>();
+#else
         static readonly string DataSlotName = typeof(LogContext).FullName;
+#endif
 
         /// <summary>
         /// When calling into appdomains without Serilog loaded, e.g. via remoting or during unit testing,
@@ -136,10 +151,13 @@ namespace Serilog.Context
         {
             get
             {
-
+#if ASYNCLOCAL
+                return data.Value;
+#else
                 var data = CallContext.LogicalGetData(DataSlotName);
 
                 ImmutableStack<ILogEventEnricher> context;
+#if REMOTING
                 if (PermitCrossAppDomainCalls)
                 {
                     context = ((Wrapper) data)?.Value;
@@ -148,17 +166,38 @@ namespace Serilog.Context
                 {
                     context = (ImmutableStack<ILogEventEnricher>)data;
                 }
-
+#else
+                context = data;
+#endif
                 return context;
+#endif
             }
             set
             {
-
-                var context = !PermitCrossAppDomainCalls ? (object)value : new Wrapper { Value = value };
-
+#if ASYNCLOCAL
+                data.Value = GetContext(value);
+#else
+                var context = GetContext(value);
                 CallContext.LogicalSetData(DataSlotName, context);
+#endif
             }
         }
+
+#if REMOTING
+        static object GetContext(ImmutableStack<ILogEventEnricher> value)
+        {
+            var context = !PermitCrossAppDomainCalls ? (object) value : new Wrapper
+            {
+                Value = value
+            };
+            return context;
+        }
+#else
+        static ImmutableStack<ILogEventEnricher> GetContext(ImmutableStack<ILogEventEnricher> value)
+        {
+            return value;
+        }
+#endif
 
         internal static void Enrich(LogEvent logEvent, ILogEventPropertyFactory propertyFactory)
         {
@@ -187,10 +226,13 @@ namespace Serilog.Context
             }
         }
 
+#if REMOTING
         sealed class Wrapper : MarshalByRefObject
         {
             public ImmutableStack<ILogEventEnricher> Value { get; set; }
         }
+#endif
     }
 }
+
 #endif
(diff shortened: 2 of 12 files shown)


khellang:
Just skimming over this, it feels a bit #ifdef heavy for my taste. Are there any cases where it might be worth #ifdefing a whole class instead?

colin-young:
That was my feeling with the #ifdefs, but I wasn't sure what direction you might want to go with serilog. There's also the issue with remoting not being supported in Core, but needing `AsyncLocal`. And .Net 4.5 doesn't support `AsyncLocal`. The other option would be to use `AsyncLocal` only for Core, and then REMOTING would be equivalent to ASYNCLOCAL, but not a strictly accurate solution. My personal preference between those 2 options would be the extra #ifdef statements.

colin-young:
I rearranged the #ifdefs a little and got a slight reduction. IMHO it makes it a little more readable, but there are still more than I'd consider ideal.

src/Serilog/Context/ImmutableStack.cs
```diff
@@ -21,15 +21,25 @@ namespace Serilog.Context
...
     {
         readonly ImmutableStack<T> _under;
         readonly T _top;
 
+#if REMOTING
         public ImmutableStack(SerializationInfo info, StreamingContext context)
         {
         }
+#else
+        public ImmutableStack(StreamingContext context)
```
nblumhardt: Will anything actually use this constructor?

src/Serilog/Context/ImmutableStack.cs
```diff
@@ -68,9 +78,11 @@ System.Collections.IEnumerator System.Collections.IEnumerable.GetEnumerator()
 
         public T Top => _top;
 
+#if REMOTING
         void ISerializable.GetObjectData(SerializationInfo info, StreamingContext context)
```
nblumhardt: Perhaps we could move this up into the first `#if REMOTING` block to piggy-back on its `#if`?

src/Serilog/Context/LogContext.cs
```diff
@@ -12,9 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+
 #if LOGCONTEXT
```
nblumhardt: Sorry if this is a duplicate, I may have already asked - but are there any platforms on which `LogContext` is unsupported now? (I.e. is `AsyncLocal` on 5.1?)

src/Serilog/Context/LogContext.cs
```diff
@@ -42,11 +49,19 @@ namespace Serilog.Context
     /// </code>
     /// </example>
     /// <remarks>The scope of the context is the current logical thread, using
-    /// <see cref="CallContext.LogicalGetData"/> (and so is
-    /// preserved across async/await calls).</remarks>
+#if ASYNCLOCAL
+    /// <seealso cref="AsyncLocal{T}"/>
+#else
+    /// <seealso cref="CallContext"/>
```
nblumhardt: Can we add the `.LogicalGetData` so that it's easier to track down docs on MSDN?

src/Serilog/Context/LogContext.cs
```diff
@@ -42,11 +49,19 @@ namespace Serilog.Context
...
+#endif
+    /// (and so is preserved across async/await calls).</remarks>
     public static class LogContext
     {
+#if ASYNCLOCAL
+        static readonly AsyncLocal<ImmutableStack<ILogEventEnricher>> data = new AsyncLocal<ImmutableStack<ILogEventEnricher>>();
```
nblumhardt: Name probably should be capitalized `Data`.

src/Serilog/Context/LogContext.cs
```diff
@@ -132,33 +147,63 @@ static ImmutableStack<ILogEventEnricher> GetOrCreateEnricherStack()
...
         static ImmutableStack<ILogEventEnricher> Enrichers
         {
             get
             {
-
                 var data = CallContext.LogicalGetData(DataSlotName);
 
                 ImmutableStack<ILogEventEnricher> context;
+#if REMOTING
                 if (PermitCrossAppDomainCalls)
```
nblumhardt: `PermitCrossAppDomainCalls` should be `#if REMOTING` since `AppDomain` doesn't appear on the modern targets.

src/Serilog/Context/ImmutableStack.cs
colin-young: Nope. Removed.

src/Serilog/Enrichers/EnvironmentUserNameEnricher.cs
```diff
@@ -21,7 +21,11 @@
 namespace Serilog.Enrichers
 {
     /// <summary>
+#if USERNAMEFROMENV
```
nblumhardt: I'd prefer not including changes to this enricher in this PR - let's keep it to `LogContext` and leave the enricher work for a later PR (there's some factoring work being done that will affect this).

src/Serilog/project.json
```diff
@@ -7,22 +7,10 @@
     "licenseUrl": "http://www.apache.org/licenses/LICENSE-2.0",
     "iconUrl": "http://serilog.net/images/serilog-nuget.png",
     "frameworks": {
-        "net40": {
```
nblumhardt: I think this change is inevitable - :+1:

src/Serilog/Context/LogContext.cs
colin-young: I'm not entirely clear on what 5.1 is. I thought from [standard platform](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/standard-platform.md) that it was 4.5, but if I try to compile with either `AsyncLocal` or `CallContext` it fails on 5.1.

src/Serilog/Context/LogContext.cs
colin-young: Yup. Done.

src/Serilog/Context/ImmutableStack.cs
colin-young: Good call. Done.

src/Serilog/Context/LogContext.cs
colin-young: Yup. That's what happens when you learn to rely on Resharper to catch that stuff and then a project goes and breaks Resharper on you.

src/Serilog/Enrichers/EnvironmentUserNameEnricher.cs
colin-young: Hmm. I'll have to take another look at that later tonight. IIRC the only reason I did that was because it wouldn't work otherwise. But lots of other things have changed since then so that may not be true any longer.

src/Serilog/Context/LogContext.cs
nblumhardt: Ah, that's a shame. What do you think (+ @merbla @khellang) about using `[ThreadStatic]` for this on 5.1?

There are two benefits:

1. `LogContext` is a pretty fundamental API, and it is still useful even without async support (better than nothing, if you're targeting very limited surface area!), but more importantly:
2. Libraries may _build_ against the 5.1 surface area, but then be installed in applications that do in fact target 4.5+ or 5.4+ - in these cases, the binary used at runtime would be the version with full `LogContext` support ([bait-and-switch](http://log.paulbetts.org/the-bait-and-switch-pcl-trick/))

Worth a shot?

src/Serilog/Context/LogContext.cs
colin-young: You mean the property itself, right?

src/Serilog/Enrichers/EnvironmentUserNameEnricher.cs
colin-young: That seems to compile without any changes in the enricher now. I've removed that change.

src/Serilog/Context/LogContext.cs
merbla: This a site @khellang pointed me to.  Not always 100% however a great start.

- http://packagesearch.azurewebsites.net/?q=CallContext
- http://packagesearch.azurewebsites.net/?q=AsyncLocal

These would indicate `5.1` could be difficult.

nblumhardt: @merbla thanks for the refs. The idea is not to use either of those, but something like:

```csharp
[ThreadStatic]
static ImmutableStack<ILogEventEnricher> Data;
```

This should work on all platforms and does the job except for carrying state across async calls.

src/Serilog/Context/LogContext.cs
nblumhardt: Yes, sorry about the vague comment :+1:

colin-young: Moved `PermitCrossAppDomainCalls` property to inside the REMOTING block at the bottom of the file.

colin-young:
Any comments on the updated PR incorporating all (I hope) previous comments?

nblumhardt:
Thanks Colin - looking good. Did you have a chance to consider:

```csharp
[ThreadStatic]
static ImmutableStack<ILogEventEnricher> Data;
```

for .NET 5.1 support? It'd be great to cover all platforms with this and call it "done done" :-)

(I also think extracting some "policy" classes would be a win here, given the amount of conditional compilation required, but probably better to move this forward and do some refactoring down the line.)

(Let me know if you're out of time, great effort on everything so far!)

colin-young:
I did not. I'll take a look at that tomorrow. Fortunately I'm in the middle of wrestling with some versioning issues with my libraries (a.b.c-beta-9 is apparently newer than a.b.c-beta-10 and the build tools absolutely refuse to forget it ever existed even if I've removed it from my NuGet server and cleared the cache), so I'll be running a bunch of builds tomorrow. Nobody will notice a few extra changes... plus it's kind of related to what I'm supposed to be doing anyway.

I have to admit though, I have no clue what .NET 5.1 is.

IanYates:
@nblumhardt  might have meant 4.5.1 perhaps?

colin-young:
No, there is a framework target of dotnet5.1.

IanYates:
Oh yeah - that's soon to be "netstandard" IIRC.  (http://www.hanselman.com/blog/ASPNET5IsDeadIntroducingASPNETCore10AndNETCore10.aspx)

khellang:
Yep. `dotnet5.1` is the equivalent of `netstandard1.0` in [this document](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md#mapping-the-net-platform-standard-to-platforms). `dotnet5.x` was the "working TFM" before it became `netstandard1.x`. It's basically the lowest API set possible that we can support, which means we pretty much target anything. The reason `net451` is still there is because some clients don't yet support the new TFMs. By the time we ship 2.0, we'll probably move to `netstandard1.0` only.

khellang:
BTW, we should look at moving to a newer `netstandard` than `1.0` as well. I'm not sure how much sense it makes to support those older frameworks. It's very limiting in terms of API surface.

colin-young:
@khellang the bit that doesn't make sense to me is that my understanding from that page is 1.0 is 4.5, which is supposed to have AsyncLocal, but that's not what the compiler tells me. It's entirely possible that I'm misinterpreting that page.

khellang:
1.0 is not 4.5, 1.1 is 4.5. And even though .NET 4.5 might have it, other platforms might not. `netstandard` is basically the minimum API set available on all the supported platforms. As you can see, it also supports Windows Phone 8.0, which might or might not have that API.

colin-young:
Whoops. I keep getting dotnet5.1 mixed up with netstandard1.1 when it's really netstandard1.0.

So it looks like we can't just use dotnet5.{1|2} because of Windows Phone 8.1 (and maybe Windows 8.1) not supporting AsyncLocal or CallContext (things are starting to make a little more, well, sense is a bit strong, but at least there is a sort of twisted logic to it now). I'm not entirely sure if the separate net452 framework specifier would take precedence over the dotnet5.n specifier if running on the full .Net 4.5 framework. Plus it's just plain confusing to have overlapping specifications IMHO.

[This](https://github.com/NuGet/Home/issues/1691) is the most useful thing I've found to translate. I get that naming is hard, but it would be nice if all the names were collected in one place to make it easier to translate.

colin-young:
Yeah, that was an premature commit. I just fixed it so they are all at 5.1.

src/Serilog/Context/ImmutableStack.cs
```diff
@@ -15,22 +15,34 @@
 #if LOGCONTEXT
```
nblumhardt: If we now have dotnet5.1 TFM support, there's no need for `#if LOGCONTEXT` anymore - that's all the target platforms covered (I hope :-)) - they should all support `LogContext` now in one form or another.

colin-young: Ha ha. You're right. I'll see how far I can get on my Mac tonight, otherwise it will be tomorrow morning. I've been meaning to see how the cross-platform bits are working.

colin-young:
I'll check out the full build and tests tomorrow. A bit limited on OS X since I could only check dotnet5.4 framework.

nblumhardt:
Legendary; thanks @colin-young !

PR #654, opened 2016-02-01
Less surprising "child logger" level configuration

nblumhardt:
Found a few rough edges in `WriteTo.Logger()` - fixes:

 - When using `WriteTo.Logger(lc => ...)` the sub-logger inherits the level configured on the parent
 - Last-in always wins when setting minimum levels even when switching between `LoggingLevelSwitch` and basic level config
 - `WriteTo.Logger` no longer ignores the passed-in level switch argument if present (consistent with other sinks)

### DIFF
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -32,15 +32,18 @@ namespace Serilog.Configuration
     {
         readonly LoggerConfiguration _loggerConfiguration;
         readonly Action<ILogEventSink> _addSink;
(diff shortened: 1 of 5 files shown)


merbla:
LTGM.  Keen to get these sort of semantics into the v2 doco.

nblumhardt:
Yeah, it will be tedious but I'm hoping we can get a pretty exhaustive changelist up when 2.0 ships.

This PR should be ready to go... :)

nblumhardt:
@merbla considering your LGTM a :shipit:... :boom:

merbla:
Sorry @nblumhardt my button clicking leaves a lot to be desired...

Issue #650, opened 2016-01-27
Add `buffered` Boolean configuration parameter to File/RollingFile sinks


nblumhardt:
Both `File` and `RollingFile` flush the file stream after each write. This is a good default, as it's generally the event immediately before an application crash that you need when debugging.

Providing a `buffered` Boolean parameter to both of these configuration methods, defaulting to `false`, seems reasonable for anyone looking for less latency and prepared to make the trade-off.

http://stackoverflow.com/questions/35020648/serilog-file-rollingfile-sink-and-buffering

Pvlerick:
I'm happy to have a look into that :smile:

saitejkuralla:
Hi @nblumhardt
We are facing an issue on performance in logging with serilog 
Below attachment is the delay stats between logs for each log it is taking minimum 400 ms 
Please help us on this.
![image](https://user-images.githubusercontent.com/11627655/49520626-b09d1680-f8c9-11e8-9531-4e03ae3d9d1d.png)
Below is the configuration we used for logging 
![image](https://user-images.githubusercontent.com/11627655/49520715-e641ff80-f8c9-11e8-976e-cfaf2fd2c3b6.png)

BernardTheDog:
I tested it, and when logging strings, its was about 20ms when i logged a class it was a bit more overhead, and when i added bigger and bigger logs, the time for saving increased. So when i tested with 100mb data it took around 400ms to store it.
But if you are able to use async sink it seems to be much faster: https://github.com/serilog/serilog-sinks-async. i got around 15 ms for 300mb of data

PR #656, opened 2016-02-04
Adding buffered boolean param to FileSink and RollingFileSink

Pvlerick:
This is a rough draft implementation to address #650. I open this PR so we have code to discuss; this code is definitely not ready :wink: 

In all honesty, I don't like the boolean parameter too much. Very often boolean parameters are passed around to mean "behave in this way if true, in that way if false"; I think if two behaviours are distinct enough they should have they own classes (```BufferedFileSink```?).

The second point is the task that is started to process what is put in the queue. I don't like it too much either and I'm wondering if there would be a way to use the explicit implementation of ```IProducerConsumerCollection<T>```. I didn't look into this too much so I will explore, but maybe someone already has an idea.

Let me know what you think.

### DIFF
--- a/src/Serilog.Sinks.File/Sinks/File/FileSink.cs
+++ b/src/Serilog.Sinks.File/Sinks/File/FileSink.cs
@@ -30,6 +30,7 @@ namespace Serilog.Sinks.File
         const int BytesPerCharacterApproximate = 1;
         readonly TextWriter _output;
         readonly ITextFormatter _textFormatter;
+        readonly bool _buffered;
         readonly object _syncRoot = new object();
 
         /// <summary>Construct a <see cref="FileSink"/>.</summary>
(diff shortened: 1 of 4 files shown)


src/Serilog.Sinks.File/Sinks/File/FileSink.cs
```diff
@@ -88,17 +106,26 @@ static void TryCreateDirectory(string path)
...
+            else
+                EmitImpl(logEvent);
+        }
+
+        private void EmitImpl(LogEvent logEvent)
+        {
             lock (_syncRoot)
             {
                 _textFormatter.Format(logEvent, _output);
                 _output.Flush();
```
nblumhardt: Here is the `Flush()` call that `buffered: true` should switch off.

nblumhardt:
Hi Phillipe! Thanks for the PR.

The buffering I intended to refer to in the issue is the buffering provided under the hood by `System.IO.Stream` implementations.

> Both File and RollingFile flush the file stream after each write.

Turning "on" buffering would mean skipping the `Flush()` calls that these sinks make.

Sorry about the confusion, hope this clarifies it!

Pvlerick:
Oh I see now... Much simple than what I thought :smile: 

Ok then, I'll modify the PR accordingly.

Pvlerick:
@nblumhardt I reworked the approach a little bit. I added a ```autoFlush``` paramter to both ```FileSink``` and ```RollingFileSink``` that is relayed to the underlying ```StreamWriter```. With this, the call to ```Flush()``` can be suppressed and there is no need for an ```if``` statement.

Of course this parameter could be renamed to ```buffered``` and invert it when creating the ```StreamWriter```.

nblumhardt:
Thanks @Pvlerick!

A couple of thoughts -

 * `AutoFlush` will cause flushing after every write, which will be a lot more flushes than currently required; when events are formatted out as text there are a lot of fine-grained writes for each property etc. so the single `Flush()` call effectively batches these - using `AutoFlush` will most likely be a performance regression
 * Serilog by convention uses `false` as the default for all `Boolean` parameters, so `buffered = false` is preferred over `autoFlush = true` - sticking with this makes the API a bit more predictable

I think the `if()` version is probably the one to go with. All makes sense?

Cheers!

Pvlerick:
@nblumhardt makes sense. I updated the PR.

Pvlerick:
One more thing: should ```FileLoggerConfigurationExtensions.cs``` and ```RollingFileLoggerConfigurationExtensions.cs``` also be updated to take this into account, or is this left for advanced user that will not use the extension methods?

nblumhardt:
Looks great :+1:

I think this is a common enough scenario we should update the extension methods.

Pvlerick:
@nblumhardt all done.

nblumhardt:
Looking great! Nicely done.

PR #670, opened 2016-02-18
Moved AppSettings support to Serilog.Settings.AppSettings

nblumhardt:
From #593.

Includes some more conditional compilation cleanup and housekeeping.


merbla:
Just wondering about the project name of `Serilog.Settings.AppSettings`. Feel like there should be a config in there somewhere... 

Perhaps _Serilog.Configuration.AppSettings_

I am terrible at naming :see_no_evil:

IanYates:
:+1:

nblumhardt:
Yeah, it's not ideal. Unfortunately _Serilog.Configuration.*_ is taken by the `LoggerConfiguration` support classes, also public and under that namespace.

Hence when "settings" support was added to Serilog, it came in under _Serilog.Settings.*_.

I'm not a big fan either - _Serilog.Settings.KeyValuePairs_ doesn't seem to bad, but the _AppSettings_ one is a bit ugly. (_Serilog.Settings.Xml_ would be nicer to read but less accurate.)

I'm hopeful that the .NET Core one can get a better name - _Serilog.Settings.SomethingWithoutTheWordSettingsInIt_... but it's not coming to me. _Serilog.Settings.Configuration_ is probably the most accurate, since the MS library is _Microsoft.Extensions.Configuration_.

In that case, _Serilog.Configuration.Configuration_ is a worse place to be stuck :-)

To cut a long story short, I think the ugly name in isolation is indeed ugly, but in the bigger picture I think it's the best option we've got. Open to other suggestions of course :-)

IanYates:
Good points.  Naming is hard but I always like a good history story.  Your explanation was bordering on an *old new thing* style post :D

nblumhardt:
LOL :D ... I'm realising the project has been around long enough now to accumulate this kind of baggage......

PR #671, opened 2016-02-18
A tiny internal change - make `Logger` implement `ILogEventSink` explicitly

nblumhardt:
...so that it's clearly an implementation detail and not something we depend on accidentally.

### DIFF
--- a/src/Serilog/Core/Pipeline/Logger.cs
+++ b/src/Serilog/Core/Pipeline/Logger.cs
@@ -148,7 +148,7 @@ namespace Serilog.Core.Pipeline
             Dispatch(logEvent);
         }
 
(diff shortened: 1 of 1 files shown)


khellang:
Hmmm. We've been bad at assigning stuff to milestones. What's the etiquette here?

nblumhardt:
@khellang great point on milestones. I've been thinking we need a more disciplined approach to using this tracker, but I don't have a model in mind. At least in the 2.0 dev branch it's been a free-for-all so we'll need to piece together some release notes this time around...

Do you know if there's a good methodology out there in a similar project that we can adopt? (Ideally a documented one?)

Open to ideas :+1:

PR #678, opened 2016-02-23
Fixing ColoredConsoleSink to prevent printing empty properties

### LINKED ISSUE #649: ColoredConsoleSink prints missing output properties
Normally, when sinks are configured with an output template `{X}` and property `X` is missing from the event, nothing is rendered. The `ColoredConsoleSink` has its own custom rendering and retains the old behaviour, printing `{X}` literally to show the missing property.

http://stackoverflow.com/questions/35014396/is-it-possible-to-only-display-sourcecontext-in-serilog-if-usecontextclass-has

Pvlerick:
Fixes #649,

Strictly speaking, this is the _minimum_ in order to fix the issue. It uses the same lines as in ```MessageTemplateTextFormatter```, although in this case the if could be inverted.

I don't like it too much because this feels like duplicated logic, on the other hand ```ColoredConsoleSink``` needs to de-construct the message in order to apply different colours for different tokens.

### DIFF
--- a/src/Serilog.Sinks.ColoredConsole/Sinks/ColoredConsole/ColoredConsoleSink.cs
+++ b/src/Serilog.Sinks.ColoredConsole/Sinks/ColoredConsole/ColoredConsoleSink.cs
@@ -20,7 +20,7 @@ using System.Collections.Generic;
 using Serilog.Formatting.Display;
 using Serilog.Parsing;
 
-using IPropertyDictionary = System.Collections.Generic.IReadOnlyDictionary<string, Serilog.Events.LogEventPropertyValue>; 
+using IPropertyDictionary = System.Collections.Generic.IReadOnlyDictionary<string, Serilog.Events.LogEventPropertyValue>;
 
 namespace Serilog.Sinks.SystemConsole
 {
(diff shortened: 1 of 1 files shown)


src/Serilog.Sinks.ColoredConsole/Sinks/ColoredConsole/ColoredConsoleSink.cs
```diff
@@ -91,7 +91,12 @@ public void Emit(LogEvent logEvent)
                         {
                             RenderOutputToken(palette, outputToken, outputProperties, output);
                         }
-                        else switch (propertyToken.PropertyName)
+                        else
+                        {
+                            LogEventPropertyValue propertyValue;
+                            if (!outputProperties.TryGetValue(propertyToken.PropertyName, out propertyValue))
+                                continue;
```
khellang: Nit: Can we wrap this in braces? :stuck_out_tongue:

khellang:
LGTM! :shipit

src/Serilog.Sinks.ColoredConsole/Sinks/ColoredConsole/ColoredConsoleSink.cs
Pvlerick: Yes, or even inverting the if to get rid of the continue. I'll try both and see which one looks the pretties :wink:

khellang: If you decide to keep `continue`, you can remove the redundant `else` :smile:

nblumhardt:
:+1: also - ready to go @Pvlerick ?

Pvlerick:
@khellang @nblumhardt all done!

Note: the two extra lines modified at the top and bottom were beause I did an auto-format on the document.

Issue #695, opened 2016-03-15
Move sinks and enrichers to new repositories


merbla:
**Enrichers**
- ~~[Environment](https://github.com/serilog/serilog-enrichers-environment)~~
- ~~[Process](https://github.com/serilog/serilog-enrichers-process)~~
- ~~[Thread](https://github.com/serilog/serilog-enrichers-thread)~~

**Sinks**
- ~~[Console](https://github.com/serilog/serilog-sinks-console)~~
- ~~[Colored Console](https://github.com/serilog/serilog-sinks-coloredconsole)~~
- ~~[File](https://github.com/serilog/serilog-sinks-file)~~
- ~~[Rolling File](https://github.com/serilog/serilog-sinks-rollingfile)~~
- ~~[Trace](https://github.com/serilog/serilog-sinks-trace)~~
- ~~[Text Writer](https://github.com/serilog/serilog-sinks-textwriter)~~
- ~~[Period Batching](https://github.com/serilog/serilog-sinks-periodicbatching)~~
- ~~[Observable](https://github.com/serilog/serilog-sinks-observable)~~

PR #696, opened 2016-03-15
Removed console sink to new repo 

merbla:
Removed the console sink now available at https://github.com/serilog/serilog-sinks-console.

- Need to get this in, then enable publish of packages from new console sink project in AppVeyor.
- No tests impacted with this being removed.
- First project to move as apart of #695


merbla:
ping @serilog/core 

Also please have a look at https://github.com/serilog/serilog-sinks-console as the template for sink and enricher projects.  

- Issues enabled
- Wiki disabled

nblumhardt:
Great stuff!

Solution naming convention doesn't track the projects; `serilog-sinks-console.sln` would match the other repos better. Ot.her than that, LGTM.

One thing we may consider changing down the track - the sink packages probably should have a `projectUrl` on GitHub, rather than http://serilog.net like the core package does (a bit more useful in the new model).

PR #697, opened 2016-03-15
Rolling log files to next sequence based on file size limit

mr-sbux:
Related to #428.

We had a need to roll log files based on size rather than letting them grow unbounded until the next checkpoint, so here is a proposed fix that enables size-based file rolling.

### DIFF
--- a/src/Serilog.FullNetFx/Sinks/IOFile/CharacterCountLimitedTextWriter.cs
+++ b/src/Serilog.FullNetFx/Sinks/IOFile/CharacterCountLimitedTextWriter.cs
@@ -31,6 +31,14 @@ namespace Serilog.Sinks.IOFile
             _remainingCharacters = remainingCharacters;
         }
 
+        public bool CharacterLimitExceeded
+        { 
+            get
+            {
+                return _remainingCharacters < 0;
+            }
+        }
+
         public override Encoding Encoding
         {
             get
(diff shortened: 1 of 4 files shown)


merbla:
Hi @mr-sbux, 

Thanks for the PR.  A few items.

- In preparation for support for [DotNetCore](https://github.com/serilog/serilog/issues?q=is%3Aissue+is%3Aopen+label%3Av2) we have moved this sink to a new repo -https://github.com/serilog/serilog-sinks-rollingfile.  We are in the process of removing the Rolling Sink from this repo to its own dedicated location.
- On the larger repositories we have a `dev` branch.  This is where we ask people to target their pull requests to allow some integration between changes prior to a release.

If you would like to address this PR over on the new repo that is the place where this sink will evolve and change.

nblumhardt:
@mr-sbux @merbla just a thought, it'd be good to discuss the PR here in advance of re-creating it - I'm not sure at this point it's a goal of the built-in rolling file to support this.

https://github.com/BedeGaming/sinks-rollingfile covers this functionality and is already out there, so may be a better place to contribute improvements? Keen to hear thoughts.

mr-sbux:
Hey, folks!

@nblumhardt Yep, it looks like this is a solved problem. We can just drop this PR.

Thanks!

PR #699, opened 2016-03-17
Removed remaining sinks and specific sink tests.

merbla:
Relates to #695.  The following sinks have been moved:

- Rolling File
- File
- Text writer
- Periodic Batching
- Observable

Enrichers are next

ping @serilog/core for a look.

Tests are moved to the new sink repositories.  There are some tests still here in core that rely on sink packages due to appsettings/config tests.  Not sure where these live long term.


khellang:
You're doing some awesome work here, @merbla! :shipit::+1:

nblumhardt:
:+1:

I think we'll need to follow up with breaking the package dependencies to avoid surprises (e.g. accidentally testing the version of Serilog depended on by those packages, rather than the `{"target": "project"}` one) but LGTM for this round!

PR #711, opened 2016-04-04
add fixed width functionality for log event output.

iamkoch:
#574 Now supports {Level,1} style syntax. Also supports upper/lower variations.

### DIFF
--- a/src/Serilog/Formatting/Display/LogEventLevelValue.cs
+++ b/src/Serilog/Formatting/Display/LogEventLevelValue.cs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 using System;
-using System.Collections.Generic;
 using System.IO;
 
 using Serilog.Events;
@@ -27,17 +26,16 @@ namespace Serilog.Formatting.Display
     {
         readonly LogEventLevel _value;
 
-        // Assumes reading from Dictionary<,> is thread-safe after construction
-        static readonly Dictionary<LogEventLevel, string[]> _shortenedLevelMap = new Dictionary<LogEventLevel, string[]>
-                                                                                  {
-                                                                                      { LogEventLevel.Verbose, new []{ "V", "Vb", "Vrb", "Verb" } },
-                                                                                      { LogEventLevel.Debug, new []{ "D", "De", "Dbg", "Dbug" } },
-                                                                                      { LogEventLevel.Information, new []{ "I", "In", "Inf", "Info" } },
-                                                                                      { LogEventLevel.Warning, new []{ "W", "Wn", "Wrn", "Warn" } },
-                                                                                      { LogEventLevel.Error, new []{ "E", "Er", "Err", "Eror" } },
-                                                                                      { LogEventLevel.Fatal, new []{ "F", "Fa", "Ftl", "Fatl" } },
-                                                                                  };
-
+        static readonly string[][] _shortenedLevelMap 
+            = {
+                new []{ "V", "Vb", "Vrb", "Verb" },
+                new []{ "D", "De", "Dbg", "Dbug" },
+                new []{ "I", "In", "Inf", "Info" },
+                new []{ "W", "Wn", "Wrn", "Warn" },
+                new []{ "E", "Er", "Err", "Eror" },
+                new []{ "F", "Fa", "Ftl", "Fatl" }
+              };
+        
         public LogEventLevelValue(
             LogEventLevel value)
         {
@@ -101,7 +99,11 @@ namespace Serilog.Formatting.Display
 
         private static string ShortLevelFor(LogEventLevel value, int width)
         {
-            return _shortenedLevelMap[value][width - 1];
+            var index = (int)value;
+            if (index < 0 || index > (int)LogEventLevel.Fatal)
+                return string.Empty;
+
+            return _shortenedLevelMap[index][width - 1];
         }
 
         private static bool IsOutputStringTooWide(Alignment alignmentValue, string formattedValue)
--- a/src/Serilog/Formatting/Display/LogEventLevelValue.cs
+++ b/src/Serilog/Formatting/Display/LogEventLevelValue.cs
@@ -1,0 +1,99 @@
+// Copyright 2013-2015 Serilog Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+using Serilog.Events;
+using Serilog.Parsing;
+
+namespace Serilog.Formatting.Display
+{
+    // allows for the specific handling of the {Level} element.
+    // can now have a fixed width applied to it, as well as casing rules.
+    class LogEventLevelValue : LogEventPropertyValue
+    {
+        private readonly LogEventLevel _value;
+        private readonly Alignment? _alignment;
+
+        private Dictionary<LogEventLevel, string[]> _shortenedLevelMap
+            = new Dictionary<LogEventLevel, string[]>
+                  {
+                      { LogEventLevel.Verbose, new []{ "V", "Vb", "Vrb", "Verb" } },
+                      { LogEventLevel.Debug, new []{ "D", "De", "Dbg", "Dbug" } },
+                      { LogEventLevel.Information, new []{ "I", "In", "Inf", "Info" } },
+                      { LogEventLevel.Warning, new []{ "W", "Wn", "Wrn", "Warn" } },
+                      { LogEventLevel.Error, new []{ "E", "Er", "Err", "Eror" } },
+                      { LogEventLevel.Fatal, new []{ "F", "Fa", "Ftl", "Fatl" } },
+                  };
+
+        public LogEventLevelValue(
+            LogEventLevel value,
+            Alignment? alignment)
+        {
+            _value = value;
+            _alignment = alignment;
+        }
+
+        public string FormattedValue()
+        {
+            var formattedValue = _value.ToString();
+
+            if (!_alignment.HasValue || _alignment.Value.Width <= 0)
+            {
+                return formattedValue;
+            }
+
+            var alignmentValue = _alignment.Value;
+
+            if (formattedValue.Length == alignmentValue.Width)
+            {
+                return formattedValue;
+            }
+
+            if (IsCustomWidthSupported(alignmentValue.Width))
+            {
+                return ShortLevelFor(alignmentValue);
+            }
+
+            if (IsOutputStringTooWide(alignmentValue, formattedValue))
+            {
+                return formattedValue.Substring(0, alignmentValue.Width);
+            }
+
+            return formattedValue;
+        }
+
+        public override void Render(TextWriter output, string format = null, IFormatProvider formatProvider = null)
+        {
+            new LiteralStringValue(FormattedValue()).Render(output, format, formatProvider);
+        }
+
+        private string ShortLevelFor(Alignment alignmentValue)
+        {
+            return _shortenedLevelMap[_value][alignmentValue.Width - 1];
+        }
+
+        private static bool IsOutputStringTooWide(Alignment alignmentValue, string formattedValue)
+        {
+            return alignmentValue.Width < formattedValue.Length;
+        }
+
+        private static bool IsCustomWidthSupported(int width)
+        {
+            return width > 0 && width < 5;
+        }
+    }
+}
--- a/src/Serilog/Formatting/Display/LogEventLevelValue.cs
+++ b/src/Serilog/Formatting/Display/LogEventLevelValue.cs
@@ -23,21 +23,20 @@ namespace Serilog.Formatting.Display
 {
     // allows for the specific handling of the {Level} element.
     // can now have a fixed width applied to it, as well as casing rules.
-    internal class LogEventLevelValue : LogEventPropertyValue
+    class LogEventLevelValue : LogEventPropertyValue
     {
-        private readonly LogEventLevel _value;
+        readonly LogEventLevel _value;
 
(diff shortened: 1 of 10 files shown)


src/Serilog/Parsing/PropertyToken.cs
```diff
@@ -123,6 +124,11 @@ public override void Render(IReadOnlyDictionary<string, LogEventPropertyValue> p
         }
 
         /// <summary>
+        /// Return whether this is the Level output token
+        /// </summary>
+        public bool IsLevelProperty => OutputProperties.LevelPropertyName.Equals(PropertyName);
```
nblumhardt: This is probably  better kept in `OutputProperties`

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
```diff
@@ -91,6 +91,15 @@ public void Format(LogEvent logEvent, TextWriter output)
 
                     token.Render(overridden, output, _formatProvider);
                 }
+                else if (pt.IsLevelProperty)
```
nblumhardt: Instead of doing this here, we can just return a `LogEventLevelValue` from https://github.com/serilog/serilog/blob/dev/src/Serilog/Formatting/Display/OutputProperties.cs#L68

nblumhardt: I see that the alignment still needs to be passed through; might still need the conditional but rather than `pi.IsLevelProperty` it could be `pt.Value is LogEventLevelValue`? (using the as/null check pattern if possible)

nblumhardt: (In general, allocating objects is something to avoid in the logging pipeline, `Dictionary<,>` turns out to be particularly expensive.)

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public LogEventLevelValue(
+            LogEventLevel value,
+            Alignment? alignment)
+        {
+            _value = value;
+            _alignment = alignment;
+        }
+
+        public string FormattedValue()
```
nblumhardt: This could be treated as an overload of `Render()`, e.g.:

```csharp
public void Render(TextWriter output, Alignment? alignment, string format = null, IFormatProvider formatProvider = null)
```

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                return formattedValue.Substring(0, alignmentValue.Width);
+            }
+
+            return formattedValue;
+        }
+
+        public override void Render(TextWriter output, string format = null, IFormatProvider formatProvider = null)
+        {
+            new LiteralStringValue(FormattedValue()).Render(output, format, formatProvider);
```
nblumhardt: To avoid having to allocate a `LiteralStringValue` just to call `Render()` on it, could they share some (static) methods for formatting?

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,99 @@
...
+using Serilog.Events;
+using Serilog.Parsing;
+
+namespace Serilog.Formatting.Display
+{
+    // allows for the specific handling of the {Level} element.
+    // can now have a fixed width applied to it, as well as casing rules.
+    class LogEventLevelValue : LogEventPropertyValue
+    {
+        private readonly LogEventLevel _value;
```
nblumhardt: The codebase doesn't generally use explicit `private` modifiers.

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,99 @@
...
+        private readonly Alignment? _alignment;
+
+        private Dictionary<LogEventLevel, string[]> _shortenedLevelMap
```
nblumhardt: This needs to be `static`, otherwise every rendering will end up creating one of these.

Can also be `readonly`.

May need a comment `// Assumes reading from Dictionary<,> is thread-safe after construction` (which I believe is a valid assumption currently.

nblumhardt:
This is awesome, thanks Anthony!

I've added a few comments on the implementation, but I think overall the API and functionality are perfect.

Avoiding allocations is the name of the game here, so most of my comments are along those lines.

If `OutputProperties.GetOutputProperties()` could return the custom level property value, we'll cut out the need for `IsLevelProperty` on the core `PropertyToken` type, which is a bit of a backreference in the current factoring scheme.

It seems that because there's an alignment involved, the `if` block will then look something like:

```csharp
else if (pt.Value is LevelPropertyValue)
{
   var lv = (LevelPropertyValue)pt.Value;
   lvt.Render(output, pt.Alignment, pt.Format, _formatProvider);
}
```

It breaks the abstraction around `pt.Render()` but it's a bit clearer and cheaper. What do you think?

src/Serilog/Formatting/Display/LogEventLevelValue.cs
iamkoch: The class I copy pasted had that! I'll update mine and it accordingly.

src/Serilog/Formatting/Display/LogEventLevelValue.cs
iamkoch: Ah yes - good spot.

iamkoch:
I had what you describe originally, however the render method with zero functionality was what put me off. If you're OK with a no op I'll switch it back

iamkoch:
I've made most of the changes, however I have lost the padding logic that currently resides within property token. I'll look to move that out too.

iamkoch:
I've modified the PR. I'm not entirely sure about the usage of the static method for padding, so I'll await your feedback. I've tried to avoid assignations where possible, understanding performance is preferable to readability, but hopefully what's in place is clear enough.

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
```diff
@@ -91,6 +91,11 @@ public void Format(LogEvent logEvent, TextWriter output)
 
                     token.Render(overridden, output, _formatProvider);
                 }
+                else if (OutputProperties.IsLevelProperty(pt))
```
nblumhardt: Using `is`/`as` will be more efficient than the name check and subsequent cast here, and gets rid of the `IsLevelProperty` method. E.g.:

```
else
{
    var lp = propertyValue as LogEventLevelValue;
    if (lp != null)
    {
        lp.Render(...)
```

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        private string AlignedValue(Alignment? alignment)
+        {
+            if (!alignment.HasValue || alignment.Value.Width <= 0)
+            {
+                return _value.ToString();
+            }
+
+            if (_value.ToString().Length == alignment.Value.Width)
```
nblumhardt: This `if`/block isn't needed, since the behaviour will be the same without it - if the value has a custom width we'll use that in the next block, if it's too wide we'll trim it in the one after, or finally, we'll do what this block does and return it as `ToString()`.

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                return _value.ToString();
+            }
+
+            if (IsCustomWidthSupported(alignment.Value.Width))
+            {
+                return ShortLevelFor(_value, alignment.Value);
```
nblumhardt: Since the second arg is only providing a width, passing `alignment.Value.Width` here as an integer parameter might make `ShortLevelFor()` more obvious.

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+
+            if (IsOutputStringTooWide(alignment.Value, _value.ToString()))
```
nblumhardt: Adding:

```csharp
var stringValue = _value.ToString();
```

on the line above and then using it in all of the subsequent places `_value.ToString()` is used will reduce the number of calls required.

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,116 @@
...
+using Serilog.Events;
+using Serilog.Parsing;
+
+namespace Serilog.Formatting.Display
+{
+    // allows for the specific handling of the {Level} element.
+    // can now have a fixed width applied to it, as well as casing rules.
+    internal class LogEventLevelValue : LogEventPropertyValue
+    {
+        private readonly LogEventLevel _value;
```
nblumhardt: Still a bunch of explicit `private` modifiers in this file.

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,116 @@
...
+                      { LogEventLevel.Verbose, new []{ "V", "Vb", "Vrb", "Verb" } },
+                      { LogEventLevel.Debug, new []{ "D", "De", "Dbg", "Dbug" } },
+                      { LogEventLevel.Information, new []{ "I", "In", "Inf", "Info" } },
+                      { LogEventLevel.Warning, new []{ "W", "Wn", "Wrn", "Warn" } },
+                      { LogEventLevel.Error, new []{ "E", "Er", "Err", "Eror" } },
+                      { LogEventLevel.Fatal, new []{ "F", "Fa", "Ftl", "Fatl" } },
+                  };
+
+        public LogEventLevelValue(
+            LogEventLevel value)
```
nblumhardt: nit: formatting

src/Serilog/Formatting/Display/Padding.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.IO;
+
+using Serilog.Parsing;
+
+namespace Serilog.Formatting.Display
+{
+    internal static class Padding
```
nblumhardt: Explicit `internal`

src/Serilog/Formatting/Display/Padding.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <summary>
+        /// Writes the provided value to the output, applying direction-based padding when <paramref name="alignment"/> is provided.
+        /// </summary>
+        public static void Apply(TextWriter output, string value, Alignment? alignment)
```
nblumhardt: Since the method deals with padding for anything and not just `Alignment`, passing an integer here and avoiding calls when there's no alignment specified seems clearer.

nblumhardt:
Looking great. Combed through it pretty closely since it'll be called for pretty much every event that ever passes through Serilog, so please excuse some of the nit-picking :-)

The only other thought left in my mind is whether we should add uppercase and lowercase versions of the short-string table; doing this would avoid all the allocations when `ToUpperInvariant()`/`ToLowerInvariant()` are called. May be a "future improvement" to consider.

src/Serilog/Formatting/Display/LogEventLevelValue.cs
```diff
@@ -0,0 +1,116 @@
...
+        private static readonly Dictionary<LogEventLevel, string[]> _shortenedLevelMap
+            = new Dictionary<LogEventLevel, string[]>
+                  {
+                      { LogEventLevel.Verbose, new []{ "V", "Vb", "Vrb", "Verb" } },
+                      { LogEventLevel.Debug, new []{ "D", "De", "Dbg", "Dbug" } },
+                      { LogEventLevel.Information, new []{ "I", "In", "Inf", "Info" } },
+                      { LogEventLevel.Warning, new []{ "W", "Wn", "Wrn", "Warn" } },
+                      { LogEventLevel.Error, new []{ "E", "Er", "Err", "Eror" } },
+                      { LogEventLevel.Fatal, new []{ "F", "Fa", "Ftl", "Fatl" } },
+                  };
```
nblumhardt: One other option here that will be a bit more efficient:

```csharp
        static readonly string[][] _shortenedLevelMap
            = new string[][]
                  {
                      new []{ "V", "Vb", "Vrb", "Verb" },
                      new []{ "D", "De", "Dbg", "Dbug" },
                      new []{ "I", "In", "Inf", "Info" },
                      new []{ "W", "Wn", "Wrn", "Warn" },
                      new []{ "E", "Er", "Err", "Eror" },
                      new []{ "F", "Fa", "Ftl", "Fatl" },
                  };
```

And then:

```csharp
        static string ShortLevelFor(LogEventLevel value, int width)
        {
            var idx = (int)value;
            if (idx < 0 || idx > (int)(LogEventLevel.Fatal))
                return "";
            return _shortenedLevelMap[idx][width - 1];
        }
```

The check might be needed even in the dictionary-based code, since it's possible (although) unlikely, to get a level like `((LogEventLevel) 42)`.

iamkoch:
Hi Nick

Just a not to say sorry for the delay in replying - I'll pick up on your comments properly in the next day or two.

nblumhardt:
@iamkoch no worries! Looking forward to merging this one though, great feature :-)

src/Serilog/Formatting/Display/Padding.cs
iamkoch: The alignment contains the direction though, which is required in the method body

nblumhardt: Ah I see, right you are!

nblumhardt:
How's everything going? Yell out if you're short for time on this, I can also pick up some of the nitty things post-merge if that would help. I appreciate that this one has been a marathon! :-)

src/Serilog/Formatting/Display/LogEventLevelValue.cs
iamkoch: We use alignment.Value without null checks though, so this guards against a null reference error, hence its inclusion

iamkoch:
Hi Nick

Will get this tidied up today. Have pushed a few more things and will fix the dictionary on my lunch break :)

iamkoch:
Hi Nick

I've hopefully updated based on all your comments and left the once we've discussed.

Let me know if there's anything left.

The only thing might be the formatting of the multi dimensional array, but I think it's quite readable now, and hopefully you do too! :)

nblumhardt:
Awesome! It looks great.

I only have a couple of trivial formatting nits, but I'll spare us another round-trip and apply them on the way in :-)

Issue #718, opened 2016-04-18
Closing/flushing `PeriodicBatchingSink` in the `AppDomain.UnhandledException` event eats messages


nblumhardt:
The change introduced in https://github.com/serilog/serilog/pull/366 was intended to flush the log whenever the process was about to exit, via an unhandled exception or otherwise.

Unfortunately, by _closing_ the logger, this means exceptions logged _in the UnhandledException handler_ never make it through this sink.

Requires a fix here and in https://github.com/serilog/serilog-sinks-periodicbatching.

Thank you @chriswithpants for helping to _flush_ this one out :-)

nblumhardt:
Note, the 2.0 design for .NET core doesn't use this mechanism; perhaps in the v2 sink we clobber it entirely (don't subscribe to any of the events even with `AppDomain` available), and just remove the handler in the 1.5 patch?

nblumhardt:
(Practically all non-local sinks are impacted.)

nblumhardt:
I've submitted the two PRs above to address the 2.0 case, where we can afford to make the minor breaking changes this entails.

Making a change in 1.5 may not be feasible, because although some apps will be broken by this behaviour, others will be relying on it. A fix by introducing a `Flush()` in `PeriodicBatchingSink` could work here, but the implementation would need significant rework to support it because of the limited APIs available for controlling the `Timer`.

A **workaround** has been confirmed to fix the issue in 1.5: applications should make sure to attach to the `AppDomain.UnhandledException` event _before_ configuring the Serilog logger. This will (at least on the desktop .NET) ensure the handler runs before the sink's handler, meaning the close/flush sequence will be correct.

nblumhardt:
Because of the possible wide breakage I think 2.0 _is_ the fix for this. Not expecting to move this one on further here.

PR #719, opened 2016-04-18
Make the return value of `LoggerConfiguration.CreateLogger()` concrete

nblumhardt:
This PR communicates the dispose requirement of the logging pipeline by returning a concrete disposable type, `Logger`, from `CreateLogger()`.

One consequence is that `using` syntax can be applied:

```csharp
using (var log = new LoggerConfiguration()
    .WriteTo.Console()
    .CreateLogger())
{
    // <app runs here>
}
```

It's not anticipated that users will generally wrap logger creation in a `using` block however; rather in 2.0 the `CloseAndFlush()` method will cover apps that base logging off of the static `Log` class.

```csharp
    .WriteTo.Console()
    .CreateLogger();

// <app runs here>

Log.CloseAndFlush();
```

Doing this however clarifies at the API level that the logger needs to be flushed in some way. This is important as .NET Core doesn't expose any `AppDomain` unload-time hooks to trigger flushing, which historically we've relied on.

See also #718, where working around this design issue has lead us astray.

## Caveats

This regresses one minor point - a logger with no sinks configured is no longer switched for a `SilentLogger` behind the scenes; users relying on this behaviour instead need to avoid configuring the logger at all. The alternative - returning something that could be shared in common with `SilentLogger`, would require one of the less-appealing options discussed in _Alternatives_ below.

Another caveat is that in order to prevent a public-surface-area explosion, the constructors of this type are internal. This should not be a significant issue as `ILogger` should be the interface used in application code.

## Alternatives

Instead of exposing the concrete `Logger` class, the following options were also available:

 * Return a disposable interface - `ILoggingPipeline` or similar: disposable interfaces are a semantic wart since only implementations determine the need for disposal
 * Return a new abstract base class - this adds the need for virtual dispatch in more places, for example level checking, which adds up if not kept under control

### DIFF
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -20,9 +20,9 @@ using Serilog.Debugging;
 using Serilog.Events;
 using Serilog.Parameters;
 
-namespace Serilog.Core.Pipeline
+namespace Serilog.Core
 {
-    sealed class Logger : ILogger, ILogEventSink, IDisposable
+    public sealed class Logger : ILogger, ILogEventSink, IDisposable
     {
         readonly MessageTemplateProcessor _messageTemplateProcessor;
         readonly ILogEventSink _sink;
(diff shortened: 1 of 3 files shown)


merbla:
👍 I like this.  It is clear where the responsibility lies.  I have been caught out not closing and flushing loggers.  This clears up this problem.

:shipit:

Issue #723, opened 2016-04-25
Consider a FormatterConfiguration


Mike-EEE:
I encountered this problem while trying to format a `MethodBase` object in my entries.  It would be nice to be able to provide custom configurations for formatting objects.  That is, whenever an object of a certain type is encountered, Serilog looks to a formatting configuration of some sort to see how to format the object.

So something like:

```
var logger = new LoggerConfiguration()
             .Format
             .With<MethodBase>( m => $"{m.DeclaringType.Name}.{m.Name}" )
             .CreateLogger();
```

Then in the [ScalarValue rendering](https://github.com/serilog/serilog/blob/dev/src/Serilog/Events/ScalarValue.cs#L74), it checks to see if a registration like the above has been made.  If it has, then use it, otherwise, continue with existing functionality.

Thank you for any consideration!

nblumhardt:
Hi Mike,

This is essentially `IScalarConversionPolicy`, which exists in Serilog today but isn't exposed fully via the API. It'd be great to check out what is involved in providing some configuration for it.

 * https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/IScalarConversionPolicy.cs
 * https://github.com/serilog/serilog/blob/dev/src/Serilog/Parameters/PropertyValueConverter.cs#L114

(There are some perf tradeoffs to examine if more than a few of these are expected to be plugged in per app, the current impl is O(n) which is net positive when n <= 7 vs a dictionary, but getting a dictionary in for larger numbers of items will need some clever synchronization to avoid any additional locks/fences...)

Any investigation/proposals would be great! :-)

cc @adamchester

Mike-EEE:
Excellent @nblumhardt!  I will check that out and update this thread on any findings/successes/questions.  Thanks for the pointer!

Mike-EEE:
OK I see what you are saying now.  It is indeed available, but not exposed.  Essentially it would involve adding a new parameter to the [`PropertyValueConverter` Constructor`](https://github.com/serilog/serilog/blob/dev/src/Serilog/Parameters/PropertyValueConverter.cs#L50) and passing it in when the `LoggerConfiguration` creates the `ILogger`.  Since it will involve some work, I still gravitate towards the addition of a new configuration object (`Format` or `Formatting`) and add some awesome magic there (that could ultimately tie into creating `IScalarConversionPolicy` implementations).

nblumhardt:
[...]

PR #727, opened 2016-04-29
Internalize IScalarConversionPolicy, since no API allows one to be plugged in

nblumhardt:
This interface is public, but no API allows an instance to be plugged in, so it is unusable as-is.

Internally, the implementation around this interface may have to change if the feature is ever opened up. Let's defer committing to a design for the public interface until the rest of the feature is fully planned out (if ever).

See discussion in #723.


merbla:
LGTM.

PR #728, opened 2016-04-30
Allow destructuring policies to transform reflection types

adamchester:
See #723 for details.

### DIFF
--- a/src/Serilog/Parameters/PropertyValueConverter.cs
+++ b/src/Serilog/Parameters/PropertyValueConverter.cs
@@ -61,13 +61,13 @@ namespace Serilog.Parameters
                 new NullableScalarConversionPolicy(),
                 new EnumScalarConversionPolicy(),
                 new ByteArrayScalarConversionPolicy(),
-                new ReflectionTypesScalarConversionPolicy()
             };
 
             _destructuringPolicies = additionalDestructuringPolicies
-                .Concat(new []
+                .Concat(new IDestructuringPolicy []
                 {
-                    new DelegateDestructuringPolicy()
+                    new DelegateDestructuringPolicy(),
+                    new ReflectionTypesScalarDestructuringPolicy()
                 })
                 .ToArray();
         }
--- a/src/Serilog/Policies/ReflectionTypesScalarDestructuringPolicy.cs
+++ b/src/Serilog/Policies/ReflectionTypesScalarDestructuringPolicy.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2013-2015 Serilog Contributors
+﻿// Copyright 2013-2016 Serilog Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


nblumhardt:
This looks like the right thing to do :+1:

I need to run my eyes over it again with some more time, will aim to have this turned around tomorrow morning.

src/Serilog/ReflectionTypesScalarDestructuringPolicy.cs
```diff
@@ -0,0 +1,29 @@
...
+using Serilog.Core;
+using Serilog.Events;
+
+namespace Serilog.Policies
+{
+    class ReflectionTypesScalarDestructuringPolicy : IDestructuringPolicy
+    {
+        public bool TryDestructure(object value, ILogEventPropertyValueFactory propertyValueFactory, out LogEventPropertyValue result)
+        {
+            var type = value as System.Type;
```
nblumhardt: Any reason not to keep the impl from above? Seems equivalent and a bit more terse:

```csharp
           if (value is Type || value is MemberInfo)
           {
               result = new ScalarValue(value);
               return true;
           }

            result = null;
            return false;
```

adamchester: Good point, will do that. Original is clearer.

adamchester:
Since then, I've thought this PR title is misleading; reflection types are still scalar by default after these changes.

src/Serilog/ReflectionTypesScalarDestructuringPolicy.cs
adamchester: Force-pushed in 29c14d0

giphy

src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
```diff
@@ -114,6 +115,24 @@ public LoggerConfiguration ByTransforming<TValue>(Func<TValue, object> transform
...
         /// <summary>
+        /// When destructuring objects, transform instances derived from the specified type with
+        /// the provided function.
+        /// </summary>
+        /// <param name="transformation">Function mapping instances of <typeparamref name="TValue"/>
+        /// to an alternative representation.</param>
+        /// <typeparam name="TValue">The base class Type of values to transform.</typeparam>
+        /// <returns>Configuration object allowing method chaining.</returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        public LoggerConfiguration ByTransformingDerived<TValue>(Func<TValue, object> transformation)
```
adamchester: This name might need some more thought

nblumhardt:
I think `ByTransformingDerived` is an interesting addition, but I'd prefer keeping it separate from this PR since the reflection types scenario is unblocked without this change. Works for you? Cheers!

adamchester:
OK - but the issue is that at runtime (in most cases I tested), you never actually get an instance of `System.Type` (you get, for example, an internal `System.RuntimeType`):

image

adamchester:
What that means, of course, is that you can't ever use `Destructure.ByTransforming` to match reflection types in reality.

nblumhardt:
Yep, gotcha :+1: - but with the first PR'd change it's possible to plug in an `IDestructuringPolicy` to cover the scenario, so still a step forward.

`ByTransformingDerived` is probably an improvement in itself, but looking at other places it'd be used might suggest some API tweaks or alternatives we won't spot just looking at reflection types. I'm also (guiltily) interested in the possibility of `ByTransforming(..., includeSubtypes: true)` etc. Trying to keep all of that discussion from derailing the PR here :-)

adamchester:
Ahh! Didn't remember `IDestructuringPolicy` was public :+1: 

Will remove `ByTransformingDerived` because it's possible to use without it.

adamchester:
Removed all trace of `ByTransformingDerived` now.

nblumhardt:
Awesome, thanks!

adamchester:
image

Issue #737, opened 2016-05-10
Improve the way malformed templates/arguments are handled


adamchester:
Consider the following scenario:

```csharp
log.Information("Hello, {name}, you have outcome}", "Adam", "Failed");
```
The message template above was accidentally malformed, missing the opening brace before `outcome}`. 
Serilog currently would recognise and match the `{name}` field, but would totally ignore the `{outcome}` field because it doesn't have a corresponding/matching valid placeholder field in the message template. Because we provided the 2nd argument value `Failed`, Serilog would write a warning to `SelfLog` as follows:
> 2016-05-11T09:32:39 Named property count does not match parameter count: Hello, {name}, you have outcome}

I believe Serilog should *include* the `Failed` value with some default/obvious name instead of dropping / ignoring the value.

Related:
logary/logary#119 and [MessageTemplates](https://github.com/adamchester/MessageTemplates)

adamchester:
:tada:

adamchester:
Wait, still going attempt something with positionals

nblumhardt:
Time to close this one? Cheers!

PR #741, opened 2016-05-14
Log parameters not in message templates

adamchester:
Opening PR for discussion. This attempts to address #737.

Consider the following scenario:
```csharp
log.Information("Hello, {name}, you have outcome}", "Adam", "Failed");
```

The message template above was accidentally malformed, missing the opening brace before outcome}. Serilog currently would recognise and match the {name} field, but would totally ignore the {outcome} field because it doesn't have a corresponding/matching valid placeholder field in the message template. Because we provided the 2nd argument value "Failed", Serilog would write a warning to SelfLog as follows:

> 2016-05-11T09:32:39 Named property count does not match parameter count: Hello, {name}, you have outcome}

I believe Serilog should include the "Failed" value with some default/obvious name instead of dropping / ignoring the value.

### Functionality
- [x] ~~Decide if it's a breaking change, and therefore if it needs to be in a major release~~ So far it doesn't seem to be - and probabaly will go into 2.0 anyway.
- [x] Decide how to name extra parameters that are not in the message template

### Testing strategy
- [x] ~~Decide if tests should work against the internal `PropertyBinder`~~ Seems OK
- [x] ~~Decide if `IEqualityComparer<LogEventProperty>` is appropriate~~ Seems OK

### DIFF
--- a/src/Serilog/Parameters/PropertyBinder.cs
+++ b/src/Serilog/Parameters/PropertyBinder.cs
@@ -70,10 +70,10 @@ namespace Serilog.Parameters
             var positionalProperties = template.PositionalProperties;
 
             var result = new LogEventProperty[messageTemplateParameters.Length];
-            for (int position = 0; position < messageTemplateParameters.Length; position++)
+            for (var position = 0; position < messageTemplateParameters.Length; position++)
             {
                 PropertyToken propertyToken = null;
-                for (int propIndex = 0; propIndex < positionalProperties.Length; propIndex++)
+                for (var propIndex = 0; propIndex < positionalProperties.Length; propIndex++)
                 {
                     int propertyTokenPosition;
                     if (positionalProperties[propIndex].TryGetPositionalValue(out propertyTokenPosition)
--- a/src/Serilog/Parameters/PropertyBinder.cs
+++ b/src/Serilog/Parameters/PropertyBinder.cs
@@ -51,18 +51,23 @@ namespace Serilog.Parameters
                 return NoProperties;
             }
 
-            if (messageTemplate.PositionalProperties != null)
-                return ConstructPositionalProperties(messageTemplate, messageTemplateParameters);
+            if (messageTemplate.PositionalProperties == null)
+                return ConstructNamedProperties(messageTemplate, messageTemplateParameters);
 
-            return ConstructNamedProperties(messageTemplate, messageTemplateParameters);
+            // all are positional
+            if (messageTemplateParameters.Length == messageTemplate.PositionalProperties.Length)
+            {
+                // exactly the right number of positional parameters were provided
+                return ConstructPositionalPropertiesOptimisedAllMatching(messageTemplate, messageTemplateParameters);
+            }
+
+            SelfLog.WriteLine("Positional property count does not match parameter count: {0}", messageTemplate);
+            return ConstructPositionalPropertiesWithSomeMissing(messageTemplate, messageTemplateParameters);
         }
 
-        IEnumerable<LogEventProperty> ConstructPositionalProperties(MessageTemplate template, object[] messageTemplateParameters)
+        IEnumerable<LogEventProperty> ConstructPositionalPropertiesWithSomeMissing(MessageTemplate template, object[] messageTemplateParameters)
         {
             var positionalProperties = template.PositionalProperties;
-
-            if (positionalProperties.Length != messageTemplateParameters.Length)
-                SelfLog.WriteLine("Positional property count does not match parameter count: {0}", template);
 
             var result = new LogEventProperty[messageTemplateParameters.Length];
             for (int position = 0; position < messageTemplateParameters.Length; position++)
@@ -79,9 +84,27 @@ namespace Serilog.Parameters
                 }
 
                 result[position] = propertyToken == null
-                    ? new LogEventProperty("__" + position, 
+                    ? new LogEventProperty("__" + position,
                         _valueConverter.CreatePropertyValue(messageTemplateParameters[position]))
                     : ConstructProperty(propertyToken, messageTemplateParameters[position]);
+            }
+
+            return result;
+        }
+
+        IEnumerable<LogEventProperty> ConstructPositionalPropertiesOptimisedAllMatching(MessageTemplate template, object[] messageTemplateParameters)
+        {
+            var positionalProperties = template.PositionalProperties;
+            var result = new LogEventProperty[positionalProperties.Length];
+            for (var index = 0; index < positionalProperties.Length; index++)
+            {
+                var property = positionalProperties[index];
+                int position;
+                if (!property.TryGetPositionalValue(out position))
+                {
+                    throw new Exception("should never happen here");
+                }
+                result[position] = ConstructProperty(property, messageTemplateParameters[position]);
             }
 
             return result;
--- a/src/Serilog/Parameters/PropertyBinder.cs
+++ b/src/Serilog/Parameters/PropertyBinder.cs
@@ -54,14 +54,6 @@ namespace Serilog.Parameters
             if (messageTemplate.PositionalProperties == null)
                 return ConstructNamedProperties(messageTemplate, messageTemplateParameters);
 
-            // all are positional
-            if (messageTemplateParameters.Length == messageTemplate.PositionalProperties.Length)
-            {
-                // exactly the right number of positional parameters were provided
-                return ConstructPositionalPropertiesOptimisedAllMatching(messageTemplate, messageTemplateParameters);
-            }
-
-            SelfLog.WriteLine("Positional property count does not match parameter count: {0}", messageTemplate);
             return ConstructPositionalPropertiesWithSomeMissing(messageTemplate, messageTemplateParameters);
         }
 
@@ -87,26 +79,6 @@ namespace Serilog.Parameters
                     ? new LogEventProperty("__" + position,
                         _valueConverter.CreatePropertyValue(messageTemplateParameters[position]))
                     : ConstructProperty(propertyToken, messageTemplateParameters[position]);
-            }
-
-            return result;
-        }
-
-        IEnumerable<LogEventProperty> ConstructPositionalPropertiesOptimisedAllMatching(MessageTemplate template, object[] messageTemplateParameters)
-        {
-            var positionalProperties = template.PositionalProperties;
-            var result = new LogEventProperty[positionalProperties.Length];
-            for (var index = 0; index < positionalProperties.Length; index++)
-            {
-                var property = positionalProperties[index];
-                int position;
-                if (!property.TryGetPositionalValue(out position))
-                {
-                    throw new ArgumentException(
-                        "This method can only process a template with all positional properties.", nameof(template));
-                }
-
-                result[position] = ConstructProperty(property, messageTemplateParameters[position]);
             }
 
             return result;
--- a/src/Serilog/Parameters/PropertyBinder.cs
+++ b/src/Serilog/Parameters/PropertyBinder.cs
@@ -65,30 +65,32 @@ namespace Serilog.Parameters
                 SelfLog.WriteLine("Positional property count does not match parameter count: {0}", template);
 
             var result = new LogEventProperty[messageTemplateParameters.Length];
-            foreach (var property in positionalProperties)
+            int nextUnmatchedPosition = 0;
+            for (int position = 0; position < messageTemplateParameters.Length; position++)
             {
-                int position;
-                if (property.TryGetPositionalValue(out position))
+                PropertyToken propertyToken = null;
+                for (int propIndex = 0; propIndex < positionalProperties.Length; propIndex++)
                 {
-                    if (position < 0 || position >= messageTemplateParameters.Length)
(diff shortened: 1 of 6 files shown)


nblumhardt:
Great stuff!

I wouldn't say it's a (noticeably) breaking change, but since 2.0 is still open it seems to be on target for that release anyway.

Regarding names, if it turns out to be sane, the ordinal naming we inherit from C# format strings is worth investigating. E.g. in your snippet above the property name would be `"1"`. It's language-agnostic, and consumers already have to deal with these because of format string compatibility.

```csharp
log.Information("Hello, {0, you have {1}", "Adam", "Failed");
```

^ here's the counterexample where `1` will already appear in the event :-| Ultimately any naming scheme will have to deal with conflicts, but in this example we seem to be more exposed to them than, e.g. if we call them `_MissingPropertyName0` etc.

Other options?

 * `_1` - numeric but less likely to be written by a human
 * `__1` - double-underscore is often a signal for an implementation-reserved name in programming languages/libraries

adamchester:
@nblumhardt how do you feel about the tests using `IEqualityComparer<LogEventPropertyValue>` ? What made me uneasy about it is that other tests seem to avoid that style. Secondly, it's not clear how likely it is that those structures change in the future. If they ever change (e.g. new property added), how will we know to fix these tests?

nblumhardt:
I really like the style of the tests- it'd be nice to improve our test coverage around a lot of this stuff, and the clean/clear assertions seem like a good foundation for it :+1:

adamchester:
Closing to build again

nblumhardt:
`__0`, `__1`, `__2` has my vote after some time to think (double-underscore index) ... did you reach a conclusion on the naming?

nblumhardt:
(What to do with name conflicts in this case can probably fall out from what's easiest/most efficient to implement, IMO.)

adamchester:
I like your idea of using double-underscores+index, I will push forward with that.

src/Serilog/Parameters/PropertyBinder.cs
```diff
@@ -65,31 +65,33 @@ IEnumerable<LogEventProperty> ConstructPositionalProperties(MessageTemplate temp
...
-            for (var i = 0; i < result.Length; ++i)
-            {
-                if (result[i] != null)
+                if (propertyToken == null)
                 {
-                    result[next] = result[i];
-                    ++next;
+                    // found no property for this position; generate a default name
+                    result[position] = new LogEventProperty(
+                        "__" + nextUnmatchedPosition++,
```
nblumhardt: What do you think about using `i` for this, so that for example in `("{A}", a, b)` the property generated for `b` will be named `__1`?

When the example is `("{0}", a, b)` this would make the two properties `0` and `__1`, rather than `0` and `__0` as in the code above.

It feels natural to me to consider the underscored names to be based on the index of the parameter rather than its index in the list of unmatched properties, but curious if there are advantages to the alternative.

adamchester: Agree, it's weird they're not based on the index, I'll change it

nblumhardt:
Okay... I think we're pretty much good to go with the named property case (question of unmatched property numbering applies there too, but it's somewhat cosmetic).

I've wracked my brains for a full hour regarding the O(n2) issue, here's what I came up with, keen to hear your thoughts.

First, I don't think we should sacrifice any performance binding correct templates in order to preserve data attached to invalid ones. In fact, if we could ignore `{@0}`, which is supported today but hardly reasonable, then the algorithm for binding arguments in positional templates could always just rip through left -> right naming arguments `0`, `1`, `2` without consulting the corresponding property tokens at all:

```csharp
Log.Information("{2}, {0}", a, b, c);
```

The names of `a`, `b` and `c` are obvious, and even if we use `__n` in the named template scenario, calling `b` `1` seems less surprising here, and can be generated without needing to look at the message template.

Back to the case `{@0}`, it doesn't even always make sense:

```csharp
Log.Information("{@0}, {0}, {1}", a, b);
```

I.e. should `a` be destructured, or not? Not really a reasonable question to even ask; it seems like an invalid template.

It would still be a breaking change today though if we don't support:

```csharp
Log.Information("{@0}, {1}", a, b);
```

Since we already iterate over all of the properties in the `MessageTemplate` constructor in order to cache the information about whether they're positionally or nominally bound, perhaps we check in the positional case to make sure nothing specifies non-default destructuring? This could be an internal Boolean on `MessageTemplate` like the other two helper properties, and by checking it in `ConstructPositionalProperties()` we could decide between the new, faster, purely-positional implementation that can ignore destructuring, and the slower O(n2) one, to maintain whatever the current behaviour is in the odd cases?

Thinking aloud here more than anything; it seems like the parser/binder are suffering from a bit of tech debt so there may be much better ways to approach it. Hoped that throwing an alternative out there might spark some other ideas... :-)

adamchester:
Now has updated positional numbering for argument values missing from the message template, the numbers are now based on the parameter position instead of the index of unmatched values.

src/Serilog/Parameters/PropertyBinder.cs
```diff
@@ -51,45 +51,62 @@ public IEnumerable<LogEventProperty> ConstructProperties(MessageTemplate message
...
+        IEnumerable<LogEventProperty> ConstructPositionalPropertiesWithSomeMissing(MessageTemplate template, object[] messageTemplateParameters)
         {
             var positionalProperties = template.PositionalProperties;
 
-            if (positionalProperties.Length != messageTemplateParameters.Length)
-                SelfLog.WriteLine("Positional property count does not match parameter count: {0}", template);
-
             var result = new LogEventProperty[messageTemplateParameters.Length];
-            foreach (var property in positionalProperties)
+            for (int position = 0; position < messageTemplateParameters.Length; position++)
```
nblumhardt: `var` is usually used for these in the codebase (one below as well)

src/Serilog/Parameters/PropertyBinder.cs
```diff
@@ -51,45 +51,62 @@ public IEnumerable<LogEventProperty> ConstructProperties(MessageTemplate message
...
+            for (var index = 0; index < positionalProperties.Length; index++)
             {
-                if (result[i] != null)
+                var property = positionalProperties[index];
+                int position;
+                if (!property.TryGetPositionalValue(out position))
                 {
-                    result[next] = result[i];
-                    ++next;
+                    throw new Exception("should never happen here");
```
nblumhardt: How aboue `new ArgumentException("ConstructPositionalPropertiesOptimisedAllMatching() may only be called when template and provided parameters match exactly.")` ?

test/Serilog.Tests/Core/LogEventPropertyCapturingTests.cs
```diff
@@ -0,0 +1,230 @@
...
+
+            Assert.False(sut.Equals(scalarStringB, scalarStringBStructurallyNotEqual));
+            Assert.False(sut.Equals(scalarIntA1, scalarIntB1));
+            Assert.False(sut.Equals(scalarIntA1, scalarIntA1DiffValueSameType));
+
+            Assert.False(sut.Equals(scalarGuid1, scalarGuid2));
+            Assert.False(sut.Equals(scalarGuid1, scalarGuid1StructurallyNotEqual));
+        }
+
+        class LogEventPropertyStructuralEqualityComparer : IEqualityComparer<LogEventProperty>
```
nblumhardt: Can we move these two supporting classes to `/Support`?

nblumhardt:
:+1: I think we've made it :-) .. Just a few small comments. Looking great.

test/Serilog.Tests/Core/LogEventPropertyCapturingTests.cs
adamchester: Done in 9c87109

src/Serilog/Parameters/PropertyBinder.cs
adamchester: Done in f49a38a

src/Serilog/Parameters/PropertyBinder.cs
adamchester: In 7a672ef, I did `ArgumentException("This method can only process a template with all positional properties")` - I hope that's clear enough.

adamchester:
OK seems ready ! 🎉

nblumhardt:
Great :-)

One last thing, maybe not important, but what behavior do we expect from:

```csharp
Capture("Hello {0} {0} {1}", 0, 1),
```

?

This is valid in a positional template (format-string style) but I don't think we have a test for it. As far as I can tell, we'll match it correctly but emit a spurious (parameter count) warning. Is that your expectation too?

If so, the warning is inaccurate, but consistent with pre-PR behaviour. Probably not a blocker, just keen to verify we're on the same page :-)

adamchester:
Hmm good catch! I definitely need a test for that scenario, I hadn't explored that one much.

Let me have a quick go at fixing it, even though it's consistent with pre-PR behaviour 👍

adamchester:
While investigating, I have another similar scenario that's not behaving as I expect:

```csharp
[Fact]
public void CapturingMultipleSamePositionalsStillCapturesMissing()
{
    Assert.Equal(new[]
        {
            new LogEventProperty("0", new ScalarValue(Environment.NewLine)),
            new LogEventProperty("1", new ScalarValue("Adam")),
            new LogEventProperty("__2", new ScalarValue("Missing")),
        },
        Capture("Hello {0}{1} {0}It's me", Environment.NewLine, "Adam", "Missing"),
        new LogEventPropertyStructuralEqualityComparer());
}
```
Results
```
Serilog.Tests.Core.LogEventPropertyCapturingTests.CapturingMultipleSamePositionalsStillCapturesMissing [FAIL]
      Assert.Equal() Failure
      Expected: LogEventProperty[] [LogEventProperty { Name = "0", Value = "
      " }, LogEventProperty { Name = "1", Value = "Adam" }, LogEventProperty { Name = "__2", Value = "Missing" }]
      Actual:   LogEventProperty[] [LogEventProperty { Name = "0", Value = "
      " }, LogEventProperty { Name = "1", Value = "Adam" }, null]
      Stack Trace:
        Core\LogEventPropertyCapturingTests.cs(61,0): at Serilog.Tests.Core.LogEventPropertyCapturingTests.CapturingMultipleSamePositionalsStillCapturesMissing()
```

In other words, currently we will get a `null` which is almost definitely a bug.

This exposes another problem with the fast vs slow positional matching I think. It's relying on the parameter count, which is wrong.

nblumhardt:
Nice catch, thanks for the update.

adamchester:
OK, so now I've reversed/removed the optimisation that caused the bug.

I'm not yet sure how to re-implement it without causing problems. I think we'd need to pre-calculate how many unique positional property tokens there are (e.g. {0}, {1}, {0} ==> 2 unique numbers starting from zero and sequential, therefore we can correctly expect exactly 2 parameters to be passed into the capturing method.

nblumhardt:
Thanks for the update. Do you think pushing on with this one will pay off? It seems like it's going to be hard to deal with the "invalid usage" case more gracefully without detracting from the common case. I'm fine either way, just seems like we should take stock.

adamchester:
> Thanks for the update. Do you think pushing on with this one will pay off? It seems like it's going to be hard to deal with the "invalid usage" case more gracefully without detracting from the common case. I'm fine either way, just seems like we should take stock

I'm not 100% sure what you mean here. Seems be a few options at this point:

 1. Do not try to log missing template parameters (close PR)
 1. Continue log missing template parameters (accept PR as-is)
 1. Modify the rules of positional matching and/or warnings to make the 'common case' faster at the expense of the 'invalid case' (more work to do).

Which of these are you suggesting (or something different entirely)?

nblumhardt:
Howdy - sorry, my comment above was a bit vague.

I was intending to ask if you thought 1) or 3) is the better way to go. I don't think improving the invalid case brings enough benefit for us to bring unbounded O(n2) matching into the positional case, even if its unlikely to explode in normal use. I personally can't think why I'd ever pipe dozens of positional parameters through in one template, but I am wary of underestimating how much crazy logging code is out there :-)

The 4) which might also be an option is to let the behaviour on the invalid case diverge between the named and positional matching options. We could take the improvement to the named argument case, and just ignore the positional one?

Alternatively - 4a - we could drop support for `@0` etc. from positional capturing, and always grab all the args as `0`, `1`, `2` regardless of whether they appear in the text of the template or not. This would ensure all values are recorded, run faster than the current (1.5) impl, and probably not cause a great deal of harm.

I'm leaning towards 4a despite the breakage, since I think `{@0}` in a format string is a Frankenstein's monster (and the analyzer could detect and warn on this case too).

Which direction are you leaning? Cheers!

adamchester:
OK this PR no longer follows the [SRP-for-PRs](http://en.wikipedia.org/wiki/Single_responsibility_principle).

I'm thinking of splitting it up as follows:
 - [ ] - Capturing tests that cover existing behaviour, including EqualityComparers and tests.
 - [ ] - Capturing additional parameters in _Named_ templates
 - [ ] - Capturing additional parameters in _Positional_ templates.

Does this make sense?

nblumhardt:
Sounds like a plan, great idea.

adamchester:
Closing in favour of smaller targeted PRs

PR #749, opened 2016-05-22
Corrected script path to match new structure in dotnet/cli

merbla:
To fix build issues created re: change in dotnet/cli [commit](https://github.com/dotnet/cli/commit/78a70f049a3074bceb7f4106ceab7f420c217531)


merbla:
See dotnet/cli#3165

PR #750, opened 2016-05-23
Optimize ForContext() by removing allocations

nblumhardt:
The `ForContext()` set of methods are performance and allocation-sensitive because they're on the code path even when logging is off or at a level that produces few events. The change speeds up `ForContext()` by almost 4x in common usage.

This is one of three incoming 2.0 PRs that break `ILogger`, so ideally won't be merged until all three are ready.

## Breaking change

**Adds `ForContext(ILogEventEnricher)` to `ILogger`, thus a breaking change.**

Impact should be low - will not break packages that consume `ILogger`, but will break code that implements it.

## Measurements

Most of the allocations were due to `ForContext()` requiring a list or enrichers - since most `ForContext()` calls are to add a single property, this is hugely wasteful - array allocations, iteration, etc.

The PR shifts `Logger` to call through a single enricher, and if more than one is provided a new `SafeAggregateEnricher` (internal) replicates the existing behaviour.

Benchmark is:

```csharp
    .WriteTo.Console()
    .CreateLogger();

var sw = Stopwatch.StartNew();

for (var i = 0; i < 10000000; ++i)
{
    var ctx = logger.ForContext("I", i);
    if (i == 9999999)
        ctx.Information("Event!");
}

Console.WriteLine(sw.Elapsed);
```

Best of three runs timing:

 * 4.30 sec for Serilog 2.0 RC on .NET 4.6
 * 1.08 sec for the patched version, same configuration

Existing code the calls `ForContext(p, q)` will benefit without modification.

## Other changes

* Since auditing is no longer an interesting scenario, an additional indirection through `SafeAggregateSink` has been eliminated when filters are active.
* A couple of `Logger` methods that could throw were fixed

### DIFF
--- a/src/Serilog/Core/Enrichers/EmptyEnricher.cs
+++ b/src/Serilog/Core/Enrichers/EmptyEnricher.cs
@@ -1,0 +1,25 @@
+﻿// Copyright 2013-2015 Serilog Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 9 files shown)


merbla:
LGTM, we need to revive https://github.com/serilog/serilog/blob/dev/test/Serilog.PerformanceTests/Program.cs#L9

nblumhardt:
Merging this one to keep things sane; resolving conflicts in the other.

Issue #746, opened 2016-05-20
Efficient creation of `LogEvent` with destructured properties


nblumhardt:
A longtime API rough spot: it's non-trivial to construct a `LogEvent` from its constituent parts except by calling a method on an `ILogger`.

It's not _impossible_, but there are two complications:

 - Parsing and binding parameters to a `MessageTemplate` isn't straightforward
 - Properties not in the template can only be added through `ForContext()`, which is indirect and not super-efficient

The current state of things is summed up pretty well by this method from _Serilog.Extensions.Logging_, which takes a payload from a _Microsoft.Extensions.Logging_ call and passes it through to the Serilog pipeline:

https://github.com/serilog/serilog-extensions-logging/blob/dev/src/Serilog.Extensions.Logging/Extensions/Logging/SerilogLogger.cs#L51

In there you'll observe a lot of.

```csharp
logger = logger.ForContext(property.Key, property.Value);
```

This pattern takes a name/value pair and attaches them as a property by way of enrichment. It's not correct to just call `new LogEventProperty(name, new ScalarValue(value))`, since policies may be in effect for which types are captured as scalar values. (When destructuring is required this is especially true.)

The issue of binding a `MessageTemplate` is side-stepped because the positional mapping is already done by the external logging API, but it's still unfortunate that template parsing in this class can't make use of the internal cache shared by Serilog loggers.

I have one proposal to offer but there are several ways to approach it so this thread is open for ideas :-)

nblumhardt:
### Proposal 1

```csharp
var evt = logger.CreateEvent(
    DateTimeOffset.Now, level, exception, "Some {Number}", 42);

evt.AddOrUpdateProperty(
    logger.CreateProperty("Ticks", 123));

evt.AddOrUpdateProperty(
    logger.CreateProperty("User", user, destructureObjects: true));

logger.Write(evt);
```

 - `CreateEvent()` parses the message template and binds parameters (invoking destructuring policies etc. as needed)
 - `CreateProperty()` just handles the scalar conversion/destructuring duties.

It's not an incredibly elegant interface, lots of long method names - but this kind of thing is only used in glue code, not typical application code.

#### Issues/questions

First, it's a breaking change for the `ILogger` interface. At this point very few 2.0-compatible packages will _implement_ `ILogger`, so it's the last opportunity to get this change in, I think. If we make this change we should also do #585.

[...]

PR #751, opened 2016-05-23
Add BindMessageTemplate() and BindProperty() to ILogger

nblumhardt:
## Breaking change

Adds two methods to `ILogger`. Change 2 of 3.

## Details

A modification of Proposal 1 from #746

```csharp
MessageTemplate messageTemplate;
IEnumerable<LogEventProperty> properties;
if (!logger.BindMessageTemplate("Some {Number}", new object[] { 42 }))
    return;

var evt = logger.CreateEvent(DateTimeOffset.Now, level, exception, messageTemplate, properties);

LogEventProperty ticksProperty;
if (logger.CreateProperty("Ticks", 123, false, out ticksProperty))
    evt.AddOrUpdateProperty(ticksProperty);

LogEventProperty userProperty;
if (logger.CreateProperty("User", user, true, out userProperty))
    evt.AddOrUpdateProperty(userProperty);

logger.Write(evt);
```

Not the prettiest API in the world, but:

 * Keeps things orthogonal - there's already a public constructor on `LogEvent` so there's no need to return one
 * Maps 1:1 with what `Logger` internally needs to do, so adds little/no overhead
 * The methods can fail, but `ILogger` never throws - hence the `bool` return types

Notes:

Internally, `Logger.Write()` makes allowances for calls like:

```csharp
var myInts = new[] { 1, 2, 3 };
log.Information("Numbers are {Ints}", myInts);
```

https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/Logger.cs#L174

The C# binder will consider the `int[]` as covariant with `object[]`, and so without our little hack, `Ints` ends up with the value `1`.

This behavior is _not_ propagated by `BindMessageTemplate()`, since that API is designed for programmatic scenarios we can't foresee and might break via the heuristic mentioned above.

### DIFF
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -100,6 +100,10 @@ namespace Serilog.Core
         /// <summary>
         /// Create a logger that enriches log events with the specified property.
         /// </summary>
+        /// <param name="propertyName">The name of the property. Must be non-empty.</param>
+        /// <param name="value">The property value.</param>
+        /// <param name="destructureObjects">If true, the value will be serialized as a structured
+        /// object if possible; if false, the object will be recorded as a scalar or simple array.</param>
         /// <returns>A logger that will enrich log events as specified.</returns>
         public ILogger ForContext(string propertyName, object value, bool destructureObjects = false)
         {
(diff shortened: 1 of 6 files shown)


src/Serilog/Core/Logger.cs
```diff
@@ -168,21 +172,19 @@ public bool IsEnabled(LogEventLevel level)
...
-            var now = DateTimeOffset.Now;
-
             MessageTemplate parsedTemplate;
-            IEnumerable<LogEventProperty> properties;
-            _messageTemplateProcessor.Process(messageTemplate, propertyValues, out parsedTemplate, out properties);
+            IEnumerable<LogEventProperty> boundProperties;
+            _messageTemplateProcessor.Process(messageTemplate, propertyValues, out parsedTemplate, out boundProperties);
 
-            var logEvent = new LogEvent(now, level, exception, parsedTemplate, properties);
+            var logEvent = new LogEvent(DateTimeOffset.Now, level, exception, parsedTemplate, boundProperties);
```
merbla: Should this be UTC?

merbla: Ahh I see... previously was `DateTimeOffset.Now`

https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/Logger.cs#L179

nblumhardt: Yep; didn't see any reason for keeping the single-use variable :+1:

src/Serilog/Core/Logger.cs
```diff
@@ -331,21 +335,19 @@ public void Write<T>(LogEventLevel level, Exception exception, string messageTem
         [MessageTemplateFormatMethod("messageTemplate")]
         public void Write(LogEventLevel level, Exception exception, string messageTemplate, params object[] propertyValues)
         {
-            if (messageTemplate == null) return;
             if (!IsEnabled(level)) return;
+            if (messageTemplate == null) return;
```
khellang: 😄

nblumhardt: Microoptimisation FTW ;-)

nblumhardt:
Thanks!

Issue #585, opened 2015-11-17
ILogger generic overloads


nblumhardt:
`ILogger` generic overloads and public `Logger` abstract base class with template method for implementors; from https://github.com/serilog/serilog/issues/479.

In 1.5, generic overloads were added to all `params` methods on `Log` so that allocations and boxing are avoided in the do-nothing scenario.

Making the same changes on `ILogger` is a binary-breaking change. By providing a public `Logger` ABC at the same time, we can make it easier for those implementing ILogger to provide the new methods without a huge amount of work, and make this task easier in future (note, the public class will be different from the internal pipeline's Logger, which may need a rename).

nblumhardt:
Closed by #752. Thanks @justinvp for the original push :-)

PR #752, opened 2016-05-23
Add allocation-saving generic overloads to ILogger

nblumhardt:
#585 

## Breaking change

Adds many methods to ILogger.

## Description

Removes one `params` array allocation per call, and potentially one boxing allocation per argument, for all `ILogger` methods of three or fewer parameters when logging for that level is off.

```csharp
                .WriteTo.Sink(new DelegatingSink(e => { }))
                .CreateLogger();

            var sw = System.Diagnostics.Stopwatch.StartNew();

            for (var i = 0; i < 10000000; ++i)
            {
                logger.Debug("Iteration {I}", i);
            }

            Console.WriteLine(sw.Elapsed);
```

Before: 0.153 seconds; after: 0.016 seconds.

Notes:

Ridiculously large and tedious change, needs another pair of eyes on it :-)

### DIFF
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -29,6 +29,8 @@ namespace Serilog.Core
     /// </summary>
     public sealed class Logger : ILogger, ILogEventSink, IDisposable
     {
+        static readonly object[] NoPropertyValues = new object[0];
+
         readonly MessageTemplateProcessor _messageTemplateProcessor;
         readonly ILogEventSink _sink;
         readonly Action _dispose;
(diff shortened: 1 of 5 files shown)


src/Serilog/Core/Logger.cs
```diff
@@ -29,6 +29,8 @@ namespace Serilog.Core
     /// </summary>
     public sealed class Logger : ILogger, ILogEventSink, IDisposable
     {
+        static readonly object[] NoPropertyValues = new object[0];
```
khellang: Is `Array.Empty<object>()` available?

nblumhardt: 4.6+ :-) Thought about it but didn't seem worth the `#if` noise.

merbla:
My eyes 🙈 ....

I could not see any omissions, well done.  For my benefit, the allocation improvement relates to not calling the following?  Is that correct?

https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/Logger.cs#L172

```
        public void Write(LogEventLevel level, Exception exception, string messageTemplate, params object[] propertyValues)
        {
            if (messageTemplate == null) return;
            if (!IsEnabled(level)) return;
```

khellang:
The allocation benefit comes from checking `IsEnabled(level)` _before_ calling the `params object[]` overloads. This will save the params array and boxing allocations when a logging level isn't enabled :smile:

merbla:
👍  thanks @khellang just wanted to make sure I understood correctly.

nblumhardt:
@khellang @merbla guess we're good to go on these three `ILogger`-related ones?

khellang:
Boom!

khellang:
These are definitely candidates for source generation in the future :grin:

PR #756, opened 2016-05-26
Bash script for dotnet core.

merbla:
Updated the existing `build.sh` to work with dotnet core.


nblumhardt:
LGTM.

khellang:
Why was this hard coded to specific paths? Wouldn't it be better to keep the loop as it were?

nblumhardt:
No preference personally - @merbla ?

merbla:
@khellang @nblumhardt no good reason... #757 is the rework.

PR #758, opened 2016-05-30
Enable(action) for SelfLog

nblumhardt:
The `TextWriter`-based `SelfLog.Out` is hard to use on remote machines because there's no clear 'action' on which a freshly-written self-log event can be sent across the network. This patch obsoletes `Out` and replaces it with `Enable(Action<string>)`, `Enable(TextWriter)` and `Disable()`.

```csharp
// Old code
SelfLog.Out = file;

// New code
SelfLog.Enable(file);

// or
SelfLog.Enable(m => SendToSantaClaus(m));
```

### DIFF
--- a/src/Serilog/Debugging/SelfLog.cs
+++ b/src/Serilog/Debugging/SelfLog.cs
@@ -23,7 +23,7 @@ namespace Serilog.Debugging
     /// </summary>
     public static class SelfLog
     {
-        static readonly Action<string> _output;
+        static Action<string> _output;
 
         /// <summary>
         /// The output mechanism for self-log events.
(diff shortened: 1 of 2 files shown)


test/Serilog.Tests/Debugging/SelfLogTests.cs
```diff
@@ -0,0 +1,35 @@
...
+        [ThreadStatic]
+        static List<string> Messages;
+        
+        [Fact]
+        public void MessagesAreWrittenWhenOutputIsSet()
+        {
+            Messages = new List<string>();
+            SelfLog.Enable(m =>
+            {
+                Messages = Messages ?? new List<string>();
```
merbla: There are some references to `messages` vs `events`.  Should these and the test case use `event` for clarity?

src/Serilog/Debugging/SelfLog.cs
```diff
@@ -23,14 +23,63 @@ namespace Serilog.Debugging
...
+            {
+                output.WriteLine(m);
+                output.Flush();
+            });
+        }
+
+        /// <summary>
+        /// Set the output mechanism for self-log events.
+        /// </summary>
+        /// <param name="output">An action to invoke with self-log messages.</param>
```
merbla: self-log events or self-log messages?

nblumhardt: Good catch; I've updated it to use 'messages' everywhere, since when we talk about events in Serilog they're mostly structured, and these are deliberately low-key strings :+1:

test/Serilog.Tests/Debugging/SelfLogTests.cs
nblumhardt: Updated to 'messages' in the other usage :+1:

Issue #754, opened 2016-05-24
RFC: MinimumLevel.Override(context, level)


nblumhardt:
```csharp
Log.Logger = new LoggerConfiguration()
  .MinimumLevel.Override("Microsoft", LogEventLevel.Warning)
  .MinimumLevel.Override("MyApp.Something.Tricky", LogEventLevel.Verbose)
  .WriteTo.LiterateConsole()
  .CreateLogger();
```

ASP.NET Core has some useful logging pipeline features, e.g. `RequestId` correlation, top-level error handing, etc., that make it compelling to plug Serilog in even when using Serilog's API directly (i.e. not just _Microsoft.Extensions.Logging_).

However, the framework can create a lot of log noise; Microsoft's own providers (e.g. its Console one), and soon perhaps _m.e.Logging_ itself, offer per-source level mappings to alleviate this. See, for example: https://github.com/aspnet/Logging/issues/437

The `MinimumLevel.Override()` method will allow useful information to be pulled from the framework, without a deluge of events related to ASP.NET plumbing.

### Alternatives

The current `.Filter.By...` API covers the same scenario, but it implements dynamic filtering that's geared towards filtering individual events. Since it relies on the events being created in the first place, it's not as efficient as a per-logger level setting could be. (It's also in need of a refresh, being a bit too flexible for pleasant ergonomics...)

### Details

 * Would we also provide an `Override()` method accepting a dictionary from string to level?
 * Will loggers only respect this when `ForContext()` is called with a type or string, or will we check the property name against `"SourceContext"` even when the name/value overload of `ForContext()` is called? (My preference is for the more limited version, since it'll incur less of a performance impact)
 * Can the second argument (the level) alternatively be a `LoggingLevelSwitch`? (Probably better to build the internals around LLS, since it'll eventually be desirable..)
 * Is it obvious enough in the example above that the pipeline's default level remains at `Information`?
 * Should we go further than the context name/namespace matching logic that we currently maintain for filtering, e.g. to support `"MyApp.*.Something"` as a context name?
 * `<appSettings>` syntax? `Config.json` syntax?

RehanSaeed:
Cool feature. It would be great if Serilog also provided a way to configure it in appSettings.json the same as the other log providers as in my issue https://github.com/aspnet/Logging/issues/437

nblumhardt:
@RehanSaeed it's your issue linked in the original description of this above :-)

[...]

PR #761, opened 2016-05-31
MinimumLevel.Override()

nblumhardt:
#754 

```csharp
    .MinimumLevel.Debug()
    .MinimumLevel.Override("Microsoft", LogEventLevel.Error)
    //
```

### DIFF
--- a/src/Serilog/Configuration/LoggerMinimumLevelConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerMinimumLevelConfiguration.cs
@@ -135,7 +135,7 @@ namespace Serilog.Configuration
             if (levelSwitch == null) throw new ArgumentNullException(nameof(levelSwitch));
 
             var trimmed = source.Trim();
-            if (trimmed == "")
+            if (trimmed.Length == 0)
                 throw new ArgumentException("A source name must be provided.", nameof(source));
 
             _addOverride(trimmed, levelSwitch);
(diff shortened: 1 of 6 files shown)


nblumhardt:
Benchmarking `ForContext()`:

 * Baseline (measured in https://github.com/serilog/serilog/pull/750) is 1.08 s
 * This PR with the feature unused: 1.10 s (1.094 s, but rounded up as per the earlier numbers)
 * This PR with one override applied: 1.10 s (1.095 s, as above)

So, negligible difference in performance as far as I can tell. Fluctuation between runs is ~0.5 s :-)

src/Serilog/Configuration/LoggerMinimumLevelConfiguration.cs
```diff
@@ -107,9 +117,40 @@ public LoggerConfiguration Error()
...
+        /// <param name="source">The (partial) namespace or type name to set the override for.</param>
+        /// <param name="levelSwitch">The switch controlling loggers for matching sources.</param>
+        /// <returns>Configuration object allowing method chaining.</returns>
+        public LoggerConfiguration Override(string source, LoggingLevelSwitch levelSwitch)
+        {
+            if (source == null) throw new ArgumentNullException(nameof(source));
+            if (levelSwitch == null) throw new ArgumentNullException(nameof(levelSwitch));
+
+            var trimmed = source.Trim();
+            if (trimmed == "")
```
khellang: `string.Empty`?

merbla:
:shipit:

PR #769, opened 2016-06-11
Revive Performance Tests for RC2

merbla:
This PR is an attempt to create some meaningful performance tests for RC2 and the future.  It makes use of [BenchmarkDotNet](https://github.com/PerfDotNet/BenchmarkDotNet/) however it is a rudimentary use of its features.

Initial tests start with areas such as:

* LogContext
* Min LogLevel
* Message Template Parsing
* ctor of LogEvents

ping @nblumhardt @khellang, I am a noob at BenchmarkDotNet, not sure if we want this building on the CI or if we want to store any output.

**Example output:**

```ini

BenchmarkDotNet=v0.9.7.0
OS=OSX
Processor=?, ProcessorCount=4
Frequency=1000000000 ticks, Resolution=1.0000 ns, Timer=UNKNOWN
HostCLR=CORE, Arch=64-bit RELEASE [RyuJIT]
JitModules=?
1.0.0-preview1-002702

Type=MinimumLevelTests  Mode=Throughput  Toolchain=Core  

```
         Method |    Median |    StdDev | Scaled |
--------------- |---------- |---------- |------- |
       Baseline | 9.4462 ms | 0.2243 ms |   1.00 |
 MinimumVerbose | 9.5953 ms | 0.3045 ms |   1.02 |
   MinimumDebug | 7.9003 ms | 0.1706 ms |   0.84 |
    MinimumInfo | 6.5008 ms | 0.2654 ms |   0.69 |
    MinimumWarn | 4.8345 ms | 0.1519 ms |   0.51 |
   MinimumError | 3.2263 ms | 0.0713 ms |   0.34 |
   MinimumFatal | 1.6205 ms | 0.0290 ms |   0.17 |

```ini

BenchmarkDotNet=v0.9.7.0
OS=OSX
Processor=?, ProcessorCount=4
Frequency=1000000000 ticks, Resolution=1.0000 ns, Timer=UNKNOWN
HostCLR=CORE, Arch=64-bit RELEASE [RyuJIT]
JitModules=?
1.0.0-preview1-002702

Type=FromLogContextPushPropertyTests  Mode=Throughput  Toolchain=Core  

```
               Method |     Median |    StdDev | Scaled |
--------------------- |----------- |---------- |------- |
             Baseline |  1.6128 ms | 0.0792 ms |   1.00 |
    Push1Property1000 |  2.6842 ms | 0.2732 ms |   1.66 |
   Push1Property10000 | 24.6510 ms | 1.3680 ms |  15.28 |
  Push2Properties1000 |  3.5626 ms | 0.3188 ms |   2.21 |
 Push2Properties10000 | 32.8187 ms | 0.9014 ms |  20.35 |

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -5,7 +5,13 @@ for path in src/*/project.json; do
     dotnet build ${dirname} -c Release
 done
 
-for path in test/Serilog.Tests/project.json; do
+# for path in test/Serilog.Tests/project.json; do
+    # dirname="$(dirname "${path}")"
+    # dotnet build ${dirname} -f netcoreapp1.0 -c Release
+    # dotnet test ${dirname} -f netcoreapp1.0  -c Release
+# done
+
+for path in test/Serilog.PerformanceTests/project.json; do
     dirname="$(dirname "${path}")"
     dotnet build ${dirname} -f netcoreapp1.0 -c Release
     dotnet test ${dirname} -f netcoreapp1.0  -c Release
--- a/build.sh
+++ b/build.sh
@@ -13,6 +13,5 @@ done
 
 for path in test/Serilog.PerformanceTests/project.json; do
     dirname="$(dirname "${path}")"
-    dotnet build ${dirname} -f netcoreapp1.0 -c Release
-    dotnet test ${dirname} -f netcoreapp1.0  -c Release
+    dotnet build ${dirname} -f netcoreapp1.0 -c Release 
 done
--- a/build.sh
+++ b/build.sh
@@ -5,11 +5,11 @@ for path in src/*/project.json; do
     dotnet build ${dirname} -c Release
 done
 
-# for path in test/Serilog.Tests/project.json; do
-    # dirname="$(dirname "${path}")"
-    # dotnet build ${dirname} -f netcoreapp1.0 -c Release
-    # dotnet test ${dirname} -f netcoreapp1.0  -c Release
-# done
+for path in test/Serilog.Tests/project.json; do
+    dirname="$(dirname "${path}")"
+    dotnet build ${dirname} -f netcoreapp1.0 -c Release
+    dotnet test ${dirname} -f netcoreapp1.0  -c Release
+done
 
 for path in test/Serilog.PerformanceTests/project.json; do
     dirname="$(dirname "${path}")"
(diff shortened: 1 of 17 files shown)


merbla:
Not sure in the current state this can be included on CI.
```
Build execution time has reached the maximum allowed time for your plan (60 minutes).
```

test/Serilog.PerformanceTests/FromLogContextPushPropertyTests.cs
```diff
@@ -0,0 +1,99 @@
...
+using Serilog;
+using Serilog.Context;
+using System;
+using Xunit;
+
+namespace Serilog.PerformanceTests
+{ 
+    public class FromLogContextPushPropertyTests
+    {
+        private ILogger log;
```
nblumhardt: Nit - explicit `private` :-)

nblumhardt:
Awesome! :tada:

I'm much more interested in the infrastructure than the tests themselves at this stage - just having something there means that performance-related PRs can be measured (and can include benchmarks proving their value).

I don't think we should try running the tests in CI. Building them should be snappy, but running the `dotnet test` by hand when required would cut it for me :+1:. We also don't get consistent hardware/performance from the build server, so not much value in the results even if there wasn't a timeout.

However we do it, I think we'll have to champion performance-as-a-feature especially hard for the next few months, to get perf testing properly bedded down into the project's workflow.

Great PR! :-)

build.sh
```diff
@@ -10,3 +10,9 @@ for path in test/Serilog.Tests/project.json; do
     dotnet build ${dirname} -f netcoreapp1.0 -c Release
     dotnet test ${dirname} -f netcoreapp1.0  -c Release
 done
+
+for path in test/Serilog.PerformanceTests/project.json; do
+    dirname="$(dirname "${path}")"
+    dotnet build ${dirname} -f netcoreapp1.0 -c Release
+    dotnet test ${dirname} -f netcoreapp1.0  -c Release
```
nblumhardt: Looks like this will cause the perf tests to run on Travis. Is that the intention?

merbla: For consistency we can turn it off however the travis build runs in 8 mins & 19mins ATM due to the reduced numbers of targets.

* `Elapsed time 8 min 9 sec`
* `Elapsed time 19 min 1 sec`

I will move to a new `.sh` and `.ps1` to run the perf tests in isolation.

PR #770, opened 2016-06-11
Remove template parser redundancy and nano optimizations

Suchiman:


src/Serilog/Parsing/MessageTemplateParser.cs
```diff
@@ -43,7 +43,7 @@ public MessageTemplate Parse(string messageTemplate)
 
         static IEnumerable<MessageTemplateToken> Tokenize(string messageTemplate)
         {
-            if (messageTemplate == "")
+            if (messageTemplate.Length == 0)
```
Suchiman: `.Length` is faster and we're guarded against null from the public `Parse` method

src/Serilog/Parsing/MessageTemplateParser.cs
```diff
@@ -86,9 +86,8 @@ static MessageTemplateToken ParsePropertyToken(int startAt, string messageTempla
             next = startAt + 1;
 
             var rawText = messageTemplate.Substring(first, next - first);
-            var tagContent = messageTemplate.Substring(first + 1, next - (first + 2));
-            if (tagContent.Length == 0 ||
-                !IsValidInPropertyTag(tagContent[0]))
```
Suchiman: `!IsValidInPropertyTag(tagContent[0])` performs the check on `T` in `{Test}` but all characters in `{Test}` have been already checked by
`while (startAt < messageTemplate.Length && IsValidInPropertyTag(messageTemplate[startAt]))`

src/Serilog/Parsing/MessageTemplateParser.cs
```diff
@@ -86,9 +86,8 @@ static MessageTemplateToken ParsePropertyToken(int startAt, string messageTempla
...
+            var tagContent = rawText.Substring(1, next - (first + 2));
```
Suchiman: reading from `rawText` instead of `messageTemplate` makes the intent clearer and allows to make the `startIndex` argument const to `1` (skipping the `{` in `rawText`)

src/Serilog/Parsing/MessageTemplateParser.cs
```diff
@@ -100,7 +99,7 @@ static MessageTemplateToken ParsePropertyToken(int startAt, string messageTempla
             if (TryGetDestructuringHint(propertyName[0], out destructuring))
                 propertyName = propertyName.Substring(1);
 
-            if (propertyName == "" || !IsValidInPropertyName(propertyName[0]))
+            if (propertyName.Length == 0)
```
Suchiman: again `.Length` is faster and safe to use since `propertyName` can't possibly be null and `!IsValidInPropertyName(propertyName[0])` is redundant since the following loop checks each character

nblumhardt:
Looks good to me, thanks @Suchiman.

@adamchester has done the most work in this area recently I think - anything I'm missing, Adam? :-)

adamchester:
LGTM, too.

I had extra tests from the F# version which ran against the Serilog code, and none of them could touch those deleted lines either, and I couldn't think of any possible way to hit them.

:+1:

Issue #771, opened 2016-06-12
Controlling Destructure


jezzsantos:
Given a simple type `Thing` that looks like this:
```
public class Thing
{
    public string Property1 { get; set; }
    public string Property2 { get; set; }
}
```
and calling: `logger.Write(LogEventLevel.Information, "{@Thing}", new Thing { Property1 = "aproperty1", Property2 = "aproperty2" });`

In my custom sink, in the `Emit()`, when I handle the `LogEvent` and want to get the value of the 'Thing' property, I call `LogEvent.Properties["Thing"].ToString()` 
but I get the following: `Thing { Property1: \"aproperty1\", Property2: \"aproperty2\" }`. Which is not valid JSON. In my sink, I want to populate a database tables with JSON strings.

What I really need for my custom sink, is to get each property value as a serialized JSON string, so that I can then deserialize back the property value into an instance of a `Thing` someway down the tool chain later. (I dont need the JSON serializer to describe the type)

What should I be doing, to get a property value is valid JSON when using the '@' desctructure?

nblumhardt:
Aha! Do I ever have the library for you :-)

This is on its way to either a standalone library or Serilog PR, please excuse the giant paste. You want to do:

```csharp
var formatter = new JsonValueFormatter();

foreach (var prop in logEvent.Properties)
{
    var json = new StringWriter();
    formatter.Format(property.Value, json);
    // json.ToString()...
}
```

Here's `JsonValueFormatter` and its base class:

```csharp
    /// <summary>
    /// An abstract base class for visitors that walk data in the
    /// <see cref="LogEventPropertyValue"/> format. Subclasses, by
    /// overriding appropriate methods, may search for, transform,
    /// or print the value structures being visited.
    /// </summary>
    /// <remarks>
    /// Stateless, designed to accommodate allocation-free visiting of multiple
    /// values by the same visitor instance.
    /// </remarks>
    /// <typeparam name="TState">The type of a state object passed through
    /// the visiting process.</typeparam>
    /// <typeparam name="TResult">The type of the result generated by visiting
    /// a node.</typeparam>
    public abstract class LogEventPropertyValueVisitor<TState, TResult>
    {
        /// <summary>
        /// Visit the root node type. This method delegates to
        /// a concrete Visit*Value() method appropriate for the value.
        /// </summary>
        /// <param name="state">Operation state.</param>
        /// <param name="value">The value to visit.</param>
[...]

Issue #565, opened 2015-11-04
Additional payload limiting controls


nblumhardt:
Serilog allows destructuring depth to be controlled, which prevents the accidental serialization of some massive object graphs.

The two remaining causes of unanticipated payloads are:
1. Very long strings
2. Very large arrays or dictionaries

Significant thought will need to go into the implications and design, but it seems like the v2 release is a good time to consider introducing _maximum string length_ and _maximum collection count_ controls on `LoggerDestructuringConfiguration`.

These would either default to "off", or to some very large/permissive values.

nblumhardt:
Mentioned: https://groups.google.com/forum/#!topic/serilog/aPRsprV5PMs

IanYates:
I too am interested in something like this.  I haven't had time to dig into the destructuring code but I ran into a nasty problem, only visible when I had verbose logging enabled, where I logged out an object from a PDF library we use.  I've since written manual destructuring code to handle it nicely but the several hundred MB of stuff generated wasn't great for performance :sparkler: 

I was just logging an obscure type of PDF field that didn't conform to norms but I think that field object had references back to the document, which had references to every man and his dog, etc.  The limits here - max collection size & depth - may well have helped but I'm wondering if there's a way to design this such that once we start descending down a property path we could allow at most `x` KB of JSON to be generated.  That's still hard since the root object really needs a high `x` and several levels down you want `x` to diminish.  I guess it could drop by a factor of 10 or something each time.

The idea is that we can get something meaningful in the log (and upon seeing it recoil and go write a custom destructuring policy) rather than accidentally chewing up lots of processing time creating the log event.  In my case, I don't really test the code path with verbose logging turned on but would rely on using the LoggingLevelSwitch to throw production into verbose logging.  Instead of a crash and caught exception I got a really long pause :smile_cat: followed by the same (expected) crash/exception.  Testing probably wouldn't have caught this since it was on the failure path (hard to get to for a borked PDF field) and I tend to have the logger set for null logging for most tests.

It's a bit of a strange heuristic, and I'm not sure if it's easy to retrofit into the current pipeline, but I'd be happy to look at this in mid-April or so once v2 of Serilog is out.  Thoughts?

nblumhardt:
JSON size would be difficult because that kind of serialization occurs later, but I like the idea!

Perhaps we could come up with an equivalent heuristic, like "nodes" where a node is a property, collection element, or _n_ string data characters?

[...]

Issue #255, opened 2014-10-23
LogEventProperty and the Visitor Pattern (or lack thereof)


IanMercer:
Currently there is no clean separation of concerns between the structure of a LogEvent and the process of serializing it to a string or to Json through the Render method. The class that implements a tree of property values is intimately bound to the TextWriter class.

So when I tried to create a Bson serializer that goes direct from the tree to Bson (without the current ToJson, Parse, ToBson approach) I was not able to.  Or, if I wanted to create an in memory object tree (i.e. `Dictionary<string,object>` all the way down) and then pass that to a different serializer, like, say Json.net so I could use all of its options, I couldn't do that either.

IMHO it would be better to separate these two concerns. Have LogEventPropertyValues as a pure tree of data representing the property values being logged and then have a separate Visitor class that deals with visiting the tree to convert it into a different form.

The abstract visitor class might look something like this:- 

    /// <summary>
    /// Visit each node in a tree of LogEventProperties
    /// </summary>
    public class LogEventPropertyVisitor<T>
    {
        protected virtual T Visit(LogEventPropertyValue logEvent)
        {
            if (logEvent is ScalarValue)
                return VisitScalar(logEvent as ScalarValue);
            else if (logEvent is SequenceValue)
                return VisitSequence(logEvent as SequenceValue);
            else if (logEvent is StructureValue)
                return VisitStructure(logEvent as StructureValue);
            else
                throw new Exception(logEvent.GetType().Name + " not recognized");
        }
        protected abstract T VisitScalar (ScalarValue scalarValue);
        protected abstract T VisitSequence(SequenceValue sequence);
        protected abstract T VisitStructure (StructureValue structure);
    }

and an override for converting to Json might include methods like this ...

       protected override string VisitSequence(SequenceValue sequence)
        {
            return "[" + sequence.Elements.Select(e => Visit(e)) + "]";
        }

Now all that code that deals with turning the tree into Json has been separated from the property values tree.  Now it's easy to write a different visitor that outputs to a plain formatted string, or, in my case, a BsonDocument.

The base level visitor could also handle the depth limiting code and it could also accept plug-ins like the 'null value property excluder' I proposed earlier.

Any interest in pursuing this approach?  It would be clearly be a large undertaking and would impact many projects in the solution including all of the sinks!  LMK if you want me to help make it happen.

Coupled with the ability to pass in a replacement JsonVisitor this would enable the implementation of many new features on their own (like the null value suppressor, a string-length limiter, ...).

[...]

PR #772, opened 2016-06-13
LogEventPropertyValue visitor implementations

nblumhardt:
`LogEventPropertyValue` and its subclasses and related types represent data that have been immutably "snapshotted" during logging. This communicates the developer's serialization intentions to sinks, and makes the data safe to share between threads.

While this data forms a big part of Serilog's core model, it's fiddly and difficult to work with. Traversing the tree structure formed by `LogEventPropertyValue` is especially tedious, but used commonly for:

 * Formatting (e.g. JSON, the colored/literate console sinks)
 * Searching (e.g. filters based on property values)
 * Transforming (e.g. to remove or modify logged data)

This PR aims to alleviate this by including a template implementation of the [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) in the core library. This was first suggested in #255.

Having a core visitor type means suboptimal traversal code in Serilog and its sinks can slowly be replaced. A trivial application is JSON formatting, an area where work is desperately needed. The PR includes a visitor-based `JsonValueFormatter` as a step towards cleaving some responsibilities away from `JsonFormatter`.

Implicit in this is that sinks should start to move away from customized `JsonFormatter` and towards their own top-level formatters that shell out some responsibilities to `JsonValueFormatter`. A prototype "compact" formatter doing this requires an order of magnitude less code than current extensions of `JsonFormatter`.

Also included in the PR is a rewriter type for modifying (by copying) `LogEventPropertyValue` structures, also based on the visitor. No concrete implementation is included, but an example is used in the tests that points the way towards eventual use of this in a payload-size limiting capacity.

The PR also removes an obsolete constructor from `JsonFormatter`, and shifts some low-level JSON string escaping to the newer class.

See also: #771, #565, #255, ...


src/Serilog/Formatting/Json/JsonValueFormatter.cs
```diff
@@ -0,0 +1,325 @@
...
+                var prop = structure.Properties[i];
+                WriteQuotedJsonString(prop.Name, state);
+                state.Write(':');
+                Visit(state, prop.Value);
+            }
+
+            if (structure.TypeTag != null)
+            {
+                state.Write(delim);
+                state.Write("_typeTag");
```
nblumhardt: This is the default legacy format, but now some years on and after staring at a lot of JSON.NET-encoded data, `$type` seems more natural. Thoughts?

test/Serilog.Tests/Formatting/Json/JsonFormatterTests.cs
```diff
@@ -143,44 +143,22 @@ public void ADictionaryWithScalarKeySerializesAsAnObject()
...
         {
             const string s = "\\\"\t\r\n\f";
+#pragma warning disable CS0618 // Type or member is obsolete
             var escaped = JsonFormatter.Escape(s);
+#pragma warning restore CS0618 // Type or member is obsolete
             Assert.Equal("\\\\\\\"\\t\\r\\n\\f", escaped);
         }
 
         [Fact]
-        public void StringsWithoutSpecialCharactersAreUnchanged()
```
nblumhardt: These tests moved to the `JsonValueFormatterTests` fixture.

src/Serilog/Data/LogEventPropertyValueRewriter.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// Visit a <see cref="SequenceValue"/> value.
+        /// </summary>
+        /// <param name="state">Operation state.</param>
+        /// <param name="sequence">The value to visit.</param>
+        /// <returns>The result of visiting <paramref name="sequence"/>.</returns>
+        protected override LogEventPropertyValue VisitSequenceValue(TState state, SequenceValue sequence)
+        {
+            if (sequence == null) throw new ArgumentNullException(nameof(sequence));
+
+            var contents = new LogEventPropertyValue[sequence.Elements.Count];
```
nblumhardt: If modification is unlikey, it'd be optimal to allocate this lazily and re-visit **iff** the `Visit()` call on an element returns a different reference to what was passed in.

As modification is probably the exception, rather than the rule, should we switch? Worst-case performance would be somewhat worse, but also consider:

 * If a rewriter _does_ change a lot of data, we'll probably trigger the lazy allocation on the very first element and then fall back to equivalent performance
 * The non-allocating work required to determine that allocation is necessary, is probably significantly quicker than any of the subsequent allocation/copying

(Same case for structures/dictionaries).

nblumhardt: Also, doing a truly low/zero-allocation rewriter would add significant value not otherwise within reach of most external implementations.

nblumhardt:
Updated the rewriter so that no allocations are performed when data is unchanged. (Well, unfortunately because the relatively-rare `DictionaryValue` uses `IReadOnlyDictionary<K,V>`, an iterator is allocated when one is encountered. `ScalarValue`, `SequenceValue` and `StructureValue` are truly zero-allocation in the steady state.)

nblumhardt:
I just pushed a branch `f-accept` with the alternative virtual dispatch design: https://github.com/nblumhardt/serilog/blob/f-accept/src/Serilog/Data/LogEventPropertyValueVisitor.cs

It also reimplements Render() on LogEventPropertyValue to use a DefaultFormatter visitor: https://github.com/nblumhardt/serilog/blob/f-accept/src/Serilog/Events/SequenceValue.cs

Avoiding the switch-on-type is nice, but requiring `protected internal` (like `System.Linq.ExpressionVisitor`) is awful. Expressions can justify this wart because there are so many cases to deal with; here it doesn't seem justified.

Likewise, while separating the rendering code from `LogEventPropertyValue` is nice, it requires more, more complex code, and sits at odds with existing overrides of `Render()`.

I think I prefer the version PR'd here.

nblumhardt:
(The commit with the diff between this and the `Accept()` version mentioned above: https://github.com/nblumhardt/serilog/commit/75d7934558b83298d9e44f34b50a9516fd56994d)

nblumhardt:
Added a .ctor parameter to override the default _typeTag property name, or omit the property entirely.

nblumhardt:
Example of using `JsonFormatter` to implement a JSON sink:

```csharp
public static class ClefFormatter
{
    public static void Write(TextWriter output, LogEvent logEvent, JsonValueFormatter valueFormatter)
    {
        output.Write("{\"@t\":\"");
        output.Write(logEvent.Timestamp.ToString("O"));
        output.Write("\",\"@mt\":");
        JsonValueFormatter.WriteQuotedJsonString(output, logEvent.MessageTemplate.Text);

        var tokensWithFormat = logEvent.MessageTemplate.Tokens
            .OfType<PropertyToken>()
            .Where(pt => pt.Format != null);

        // Better not to allocate an array in the 99.9% of cases where this is false
        // ReSharper disable once PossibleMultipleEnumeration
        if (tokensWithFormat.Any())
        {
            output.Write(",\"@r\":[");
            var delim = "";
            foreach (var r in tokensWithFormat)
            {
                output.Write(delim);
                delim = ",";
                var space = new StringWriter();
                r.Render(logEvent.Properties, space);
                JsonValueFormatter.WriteQuotedJsonString(output, space.ToString());
            }
            output.Write(']');
        }

        if (logEvent.Level != LogEventLevel.Information)
        {
            output.Write(",\"@l\":\"");
            output.Write(logEvent.Level);
            output.Write('\"');
        }

        if (logEvent.Exception != null)
        {
            output.Write(",\"@x\":");
            JsonValueFormatter.WriteQuotedJsonString(output, logEvent.Exception.ToString());
        }

        foreach (var property in logEvent.Properties)
        {
            output.Write(',');
            JsonValueFormatter.WriteQuotedJsonString(output, property.Key);
            output.Write(':');
            valueFormatter.Format(property.Value, output);
        }

        output.Write('}');
    }

}

class ClefConsoleSink : ILogEventSink
{
    static readonly JsonValueFormatter Formatter = new JsonValueFormatter();

    public void Emit(LogEvent logEvent)
    {
        ClefFormatter.Write(Console.Out, logEvent, Formatter);
        Console.WriteLine();
    }
}
```

PR #774, opened 2016-06-14
Switch from ,n level width specifiers to :tn

### LINKED ISSUE #765: Level width formatting tweak
When @iamkoch and I first set to figuring out fixed-width levels (`INF` and the like) we jumped around a few possible design options but landed on using the format string width specifier syntax (`,n`) like:

```
"{Level,3:u}"
```

This is inconsistent with how width specifiers usually work in format strings; rather than control the width of the value, they merely constrain it once it's already produced from the object being formatted.

Looking back I no longer think this was the right choice; .NET integer types for example use the format specifier to implement this, e.g. `:x8` is an eight-digit hexadecimal. Given the extensive special-casing we had to implement for `Level`, this now looks like the better path to take.

The format above would instead be:

```
"{Level:u3}"
```

(Leaving the width specifier for padding.)

I think this is possibly less code overall, given the `LevelPropertyValue` direction the design eventually took. We should consider switching this before 2.0 is final.

nblumhardt:
2.0 introduced support for short level names like `INF`, `inf` and so-on in output templates.

We initially used `,n` width specifiers, e.g. `"{Level,3:u}"` to generate a three-character uppercase level name. This required some deviation from the `IFormattable`-style formatting we use elsewhere, and is at odds with how `,n` padding values normally work in format strings.

This PR instead opts for a scheme matching e.g. how `Int32` hex formats specify width. The equivalent of the three-character uppercase example is `"{Level:u3}"`.

Lowercase is supported with `w`, e.g. `:w3` for three-character lowercase, and title case like `Inf` can be achieved using a `t` prefix, e.g. `:t3`.

Widths from 1 to 99 are handled, but values greater than 10 don't have any effect on the current set of level names.

While not particularly useful, this also has the advantage of making a format like `"{Level,5:w3}"` work sensibly (five-character padded, three-character lowercase level name). Staying as close as possible to .NET format string semantics is a good guideline for us to follow.

(Note, this feature only applies to output templates as supported e.g. when rendering to the console - not message templates in general.)

Closes #765.

### DIFF
--- a/src/Serilog/Formatting/Display/LogEventLevelValue.cs
+++ b/src/Serilog/Formatting/Display/LogEventLevelValue.cs
@@ -16,26 +16,44 @@ using System;
 using System.IO;
 
 using Serilog.Events;
-using Serilog.Parsing;
 
 namespace Serilog.Formatting.Display
 {
-    // allows for the specific handling of the {Level} element.
+    // Allows for the specific handling of the {Level} element.
     // can now have a fixed width applied to it, as well as casing rules.
+    // Width is set through formats like "u3" (uppercase three chars),
+    // "w1" (one lowercase char), or "t4" (title case four chars).
     class LogEventLevelValue : LogEventPropertyValue
     {
         readonly LogEventLevel _value;
 
-        static readonly string[][] _shortenedLevelMap 
-            = {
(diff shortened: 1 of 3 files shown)


merbla:
:shipit: LGTM.

PR #776, opened 2016-06-18
Improve message template construction performance

adamchester:
By using a new method, similar to `Enumerable.OfType<TResult>` but faster.

My tests show it's ~2X or ~3X faster.

### DIFF
--- a/src/Serilog/Events/MessageTemplate.cs
+++ b/src/Serilog/Events/MessageTemplate.cs
@@ -86,11 +86,11 @@ namespace Serilog.Events
         /// <summary>
         /// Similar to <see cref="Enumerable.OfType{TResult}"/>, but faster.
         /// </summary>
-        private static TResult[] GetElementsOfTypeToArray<TResult>(System.Collections.IList tokens)
+        private static TResult[] GetElementsOfTypeToArray<TResult>(object[] tokens)
             where TResult: class
         {
-            var result = new List<TResult>(tokens.Count / 2);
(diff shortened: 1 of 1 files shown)


adamchester:
Benchmark is [here](https://gist.github.com/adamchester/57fa05c082b00354595532c0e6b1c79e).

adamchester:
Note that since serilog caches message templates, this will not impact normal logging speed much.

src/Serilog/Events/MessageTemplate.cs
```diff
@@ -84,6 +84,24 @@ public MessageTemplate(string text, IEnumerable<MessageTemplateToken> tokens)
         }
 
         /// <summary>
+        /// Similar to <see cref="Enumerable.OfType{TResult}"/>, but faster.
+        /// </summary>
+        private static TResult[] GetElementsOfTypeToArray<TResult>(object[] tokens)
```
nblumhardt: Explicit `private` (just keen to keep the codebase consistent around this).

nblumhardt:
Very nice! Guessing the initial size hint is the key to the perf boost :+1: 

While it won't make much difference to normal logging perf, in other scenarios (and in apps where the template cache is exhausted) it'll be worthwhile.

Just the one nitpick, otherwise LGTM.

adamchester:
Actually, I think the key is that LINQ does many unnecessary things, like `IDisposable` + `Iterator` instances and so on. I figured `/2` was a good size default because it seems like most messages would have a property token combined with text tokens.

PR #778, opened 2016-06-18
Cover property capturing with better tests

adamchester:
The `LogEventPropertyStructuralEqualityComparer` and `LogEventPropertyValueComparer` currently only support `ScalarValue` and `SequenceValue`. If the comparer encounters a `StructureValue` or `DictionaryValue` it throws `NotImplementedException`.


adamchester:
OK, assuming we agree with these tests, I think this is ready.

nblumhardt:
Awesome :+1:

PR #779, opened 2016-06-20
Capture additional params (named) not in template

adamchester:
See #741 for details.

This addresses #737

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,3 +13,17 @@ build_script:
 test: off
 artifacts:
 - path: artifacts/Serilog.*.nupkg
+deploy:
+- provider: NuGet
+  api_key:
+    secure: nvZ/z+pMS91b3kG4DgfES5AcmwwGoBYQxr9kp4XiJHj25SAlgdIxFx++1N0lFH2x
+  skip_symbols: true
+  on:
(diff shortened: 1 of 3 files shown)


adamchester:
I think this is ready.

nblumhardt:
Back to this! Sorry about the long delay.

Just one nitpicky thought, would the performance impact (on the "correct" case) be reduced from one comparison per argument, to one comparison per call, if we used:

```csharp
var result = new LogEventProperty[messageTemplateParameters.Length];
for (var i = 0; i < matchedRun; ++i)
{
    var property = template.NamedProperties[i];
    var value = messageTemplateParameters[i];
    result[i] = ConstructProperty(property, value);	
}

for (var i = matchedRun; i < messageTemplateParameters.Length; ++i)
{
    var value = _valueConverter.CreatePropertyValue(messageTemplateParameters[i]);
    result[i] = new LogEventProperty("__" + i, value);
}
```

I find it fractionally easier to follow broken into the two loops, but that might not be universal...

Keen to merge this either way, just thought I'd put the variation out there.

adamchester:
Yep, agreed. What you proposed is a slight improvement, and it's done in f3896ba.

nblumhardt:
:tada:

Issue #730, opened 2016-05-02
Provide API for transformations to match beyond exact type


adamchester:
This issue is intended to allow discussing what API that could be provided for easier destructuring / transformations (`IDestructuringPolicy`).

The first suggestion arises from discussion in #723 and #728.

 * allow `Destructure.ByTransforming` to match derived types. This would be extremely useful for transforming reflection types (#728) because at runtime we receive internal framework implementations of `System.Type` (I.e RuntimeType) which cannot be matched on exact type and therefore can't easily be transformed in the current API.

nblumhardt:
Closed in #782.

PR #782, opened 2016-06-20
Provide `ByTransformingWhere` for more complex transformations

adamchester:
This addresses #730


nblumhardt:
:+1: 

I think the reason I didn't expose this originally was that I wanted to keep options open for caching the result of `predicate` per type, i.e. so that matching `_destructuringPolicies` in `PropertyValueConverter` could one day move from O(n) to something less steep if necessary.

Looking at the API here now though, I think it's entirely reasonable for us to call `predicate` less often than once-per-value, so nothing much to lose :-)

nblumhardt:
Bumped the dev branch's package version to 2.1.0-*, so should be able to install this off of NuGet soon...

Issue #777, opened 2016-06-18
Are tests running on TravisCI ?


adamchester:
It seems like Linux/OSX tests aren’t run during CI builds, is that a known issue?

For example: https://travis-ci.org/serilog/serilog/jobs/138635977#L1696
When I run build.sh on OSX manually, I see this output
```
Project Serilog.Tests (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
xUnit.net .NET CLI test runner (64-bit osx.10.11-x64)
  Discovering: Serilog.Tests
  Discovered:  Serilog.Tests
  Starting:    Serilog.Tests
  Finished:    Serilog.Tests
=== TEST EXECUTION SUMMARY ===
   Serilog.Tests  Total: 223, Errors: 0, Failed: 0, Skipped: 0, Time: 1.223s
Project Serilog (.NETStandard,Version=v1.3) was previously compiled. Skipping compilation.
Project TestDummies (.NETStandard,Version=v1.2) was previously compiled. Skipping compilation.
```

Output from Travis https://travis-ci.org/serilog/serilog/jobs/138635976#L432
```
Project Serilog.Tests (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
The specified framework 'Microsoft.NETCore.App', version '1.0.0-rc2-3002702' was not found.
  - Check application dependencies and target a framework version installed at:
      /Users/travis/build/serilog/serilog/.dotnetcli/shared/Microsoft.NETCore.App
  - The following versions are installed:
      1.0.0
  - Alternatively, install the framework version '1.0.0-rc2-3002702'.
Project Serilog (.NETStandard,Version=v1.3) was previously compiled. Skipping compilation.
```

merbla:
Previous builds were ok, probably need a clear exist code.

[Last success with test runs](https://travis-ci.org/serilog/serilog/builds/137131371)

https://travis-ci.org/serilog/serilog/builds

merbla:
This will be fixed up in the `staging-2.0` branch.

merbla:
Fixed in #789

PR #789, opened 2016-06-26
dotnet cli preview2 and core RTM

adamchester:
Warning this PR is set to merge into staging-2.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ os:
 
 env:
   matrix: 
+    - CLI_VERSION=1.0.0-preview2-003121
     - CLI_VERSION=Latest
 
 matrix:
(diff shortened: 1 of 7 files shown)


adamchester:
This also fixes #777

merbla:
@adamchester is preview2 the toolset version that will align with Core RTM?  Thinking we should hang out for the 27th and see what the go live tooling is.

adamchester:
I believe it is the RTM, but yeah, anything could happen in next few days.

http://netlandingpage-rtm.azurewebsites.net/net/download

adamchester:
The only risk is bumping the SDK version number again.

nblumhardt:
:+1: should move us close to, if not perfectly in line with, what we need for RTM. Thanks Adam!

PR #792, opened 2016-06-27
Use env vars for clarity. Also ensure OSX and *nix using same version

merbla:
Previous PRs based on `dev`. 

* Specify version explicitly on all builds.
* Ensure that OSX and Nix are using `1.0.0-preview2-003121` (Currently inconsistent)
* Specify branch e.g. `rel/1.0.0-preview2` as location to source scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ os:
 env:
   matrix: 
     - CLI_VERSION=1.0.0-preview2-003121 CLI_INSTALL_SCRIPT_VERSION=rel/1.0.0-preview2 CLI_INSTALL_SCRIPT_URI=https://raw.githubusercontent.com/dotnet/cli/$CLI_INSTALL_SCRIPT_VERSION/scripts/obtain/dotnet-install.sh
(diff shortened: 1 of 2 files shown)


adamchester:
:+1:

.travis.yml
```diff
@@ -23,8 +23,8 @@ os:
 
 env:
   matrix: 
-    - CLI_VERSION=1.0.0-preview2-003121
-    - CLI_VERSION=Latest
+    - CLI_VERSION=1.0.0-preview2-003121 CLI_INSTALL_SCRIPT_VERSION=rel/1.0.0-preview2 CLI_INSTALL_SCRIPT_URI=https://raw.githubusercontent.com/dotnet/cli/$CLI_INSTALL_SCRIPT_VERSION/scripts/obtain/dotnet-install.sh
+    - CLI_VERSION=1.0.0-preview2-003121 CLI_INSTALL_SCRIPT_VERSION=rel/1.0.0-preview2 CLI_INSTALL_SCRIPT_URI=https://raw.githubusercontent.com/dotnet/cli/$CLI_INSTALL_SCRIPT_VERSION/scripts/obtain/dotnet-install.sh
```
adamchester: Should this be here twice?

merbla: No just the one should be there.

When it comes time for RTM we can just add a line to try a new version, e.g. preview2 & RTM in the same build.

merbla: Arguably, the vars should be outside the matrix as we are only targeting on framework version.

PR #816, opened 2016-07-22
Fixes #815 - increasing level through overrides

### LINKED ISSUE #815: Bug in overriding while lowering logging level
Hi. 

Tried to use #754 for our "buggy parts" of code. It seems that pipeline LogLevel ignores all overrides. Full test is in attachment. Code is below.

````
    .MinimumLevel.Debug()
    .MinimumLevel.Override(nameof(SerilogTesting), LogEventLevel.Verbose)
    .WriteTo.LiterateConsole(outputTemplate: "[{Timestamp:HH:mm:ss} {Level} {SourceContext}] {Message}{NewLine}{Exception}")
    .CreateLogger();

Log.Logger.Verbose("A"); // <-- not printed, expected

Log.Logger.ForContext<Program>().Verbose("1"); // <-- not printed, NOT expected
````

[SerilogTesting.zip](https://github.com/serilog/serilog/files/377872/SerilogTesting.zip)

### PR DESCRIPTION
**Pull request by nblumhardt**
This allows child loggers to increase verbosity, by bypassing the minimum level check in `Logger`'s (explicit) implementation of `ILogEventSink.Emit()`.

The approach taken here means that if a logger is cast to `ILogEventSink` clients of the logger could effectively do the same thing. If this is deemed an issue we could add a new internal interface to connect the logger chain instead of `ILogEventSink`, but in practice since `WriteTo.Logger()` uses `ILogger.Write()` there's no scenario today where this will be observable.

### DIFF
--- a/test/Serilog.Tests/LoggerConfigurationTests.cs
+++ b/test/Serilog.Tests/LoggerConfigurationTests.cs
@@ -329,7 +329,7 @@ namespace Serilog.Tests
         }
 
         [Fact]
-        public void LevelOverridesArePropagated()
+        public void LevelOverridesUpArePropagated()
         {
             var sink = new CollectingSink();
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/Serilog.Tests/LoggerConfigurationTests.cs [hunk 1]
```diff
@@ -345,5 +345,23 @@ public void LevelOverridesArePropagated()
 
             Assert.Equal(2, sink.Events.Count);
         }
+
+        [Fact]
+        public void LevelOverridesDownArePropagated()
```
merbla: Naming? `LevelOverridesDownArePropagated` to `LowerMinimumLevelOverridesArePropagated` & `LevelOverridesUpArePropagated ` to `HigherMinimumLevelOverridesArePropagated`

merbla:
Like it. :shipit: with or without my nit pick

nblumhardt:
Done, good suggestion thanks.

Issue #820, opened 2016-07-31
Add feature to support 'Audit' style logging


heavytread:
A feature should be added to support scenarios in which the calling code needs to confirm that an event has been logged to a sink and committed to the underlying store before proceeding with some action.

For example, say we are building an application for a government agency that handles records containing personally identifiable information (PII). Whenever data is updated, or whenever a certain query type is performed (e.g.: for a record containing personal info), the system will need to write an audit log that records the ID of the user who took the action. In such a case, the system needs to know that the audit record was written before allowing the user (or, in the case of a service, the caller) to proceed with the requested action. If, for whatever reason, the audit record cannot be written, the requested action must fail. I.e.: the audit record is *required* as part of the requested operation.

There are currently two features of Serilog that together make scenarios like this difficult (or impossible) to implement:

1) Sinks can optionally derive from ```PeriodicBatchingSink```. This sink allows events to be deferred and written to the underlying data store in bulk. Because commits to the underlying data store can be deferred, there is no way for applications to ensure that audit events have been written before proceeding with an audited operation

2) Serilog eats all exceptions thrown by sinks. The ```SafeAggregateSink``` and ```SafeAggregateEncricher``` classes wrap calls to sinks and enrichers in a catch block that protects applications from exceptions thrown during logging. This prevents apps from knowing when an attempt to write an audit log record has failed.

A simple workaround would be to provide developers with the option to defeat the behaviors described above, for instance by providing a method on ```LoggerConfiguration``` that can be called before the call to ```CreateLogger()```.

jods4:
:+1:
I was looking into replacing NLog with Serilog in our projects and this was one shortcoming of Serilog.
We have a kind of "audit" log beside our normal application log and the exact requirements described here:
- Non-buffered writes;
- Throwing when logging fails.

NLog has options to support both of these.

nblumhardt:
Thanks for the input @heavytread and @jods4 - good to know this is a desirable feature for some folks. Some early exploration was done in this space but there wasn't a lot of demand so it was put on hold.

There are a few design possibilities with pros and cons; with more time to think about it, I'm back to preferring:

```csharp
var logger = new LoggerConfiguration()
    .AuditTo.File("audit.txt")
    .WriteTo.SomeOtherSink()
    .CreateLogger();
```

[...]

Issue #824, opened 2016-08-11
Re-enable the macOS target for Travis


nblumhardt:
https://travis-ci.org/serilog/serilog/jobs/151408122

> System.TypeInitializationException: The type initializer for 'Crypto' threw an exception. ---> System.TypeInitializationException: The type initializer for 'CryptoInitializer' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'System.Security.Cryptography.Native': The specified module could not be found.

It appears that libssl or some other library dependency is missing.

merbla:
Related to 

https://github.com/dotnet/cli/issues/2229

https://github.com/serilog/serilog-sinks-splunk/issues/30

nblumhardt:
@merbla any ideas around this? Shall we disable the macOS build to keep everything else moving along?

merbla:
@nblumhardt it does not seem to be moving too quick to a resolution.

I have disabled the Travis build in the repo settings.

nblumhardt:
Needs:

```diff
 os:
+   - osx
    - linux
```

nblumhardt:
Some additional info from @joshka:

https://github.com/nblumhardt/serilog/commit/d8cd24777a04dd0af9715e5a3b81aa2dde0d3131#commitcomment-18664311

merbla:
MOAR info https://github.com/caskroom/homebrew-cask/pull/23803

adamchester:
Yep I had to do this too, and worked for me 👍

merbla:
@nblumhardt I will have a look into this again.  If nothing jumps up then I suggest we close it.

adamchester:
@merbla [this](https://github.com/messagetemplates/messagetemplates-fsharp/blob/dotnet-cli-rc4/tooling/configure-dotnet-cli-osx.sh) worked [for me](https://travis-ci.org/messagetemplates/messagetemplates-fsharp/jobs/196070943#L66)

```bash
#!/bin/bash

brew config

# instructions taken from https://www.microsoft.com/net/core#macos
brew update
brew install openssl
mkdir -p /usr/local/lib
ln -s /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib /usr/local/lib/
ln -s /usr/local/opt/openssl/lib/libssl.1.0.0.dylib /usr/local/lib/
```

nblumhardt:
In #932.

PR #826, opened 2016-08-12
Audit-style logging

nblumhardt:
#820 

Usage:

```csharp
    .WriteTo.Console()
    .AuditTo.File("audit.txt")
    .CreateLogger();
```

Exceptions raised by sinks configured through `AuditTo` will propagate out to the caller.

Sinks have to opt-in by overloading their configuration extension methods. This might lead to some code duplication and would be improved if `LoggerAuditSinkConfiguration` provided a helper method like `AsLoggerSinkConfiguration()` (trivial to write), but because it's part of the DSL we probably don't want this on the API surface. Doing it through an extension method (thus hiding it unless the namespace is included) would solve this but seems too hacky to be appealing.

### DIFF
--- a/src/Serilog/Configuration/LoggerAuditSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerAuditSinkConfiguration.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2013-2015 Serilog Contributors
+﻿// Copyright 2016 Serilog Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


nblumhardt:
This is ready to go (after a review), the build failure is covered by temporarily dropping macOS from the Travis build (#824)

merbla:
LGTM, any thoughts on how this would work with configuration files?

nblumhardt:
@merbla thanks; once this is merged I'll PR `<add key="serilog:audit-to:X" ...` into the XML version, and an `"AuditTo"` key into the JSON version :+1:

PR #827, opened 2016-08-12
Attempt to fix build under macOS

nblumhardt:
#824 

Trying some suggestions from https://github.com/dotnet/cli/issues/3964


nblumhardt:
Ah, I see we'll need to re-add the web hook to test this. Maybe one for another day :-)

nblumhardt:
No joy, going the other way and dropping off the osx target so that we keep linux CI running in the interim.

Issue #852, opened 2016-09-06
Dictionary with an anonymous type key or value 


joelweiss:
When logging out a Dictionary with an Anonymous Type in the Key or the Value it fails to render the value

```
2016-09-06 16:04:04 [Information] Test dictionary with anonymous type value "Capturing the property value threw an exception: RuntimeBinderException"
2016-09-06 16:04:04 [Information] Test dictionary with anonymous type key "Capturing the property value threw an exception: RuntimeBinderException"
```

Here is the code to reproduce.

```csharp
void Main()
{
	var logger = new Serilog.LoggerConfiguration().WriteTo.ColoredConsole().CreateLogger();
	var anonymous = new {Test = 4M};

	//Fine
	logger.Information("Test anonymous type {value}", anonymous);

	//Fine
	var enumerable = Enumerable.Repeat("Test", 1).Select(v => anonymous);
	logger.Information("Test enumerable with anonymous type {enumerable}", enumerable);

	//Fine
	var list = Enumerable.Repeat("Test", 1).Select(v => anonymous).ToList();
	logger.Information("Test list with anonymous type {list}", list);

	//Fail
	var dictionary = Enumerable.Repeat("Test", 1).ToDictionary(v => v, v =>  anonymous);
	logger.Information("Test dictionary with anonymous type value {dictionary}", dictionary);

	//Fail
	var dictionary2 = Enumerable.Repeat("Test", 1).ToDictionary(v => anonymous, v => v);
	logger.Information("Test dictionary with anonymous type key {dictionary}", dictionary);
}
```

nblumhardt:
Interesting find - thanks Joel!

joelweiss:
Thanks

PR #858, opened 2016-09-09
Fix for issue #852

goenning:
This commit should fix issue #852.

PS.: I've done this on a Mac, so I hope that the CI will test others OS...

### DIFF
--- a/src/Serilog/Parameters/PropertyValueConverter.cs
+++ b/src/Serilog/Parameters/PropertyValueConverter.cs
@@ -155,28 +155,21 @@ namespace Serilog.Parameters
                 // Only actual dictionaries are supported, as arbitrary types
                 // can implement multiple IDictionary interfaces and thus introduce
                 // multiple different interpretations.
-                var valueAsList = enumerable.Cast<object>();
                 if (IsValueTypeDictionary(valueType))
                 {
-                    var values = new List<KeyValuePair<ScalarValue, LogEventPropertyValue>>();
-                    if (valueAsList.Any())
-                    {
-                        var typeInfo = valueAsList.ElementAt(0).GetType().GetTypeInfo();
-                        var keyProperty = typeInfo.GetDeclaredProperty("Key");
-                        var valueProperty = typeInfo.GetDeclaredProperty("Value");
+                    var typeInfo = typeof(KeyValuePair<,>).MakeGenericType(valueType.GenericTypeArguments).GetTypeInfo();
+                    var keyProperty = typeInfo.GetDeclaredProperty("Key");
+                    var valueProperty = typeInfo.GetDeclaredProperty("Value");
 
-                        foreach (object kvp in enumerable)
-                        {
-                            var scalar = (ScalarValue)limiter.CreatePropertyValue(keyProperty.GetValue(kvp), destructuring);
-                            var propValue = limiter.CreatePropertyValue(valueProperty.GetValue(kvp), destructuring);
-                            values.Add(new KeyValuePair<ScalarValue, LogEventPropertyValue>(scalar, propValue));
-                        }
-                    }
-                    return new DictionaryValue(values);
+                    return new DictionaryValue(enumerable.Cast<object>()
+                        .Select(kvp => new KeyValuePair<ScalarValue, LogEventPropertyValue>(
+                                           (ScalarValue)limiter.CreatePropertyValue(keyProperty.GetValue(kvp), destructuring),
+                                           limiter.CreatePropertyValue(valueProperty.GetValue(kvp), destructuring)))
+                        .Where(kvp => kvp.Key.Value != null));
                 }
 
                 return new SequenceValue(
-                    valueAsList.Select(o => limiter.CreatePropertyValue(o, destructuring)));
+                    enumerable.Cast<object>().Select(o => limiter.CreatePropertyValue(o, destructuring)));
             }
 
             if (destructuring == Destructuring.Destructure)
--- a/src/Serilog/Parameters/PropertyValueConverter.cs
+++ b/src/Serilog/Parameters/PropertyValueConverter.cs
@@ -23,7 +23,6 @@ using Serilog.Events;
 using Serilog.Parsing;
 using Serilog.Policies;
 using System.Runtime.CompilerServices;
-using System.ComponentModel;
 
 namespace Serilog.Parameters
 {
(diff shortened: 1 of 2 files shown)


nblumhardt:
Thanks!

I think this could regress performance in the currently-working case because of the extra `List<>` allocation, and the runtime lookup of the key and value properties from the `TypeInfo`.

If we can figure out a way to do it using the same code structure that'd be great, but I gather from your fix that the problem is related to how `dynamic` and the key type interact?

Cheers!

goenning:
Yes, it's because anonymous types are internal and thus not acessible from serilog. Using `InternalsVisibleToAttribute` should solve the problem, but I don't like at all. 

Myabe if we had two differente branches, the original code for most cases and use reflection only when needed. Anyway, I'm open for suggestions.

src/Serilog/Parameters/PropertyValueConverter.cs
```diff
@@ -154,17 +155,28 @@ LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructur
                 // Only actual dictionaries are supported, as arbitrary types
                 // can implement multiple IDictionary interfaces and thus introduce
                 // multiple different interpretations.
+                var valueAsList = enumerable.Cast<object>();
```
nblumhardt: Since 99% of the time the value won't be a dictionary, we should avoid the extra allocation here and just do all of the dictionary handling within the block below.

src/Serilog/Parameters/PropertyValueConverter.cs
```diff
@@ -154,17 +155,28 @@ LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructur
...
                 if (IsValueTypeDictionary(valueType))
                 {
-                    return new DictionaryValue(enumerable.Cast<dynamic>()
-                        .Select(kvp => new KeyValuePair<ScalarValue, LogEventPropertyValue>(
-                                           (ScalarValue)limiter.CreatePropertyValue(kvp.Key, destructuring),
-                                           limiter.CreatePropertyValue(kvp.Value, destructuring)))
-                        .Where(kvp => kvp.Key.Value != null));
+                    var values = new List<KeyValuePair<ScalarValue, LogEventPropertyValue>>();
```
nblumhardt: The list allocation can be avoided by using the same pattern as above, only within the `Select()` block the reflection-based accessors can be used.

src/Serilog/Parameters/PropertyValueConverter.cs
```diff
@@ -154,17 +155,28 @@ LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructur
...
+                    if (valueAsList.Any())
+                    {
+                        var typeInfo = valueAsList.ElementAt(0).GetType().GetTypeInfo();
```
nblumhardt: Instead of `Any`/`ElementAt`, could we put the `keyProperty` and `valueProperty` members outside the `Select` block and lazily initialize them when the first item is processed?

(I'm guessing you've already considered it, but just in case you can think of a nicer way to do it than I, the value type will always be a predictable one: `typeof(KeyValuePair<,>).MakeGenericType(valueType.GetGenericArguments())`. Not sure about the perf of this approach, but would save having to examine any elements.)

nblumhardt:
Thanks for the clarification, makes sense! :+1: 

Added a few notes, keen to get your thoughts. Cheers!

src/Serilog/Parameters/PropertyValueConverter.cs
goenning: I'm curious about this, why would this instruction allocate something? I thought that the `Cast` method would just create a LINQ query that would be later used with the other `Select` query. Is this case `Cast` is not copying the content. Isn't it?

Ref. http://stackoverflow.com/questions/4304961/linq-memory-question

goenning:
Thanks for the code review, I didn't know about `MakeGenericType`+`GenericTypeArguments`. There is always something new to learn 👍. What do you think about it now?

src/Serilog/Parameters/PropertyValueConverter.cs
khellang: All LINQ operators can potentially allocate an `IEnumerable<T>`, even without it ever being enumerated. This is why performance-critical code like Roslyn etc. doesn't allow use of LINQ :smile:

khellang: All LINQ operators can potentially allocate an `IEnumerable<T>`, even without it ever being enumerated. This is why performance-critical code like Roslyn etc. doesn't allow use of LINQ :smile:

khellang: All LINQ operators can potentially allocate an `IEnumerable<T>`, even without it ever being enumerated. This is why performance-critical code like Roslyn etc. doesn't allow use of LINQ :smile:

khellang: All LINQ operators can potentially allocate an `IEnumerable<T>`, even without it ever being enumerated. This is why performance-critical code like Roslyn etc. doesn't allow use of LINQ :smile:

test/Serilog.Tests/Core/MessageTemplateTests.cs
```diff
@@ -163,5 +163,40 @@ public void MissingPositionalParametersRenderAsTextLikeStandardFormats()
...
+            Assert.Equal("Dictionary with anonymous type value [(\"MyKey\": \"{ Test = 5 }\")]", m);
+        }
+
+        [Fact]
+        public void DictionaryOfAnonymousTypeAsKeyShouldBeRendered()
+        {
+            var anonymous = new { Bar = 6M, Baz = 4M };
+            var dictionary = Enumerable.Repeat("MyValue", 1).ToDictionary(v => anonymous, v => v);
+            var m = Render("Dictionary with anonymous type value {dictionary}", dictionary);
+            Assert.Equal("Dictionary with anonymous type value [\"[{ Bar = 6, Baz = 4 }, MyValue]\"]", m);
```
nblumhardt: Nitpick, text should be `"Dictionary with anonymous type key`.

nblumhardt:
@khellang  :-)

@goenning looks good! Couple of minor nitpicks, otherwise :shipit:

nblumhardt:
Awesome, thanks @goenning! :-)

PR #865, opened 2016-09-23
1.0.0-preview2-003131

adamchester:


adamchester:
I should point out, I'm not even sure if there's a good reason to do this. If the previous version works and none of our deps changed, it might be better to avoid this.

nblumhardt:
Doesn't seem like it would hurt, but since "it ain't broke" perhaps we leave well enough alone until the next major churn?

adamchester:
Let's close it, no need to leave it laying around, initially I did it as an experiment anyway. Code is here if we need it.

PR #866, opened 2016-09-24
Payload limiting 

krajek:
Regarding #565.
I have done just string length limiting. When this will be completed I will move on to collections size limit as discussed in #565.
Things that need to be discussed:
a) default limit - for now I set it to 1000, but I am absolutely confortable to setting no limit by default(or int.MaxValue)
b) Whether we want to allow configuration of "truncation string". For now I used … (U+2026) without configuration. 

Ideas are welcomed.

### DIFF
--- a/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
@@ -164,7 +164,7 @@ namespace Serilog.Configuration
         /// <exception cref="ArgumentOutOfRangeException">When passed length is less or equal to 2</exception>
         public LoggerConfiguration ToMaximumStringLength(int maximumStringLength)
         {
-            if (maximumStringLength < 2) throw new ArgumentOutOfRangeException(nameof(maximumStringLength), maximumStringLength, "Maximum string length have to be greater than 1");
+            if (maximumStringLength < 2) throw new ArgumentOutOfRangeException(nameof(maximumStringLength), maximumStringLength, "Maximum string length must be at least two.");
             _setMaximumStringLength(maximumStringLength);
             return _loggerConfiguration;
         }
--- a/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
@@ -27,20 +27,24 @@ namespace Serilog.Configuration
         readonly Action<Type> _addScalar;
         readonly Action<IDestructuringPolicy> _addPolicy;
         readonly Action<int> _setMaximumDepth;
+        readonly Action<int> _setMaximumStringLength;
 
         internal LoggerDestructuringConfiguration(
             LoggerConfiguration loggerConfiguration,
             Action<Type> addScalar,
             Action<IDestructuringPolicy> addPolicy,
-            Action<int> setMaximumDepth)
+            Action<int> setMaximumDepth,
+            Action<int> setMaximumStringLength)
         {
             if (loggerConfiguration == null) throw new ArgumentNullException(nameof(loggerConfiguration));
             if (addScalar == null) throw new ArgumentNullException(nameof(addScalar));
             if (addPolicy == null) throw new ArgumentNullException(nameof(addPolicy));
+            if (setMaximumDepth == null) throw new ArgumentNullException(nameof(setMaximumStringLength));
             _loggerConfiguration = loggerConfiguration;
             _addScalar = addScalar;
             _addPolicy = addPolicy;
             _setMaximumDepth = setMaximumDepth;
+            _setMaximumStringLength = setMaximumStringLength;
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/Serilog/LoggerConfiguration.cs
```diff
@@ -39,6 +39,7 @@ public class LoggerConfiguration
         LogEventLevel _minimumLevel = LogEventLevel.Information;
         LoggingLevelSwitch _levelSwitch;
         int _maximumDestructuringDepth = 10;
+        int _maximumStringLength = 1000;
```
nblumhardt: So that this doesn't need to be flagged as a breaking change, perhaps `int.MaxValue` is a better option here?

src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
```diff
@@ -149,6 +153,21 @@ public LoggerConfiguration ToMaximumDepth(int maximumDestructuringDepth)
...
+        /// When destructuring objects, string values can be restricted to specified length
+        /// thus avoiding bloating payload. Limit is applied to each value separately, 
+        /// sum of length of strings can exceed limit.
+        /// </summary>
+        /// <param name="maximumStringLength">The maximum string length.</param>
+        /// <returns>Configuration object allowing method chaining.</returns>
+        /// <exception cref="ArgumentOutOfRangeException">When passed length is less or equal to 2</exception>
+        public LoggerConfiguration ToMaximumStringLength(int maximumStringLength)
+        {
+            if (maximumStringLength < 2) throw new ArgumentOutOfRangeException(nameof(maximumStringLength), maximumStringLength, "Maximum string length have to be greater than 1");
```
nblumhardt: Exception messages in Serilog are all sentences, so needs a `.` at the end of the message. Perhaps a slight rewording `"Maximum string length must be at least two."` ?

nblumhardt:
Thanks for picking this one up! Sorry about the slow response, just returned from a few days off.

One thing I think needs to be accounted for is:

```csharp
Log.Information("The string is {S}", s);
```

Here, `s` should probably not be truncated - it's not being captured via the destructuring process that this feature aims to control.

krajek:
Thanks for the review.
Initial value int.MaxValue is obviously the right choice here, done.
Rewording of exception also makes sense, done.
Regarding strings in Default destructuring mode (`string s; SomeObject o`):
- `Log.Information("The string is {S}", s);` will NOT be limited
- `Log.Information("The string is {@S}", s);` will be limited
- `Log.Information("The string is {$S}", s);` will be limited
- `Log.Information("The object is {O}", o);` will NOT be limited
- `Log.Information("The object is {$O}", o);` will be limited
- `Log.Information("The object is {@O}", o);` will be limited

krajek:
Shall I continue work on limiting collection size on this branch? Or maybe we can merge this one and open next PR for collections?

src/Serilog/Parameters/PropertyValueConverter.cs
```diff
@@ -121,11 +125,22 @@ LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructur
...
             if (destructuring == Destructuring.Stringify)
-                return new ScalarValue(value.ToString());
+            {
+                return Stringify(value);
+            }
 
             var valueType = value.GetType();
             var limiter = new DepthLimiter(depth, _maximumDestructuringDepth, this);
 
+            if (destructuring == Destructuring.Destructure)
```
nblumhardt: One last thought - an interesting implementation option would be to actually create an `IScalarConversionPolicy` for strings specifically, and (if the limit has been set), insert the policy at the head of the collection of policies. Making insertion of the policy conditional would mean the feature was truly pay-for-play, i.e. if no maximum depth was set, no policy would appear in the list and so runtime cost would be zero (no need for this `if` block, type comparison, etc.).

The "stringify" case might not fall out nicely from this option, but we could definitely reconsider having it apply there.

A small added bonus would be that using `int?` for the depth (instead of the `MaxValue` sentinel) would be possible without having to do a `.Value` runtime check.

What do you think?

krajek: I tried it and biggest drawback is that if we will use `IScalarConversionPolicy` as-is we are back to limiting strings in default mode Log.Information("The string is {S}", s); 
In my opinion this is good thing, strings would be limited in every case, no special cases.

Also the stringified mode is still going to be handled separately, but limiting logic was extracted to separate class, so no dupication there.

To sum up, I would go for this option. Are you sure you want to keep the requirement of `Log.Information("The string is {S}", s); will NOT be limited`?

https://github.com/krajek/serilog/commit/890f9be6cb54af0823fab68efc368a7ae27cce49
I have not changed limit to `int?`, it will be trivial.

nblumhardt: Ah, I see - thanks, yes you're right. We should continue not to limit regular string logging.

Let's merge this one; I'm looking forward to moving onto collection size and so-on.

src/Serilog/Parameters/PropertyValueConverter.cs
```diff
@@ -186,6 +201,23 @@ LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructur
             return new ScalarValue(value.ToString());
         }
 
+        private LogEventPropertyValue Stringify(object value)
```
nblumhardt: Nitpick (can pick this up later): the codebase doesn't use any explicit `private`/`internal` modifiers.

krajek: Sure, will be fixed.

src/Serilog/Parameters/PropertyValueConverter.cs
krajek: Thanks, I will move on to collection limit and come back with next PR.

PR #877, opened 2016-10-10
Payload limiting - collection limit

krajek:
As proposed in #565.

Things to be discussed:
1. Whether or not limit should be effective for dictionary. I think that yes it should, but I'm absolutely open for discussion.
2. Whether or not last element of collection should be replaced with … . I am 50/50 on that, for now I have not implemented that.
3. Whether or not limit should be effective for byte array. I think not, because it has its own policy of limiting output.

Looking forward for feedback.


src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
```diff
@@ -161,13 +166,28 @@ public LoggerConfiguration ToMaximumDepth(int maximumDestructuringDepth)
...
+
+        /// <summary>
+        /// When destructuring objects, collections be restricted to specified length
+        /// thus avoiding bloating payload. Limit is applied to each collection separately, 
+        /// sum of length of collection can exceed limit.
+        /// Applies limit to all <see cref="IEnumerable"/> including dictionaries.
+        /// </summary>
+        /// <returns>Configuration object allowing method chaining.</returns>
+        /// <exception cref="ArgumentOutOfRangeException">When passed length is less than 1</exception>
+        public LoggerConfiguration ToMaximumCollectionLength(int maximumCollectionLength)
```
nblumhardt: This should possibly be `ToMaximumCollectionSize` or `ToMaximumCollectionCount` since only arrays expose it as `Length` in .NET.

nblumhardt:
One minor comment on naming, otherwise LGTM 👍

krajek:
Thanks for the review. I changed 'length' to 'count', since I though 'size' could be understood as size in bytes.

src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
```diff
@@ -28,28 +28,28 @@ public class LoggerDestructuringConfiguration
         readonly Action<Type> _addScalar;
         readonly Action<IDestructuringPolicy> _addPolicy;
         readonly Action<int> _setMaximumDepth;
-        readonly Action<int> _setMaximumStringLength;
-        readonly Action<int> _setMaximumCollectionLength;
+        readonly Action<int> _setMaximumStringCount;
```
nblumhardt: Looks like a find/replace error here - should still be `_setMaximumStringLength`.

nblumhardt:
Great, agreed on the reasoning for `Count`. Just spotted one minor typo.

krajek:
Thanks again. (Side note: coding around kids playing around and screaming is not the best idea :-))
BTW, I found some incorrect nameofs is constructor checks.

nblumhardt:
>  coding around kids playing around and screaming is not the best idea :-)

I know _exactly_ what you mean :-)

PR #883, opened 2016-10-19
Improve MessageTemplateCache performance

### LINKED ISSUE #823: Improve MessageTemplateCache performance
Next thing which shows up in our multi-threaded logging scenarios is contention in MessageTemplateCache. The problem is that on every Parse (which is always on hot path of Logger.Write) lock is taken even if parsed message template is in the cache. I suggest switch to ConcurrentDictionary and can submit a PR.

Also for the interned strings this optimization makes no sense, imo, as they're already in memory:
```
            if (messageTemplate.Length > MaxCachedTemplateLength)
                return _innerParser.Parse(messageTemplate);
```
So converting to
```
            if (messageTemplate.Length > MaxCachedTemplateLength && string.IsInterned(messageTemplate) == null)
                return _innerParser.Parse(messageTemplate);
```
may improve performance to those who use large message templates.

P.S. Why DateTimeOffset.Now is used as timestamp of LogEvent? Why not UtcNow? What if destination sink (like Seq) is located in different timezone, collecting log events from two or more different locations? As I understand now events happened in different locations but in the same time won't show in Seq near each other. But I may be wrong, which happens often :)

skomis-mm:
- fixes #823 for all except  `netstandard10-12`
- added perf tests for normal (cached) and leaking scenarios: [netcoreapp10](https://github.com/skomis-mm/serilog/tree/msgtplcache/results/netcoreapp1.0/MessageTemplateCache), [net452](https://github.com/skomis-mm/serilog/tree/msgtplcache/results/net4.5.2/MessageTemplateCache)

based on old forgotten **Hashtable**. According to [docs](https://msdn.microsoft.com/en-us/library/system.collections.hashtable(v=vs.110).aspx#Anchor_10) it is thread-safe for multiple readers (no enumerations) *and* single writer. So looked like a good fit for this specific scenario with tradeoff on locking only writers in exchange on a cheap `Count` property.


nblumhardt:
That's fascinating, I completely overlooked `Hashtable`. Looks like great results!

dotnetjunkie:
Holy crap. That [HashTable](https://referencesource.microsoft.com/#mscorlib/system/collections/hashtable.cs,10fefb6e0ae510dd) got some really complex lock free code using spinwaits and `Thread.Sleep`. Must be written by @joeduffy.

dotnetjunkie:
Warning, @davkean just warned [here](https://github.com/JamesNK/Newtonsoft.Json/issues/1163#issuecomment-271852310) about HashTable ***not*** being thread-safe, not even with hread-safe for multiple readers and a single writer.

nblumhardt:
Thanks @dotnetjunkie

Issue #885, opened 2016-10-28
NaN and PositiveInfinity/NegativeInfinity formatted as invalid JSON


nblumhardt:
The "special" floating point values need to be converted into strings (or some other valid JSON representation) when formatted by `JsonValueFormatter`. See: https://github.com/serilog/serilog/blob/dev/src/Serilog/Formatting/Json/JsonValueFormatter.cs#L185

clarkis117:
@nblumhardt  From my understanding of the issue they need to be formatted as valid JSON, which in this case could just be strings. Therefore, a conditional statement like this should work:

	if ((value is double && (double.IsNaN((double)value) || double.IsInfinity((double)value)))
		|| (value is float && (float.IsNaN((float)value) || float.IsInfinity((float)value))))
	{
		FormatStringValue(value.ToString(), output);
		return;
	}

clarkis117:
PR: https://github.com/serilog/serilog/pull/887

PR #887, opened 2016-10-28
fix for Issue #885, formatting floating point special values NaN/PositiveInfinity/NegativeInfinity as JSON strings and unit tests

clarkis117:
This fixes issue #885 where special floating point values are not properly formatted as JSON.

 It adds an if statement to the JSON value formatter that catches the special values and formats them as JSON strings. In addition, it adds several unit tests that check if the special values are being formatted as JSON strings.

### DIFF
--- a/src/Serilog/Formatting/Json/JsonValueFormatter.cs
+++ b/src/Serilog/Formatting/Json/JsonValueFormatter.cs
@@ -231,7 +231,7 @@ namespace Serilog.Formatting.Json
 		{
 			if (float.IsNaN(value) || float.IsInfinity(value))
 			{
-				FormatStringValue(value.ToString(), output);
+				FormatStringValue(value.ToString(CultureInfo.InvariantCulture), output);
 				return;
 			}
 
@@ -242,7 +242,7 @@ namespace Serilog.Formatting.Json
 		{
 			if (double.IsNaN(value) || double.IsInfinity(value))
 			{
-				FormatStringValue(value.ToString(), output);
+				FormatStringValue(value.ToString(CultureInfo.InvariantCulture), output);
 				return;
 			}
 
--- a/src/Serilog/Formatting/Json/JsonValueFormatter.cs
+++ b/src/Serilog/Formatting/Json/JsonValueFormatter.cs
@@ -182,18 +182,17 @@ namespace Serilog.Formatting.Json
                     return;
                 }
 
-                if (value is double || value is float)
-                {
-					if ((value is double && (double.IsNaN((double)value) || double.IsInfinity((double)value)))
-						|| (value is float && (float.IsNaN((float)value) || float.IsInfinity((float)value))))
-					{
-						FormatStringValue(value.ToString(), output);
-						return;
-					}
+				if (value is double)
+				{
+					FormatDoubleValue((double)value, output);
+					return;
+				}
 
-					FormatApproximateNumericValue((IFormattable)value, output);
-                    return;
-                }
+				if (value is float)
+				{
+					FormatFloatValue((float)value, output);
+					return;
+				}
 
                 if (value is bool)
                 {
@@ -228,10 +227,27 @@ namespace Serilog.Formatting.Json
             output.Write(value ? "true" : "false");
         }
 
-        static void FormatApproximateNumericValue(IFormattable value, TextWriter output)
-        {
-            output.Write(value.ToString("R", CultureInfo.InvariantCulture));
-        }
+		static void FormatFloatValue(float value, TextWriter output)
+		{
+			if (float.IsNaN(value) || float.IsInfinity(value))
+			{
+				FormatStringValue(value.ToString(), output);
+				return;
+			}
+
+			output.Write(value.ToString("R", CultureInfo.InvariantCulture));
+		}
+
+		static void FormatDoubleValue(double value, TextWriter output)
+		{
+			if (double.IsNaN(value) || double.IsInfinity(value))
+			{
+				FormatStringValue(value.ToString(), output);
+				return;
+			}
+
+			output.Write(value.ToString("R", CultureInfo.InvariantCulture));
+		}
 
         static void FormatExactNumericValue(IFormattable value, TextWriter output)
         {
--- a/src/Serilog/Formatting/Json/JsonValueFormatter.cs
+++ b/src/Serilog/Formatting/Json/JsonValueFormatter.cs
@@ -184,7 +184,14 @@ namespace Serilog.Formatting.Json
 
                 if (value is double || value is float)
                 {
-                    FormatApproximateNumericValue((IFormattable)value, output);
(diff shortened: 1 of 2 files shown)


nblumhardt:
Thanks for the PR, looks great!

Just wondering, could we keep a bit of the noise out of the switch-on-type part of the code by changing:

```csharp
         static void FormatApproximateNumericValue(IFormattable value, TextWriter output)
         {
             output.Write(value.ToString("R", CultureInfo.InvariantCulture));
         }
```

...into two procedures, one for doubles and one for floats, like:

```csharp
         static void FormatDoubleValue(double value, TextWriter output)
         {
             if (double.IsNan(value) || double.IsInfinity(value)) 
             {
                 // etc.
             }

             output.Write(value.ToString("R", CultureInfo.InvariantCulture));
         }
```

(Because there's now so much type-specific code it seems like the value of the `IFormattable`-based method is now less.)

What do you think?

Cheers,
Nick

clarkis117:
If there are no qualms with casting a float to a double, the `FormatApproximateNumericValue` method could be kept as a single procedure.

```
       static void FormatApproximateNumericValue(IFormattable value, TextWriter output)
         {
             if (double.IsNaN(value) || double.IsInfinity(value)) 
             {
                 // etc.
             }
             output.Write(value.ToString("R", CultureInfo.InvariantCulture));
         }
```

nblumhardt:
Interesting thought. Although I don't believe it would necessarily lose information, issues like http://stackoverflow.com/questions/17504833/why-converting-from-float-to-double-changes-the-value could cause some confusion that I think we'd be better off avoiding.

clarkis117:
Oh, I mean only for the special cases of NaN, Positive Infinity, and Negative Infinity. Another observation, it looks the like the use of casting value types to Add to dictionary would cause boxing, which would degrade performance and cause more GC collections.

clarkis117:
I do see your point on the floating point values, as casting to larger float point values can cause issues with binary floating point.

nblumhardt:
Now that you mention it, the current casting to `IFormattable` will be causing boxing we should be avoiding, too 👍 

I'm thinking the most straightforward option of two separate methods is going to be the cleanest, even if there are a few structurally-similar lines of code in there.

clarkis117:
I'll add another commit to this PR for the proposed change (separating `FormatApproximateNumericValue` ). Then, I'll open another PR for eliminating the casting of Value Types to Interfaces in `JsonValueFormatter`. Does that sound good?

nblumhardt:
Actually, my bad - there won't be any additional boxing in the code as it stands today, since `value` is already boxed.

A changeset separating the method into two would be great, thanks.

src/Serilog/Formatting/Json/JsonValueFormatter.cs
```diff
@@ -221,10 +227,27 @@ static void FormatBooleanValue(bool value, TextWriter output)
...
 
-        static void FormatApproximateNumericValue(IFormattable value, TextWriter output)
-        {
-            output.Write(value.ToString("R", CultureInfo.InvariantCulture));
-        }
+		static void FormatFloatValue(float value, TextWriter output)
+		{
+			if (float.IsNaN(value) || float.IsInfinity(value))
+			{
+				FormatStringValue(value.ToString(), output);
```
nblumhardt: Just checked out how double behaves in different cultures; I think for this to work conveniently we should use `ToString(CultureInfo.InvariantCulture)` here.

src/Serilog/Formatting/Json/JsonValueFormatter.cs
```diff
@@ -221,10 +227,27 @@ static void FormatBooleanValue(bool value, TextWriter output)
...
+			}
+
+			output.Write(value.ToString("R", CultureInfo.InvariantCulture));
+		}
+
+		static void FormatDoubleValue(double value, TextWriter output)
+		{
+			if (double.IsNaN(value) || double.IsInfinity(value))
+			{
+				FormatStringValue(value.ToString(), output);
```
nblumhardt: `ToString(CultureInfo.InvariantCulture)` also.

test/Serilog.Tests/Formatting/Json/JsonValueFormatterTests.cs
```diff
@@ -49,7 +49,33 @@ public void DoubleFormatsAsNumber()
...
         }
 
-        [Fact]
+		[Fact]
+		public void DoubleSpecialsFormatAsString()
+		{
+			string format = "\"{0}\"";
+
+			JsonLiteralTypesAreFormatted(double.NaN, string.Format(format, double.NaN));
+			JsonLiteralTypesAreFormatted(double.PositiveInfinity, string.Format(format, double.PositiveInfinity));
```
nblumhardt: `"Infinity"`

test/Serilog.Tests/Formatting/Json/JsonValueFormatterTests.cs
```diff
@@ -49,7 +49,33 @@ public void DoubleFormatsAsNumber()
...
+			JsonLiteralTypesAreFormatted(double.NegativeInfinity, string.Format(format, double.NegativeInfinity));
```
nblumhardt: `"-Infinity"`

test/Serilog.Tests/Formatting/Json/JsonValueFormatterTests.cs
```diff
@@ -49,7 +49,33 @@ public void DoubleFormatsAsNumber()
             JsonLiteralTypesAreFormatted(123.45, "123.45");
         }
 
-        [Fact]
+		[Fact]
+		public void DoubleSpecialsFormatAsString()
+		{
+			string format = "\"{0}\"";
+
+			JsonLiteralTypesAreFormatted(double.NaN, string.Format(format, double.NaN));
```
nblumhardt: `"NaN"`

test/Serilog.Tests/Formatting/Json/JsonValueFormatterTests.cs
```diff
@@ -49,7 +49,33 @@ public void DoubleFormatsAsNumber()
...
+		{
+			JsonLiteralTypesAreFormatted(123.45f, "123.45");
+		}
+
+		[Fact]
+		public void FloatSpecialsFormatAsString()
+		{
+			string format = "\"{0}\"";
+
+			JsonLiteralTypesAreFormatted(float.NaN, string.Format(format, float.NaN));
```
nblumhardt: As above, I think we should use string literals here for clarity.

clarkis117:
@nblumhardt Do yo also want the use of `string.Format` removed from the two unit tests?

nblumhardt:
@clarkis117 I'm happy with it either way - whichever you think is the clearest to read. Cheers!

nblumhardt:
:+1:

PR #888, opened 2016-10-30
Allow overriding log level via config files

optical:
An attempt to resolve https://github.com/serilog/serilog-settings-appsettings/issues/2

Have implemented what I understand to be the main issue described there. Let me know if there is more to be done.


nblumhardt:
Excellent, thanks @optical!

Issue #878, opened 2016-10-12
Find and follow up test coverage opportunities


nblumhardt:
Thanks to @304NotModified, we now get test coverage reports out of each build through codecov.io.

Here's the current dev branch coverage: https://codecov.io/gh/serilog/serilog/tree/dev/src/Serilog

At first glance there's a lot that isn't covered by our tests, but digging deeper a lot of what isn't covered appears to be pretty trivial.

This ticket is really looking for someone to own the task of going through the coverage report and finding opportunities to meaningfully improve test coverage with some additional tests. Any takers?

clarkis117:
It looks like the class `Core.Logger` (in logger.cs) has many methods that use one of the `Core.Logger.Write...` methods which aren't covered.

nblumhardt:
@clarkis117 yes, I think that's a large part of it.

I was tempted to ignore these, but then realized it's a great opportunity to apply a reflection-driven "convention test"; i.e.:

 * loop over all of the logging levels from `Verbose` to `Fatal`
 * for each method
  - use reflection to find each expected variant
  - call the method through reflection and verify the result

The actual test would end up being quite terse I think - and we'd then catch any anomalies like missing overloads.

We could even go a step further here and validate parameter names and so-on.

What do you think - does this sound sane? :-)

clarkis117:
Considering that there’s 60 methods in question, it does sound sane. I was thinking about having a unit test for each log level group. Then, there would be a function to test for each signature in the convention listed below. Each method would be evaluated to see if it matches one of the signatures listed below. If the method does not match one of the signatures the test would fail. Upon matching a signature, the method would be invoked with some arbitrary arguments.

Each method should also be evaluated that it’s public, returns void, and has the `MessageTemplateFormatMethod` Attribute.

To clarify the convention that needs to be validated is (for each log level):
```
Method0 (string messageTemplate) : void
Method1<T> (string messageTemplate, T propertyValue) : void
Method2<T0, T1> (string messageTemplate, T0 propertyValue0, T1 propertyValue1) : void
Method3<T0, T1, T2> (string messageTemplate, T0 propertyValue0, T1 propertyValue1, T2 propertyValue2) : void
Method4 (string messageTemplate, params object[] propertyValues) : void
Method5 (Exception exception, string messageTemplate) : void
Method6<T> (Exception exception, string messageTemplate, T propertyValue) : void
[...]

PR #896, opened 2016-11-04
Convention tests for log level write methods in Serilog.Core.Logger

clarkis117:
As proposed in #878, tests that check the convention of the log level wrapper methods in` Serilog.Core.Logger`. The tests reflect over the methods in `Serilog.Core.Logger` , that match known log level names. Then the methods and their parameters are checked against several rules that match the signatures listed below. If a method matches a know signature, it is invoked with some arbitrary data. If a method does not match a know signature, the test for its log level group fails with the unmatched method printed out as an error message.

```
Method0 (string messageTemplate) : void
Method1<T> (string messageTemplate, T propertyValue) : void
Method2<T0, T1> (string messageTemplate, T0 propertyValue0, T1 propertyValue1) : void
Method3<T0, T1, T2> (string messageTemplate, T0 propertyValue0, T1 propertyValue1, T2 propertyValue2) : void
Method4 (string messageTemplate, params object[] propertyValues) : void
Method5 (Exception exception, string messageTemplate) : void
Method6<T> (Exception exception, string messageTemplate, T propertyValue) : void
Method7<T0, T1> (Exception exception, string messageTemplate, T0 propertyValue0, T1 propertyValue1) : void
Method8<T0, T1, T2> (Exception exception, string messageTemplate, T0 propertyValue0, T1 propertyValue1, T2 propertyValue2) : void
Method9 (Exception exception, string messageTemplate, params object[] propertyValues) : void
```

### DIFF
--- a/test/Serilog.Tests/Core/LoggerTests.cs
+++ b/test/Serilog.Tests/Core/LoggerTests.cs
@@ -6,6 +6,9 @@ using Xunit;
 using Serilog.Core;
 using Serilog.Events;
 using Serilog.Tests.Support;
+using System.Reflection;
+using Xunit.Sdk;
+using System.Text.RegularExpressions;
 
 namespace Serilog.Tests.Core
 {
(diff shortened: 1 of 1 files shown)


test/Serilog.Tests/Core/LoggerTests.cs
```diff
@@ -116,5 +119,266 @@ public void PropertiesCanBeBound()
...
+			ValidateConventionForMethodSet("Error");
+		}
+
+		[Fact]
+		public void FatalMethodsMatchConvention()
+		{
+			ValidateConventionForMethodSet("Fatal");
+		}
+
+		private void ValidateConventionForMethodSet(string setName)
```
nblumhardt: The methods above can be avoided with a data-driven test:

```csharp
[Theory]
[InlineData(LogEventLevel.Verbose)]
[InlineData(LogEventLevel.Debug)]
// ...
public void ValidateConventionForMethodSet(LogEventLevel level)
{
    var setName = level.ToString();
    // ...
```

test/Serilog.Tests/Core/LoggerTests.cs
```diff
@@ -116,5 +119,266 @@ public void PropertiesCanBeBound()
...
+					}
+				}
+
+				Assert.True(signatureMatch, $"{method} did not match any known convention");
+			}
+		}
+
+
+		// Method0 (string messageTemplate) : void
+		private void ValidateMethod0(MethodInfo method)
```
nblumhardt: The codebase doesn't generally use explicit `private` or `internal` modifiers.

test/Serilog.Tests/Core/LoggerTests.cs
```diff
@@ -116,5 +119,266 @@ public void PropertiesCanBeBound()
...
+
+						signatureMatch = true;
+
+						break;
+					}
+					catch (Exception e)
+					when (e is TargetInvocationException
+							&& ((TargetInvocationException)e).GetBaseException() is XunitException)
+					{
+						continue;
```
nblumhardt: I'm assuming the exceptions here are assertion failures - will they be re-thrown or bubbled up somehow?

test/Serilog.Tests/Core/LoggerTests.cs
```diff
@@ -116,5 +119,266 @@ public void PropertiesCanBeBound()
...
+			VerifyMethodSignature(method, hasExceptionArg: true, expectedArgCount: 3);
+
+			GetLoggerAndInvoke(method, new Exception("test"), "message", new object[] { "value0", "value1", "value2" });
+		}
+
+		private static void GetLoggerAndInvoke(MethodInfo method, params object[] parameters)
+		{
+			var logger = new LoggerConfiguration().CreateLogger();
+
+			method.Invoke(logger, parameters);
```
nblumhardt: Could we do something like:

```csharp
var sink = new CollectingSink();

    .MinimumLevel.Verbose()
    .WriteTo.Sink(sink)
    .CreateLogger();

method.Invoke(logger, parameters);

Assert.Equal(1, sink.Events.Count);

// By passing the level through we could also do:
var evt = sink.Events.Single();
Assert.Equal(level, evt.Level);
```

nblumhardt:
Looks great! Added some comments.

test/Serilog.Tests/Core/LoggerTests.cs
clarkis117: something like this should bubble them up
```
				var signatureMatch = false;

				var report = new StringBuilder();

				foreach (var testMethod in testMethods)
				{
					try
					{
						testMethod.Invoke(this, new object[] { method });

						signatureMatch = true;

						break;
					}
					catch (Exception e)
						when (e is TargetInvocationException
							&& (e as TargetInvocationException).GetBaseException() is XunitException)
					{
						var xuintException = (XunitException)(e as TargetInvocationException).GetBaseException();

						report.AppendLine($"Method {method} failed on {testMethod} with assertion failure: {xuintException.UserMessage}");

						continue;
					}
				}

				Assert.True(signatureMatch, $"{method} did not match any known convention\n" + report.ToString());
```

test/Serilog.Tests/Core/LoggerTests.cs
clarkis117: I wasn't sure how to best configure the logger, will update it. :)

test/Serilog.Tests/Core/LoggerTests.cs
clarkis117: But still leaves the case of determining if an Assertion that is not in the Signature test fails... I should have a try catch statement in it, that adds a boolean value to the exception. It can be used to discriminate between signature test and invocation assertion failures.

clarkis117:
Also of note, the same or similar convention based tests could also be applied to the static class Serilog.Log. 
https://codecov.io/gh/serilog/serilog/src/dev/src/Serilog/Log.cs

nblumhardt:
Given the amount of infrastructure, and the desire to also test `Log` (and possibly `ILogger`) in the future, perhaps we move this to `MethodOverloadConventionTests.cs` or similar to keep things grouped/organized? Otherwise LGTM 👍

clarkis117:
If it’s okay, you can go ahead and merge this PR. I’ll add test coverage for `Serilog.Log` and `ILogger` in a separate PR and move the code out of `LoggerTests.cs` into `MethodOverloadConventionsTests.cs`, since it will require some modifications to the existing work. It would also probably be cleaner to keep the test code just in `MethodOverloadConventionsTests.cs`.

nblumhardt:
Great, thanks! :-) :+1:

Issue #880, opened 2016-10-17
Convention (or something) to force static dependency to Sinks assemblies loaded dynamically


caioproiete:
(ideas ... to be discussed).

---

This is not a new issue (it has always been like that), but it is more apparent now that every sink is in its own assembly.

When you have Sinks that are loaded dynamically, for example, via App.config, e.g.:

```
    Log.Logger = new LoggerConfiguration()
        .ReadFrom.AppSettings()
        .CreateLogger();

    <?xml version="1.0" encoding="utf-8" ?>
    <configuration>
      <appSettings>
        <add key="serilog:minimum-level" value="Verbose" />

        <add key="serilog:using:RollingFile" value="Serilog.Sinks.RollingFile" />
        <add key="serilog:write-to:RollingFile.restrictedToMinimumLevel" value="Verbose" />
        <add key="serilog:write-to:RollingFile.pathFormat" value="..\log\myapp-{Date}.log" />
      </appSettings>
    </configuration>
```

There is no static link between the app and the sinks' assemblies and its dependencies (in this case, `Serilog.Sinks.File`, and `Serilog.Sinks.RollingFile`).

This means that if your log configuration lives on `ClassLibrary1`, which is referenced by `ConsoleApp1`, the assemblies for the sinks will **not** be copied to the bin folder of `ConsoleApp1`.

Also, even if the log configuration is defined in `ConsoleApp1`, tools like ReSharper will tell you (rightfully so) that it is safe to remove references to these assemblies, in whatever project you installed the NuGet packages, so you have to remember that every time you're doing a clean-up.

![image](https://cloud.githubusercontent.com/assets/177608/19449751/3c932836-947d-11e6-94f4-42c90efdeb44.png)

---

Couple of ideas that come to mind, which could solve this issue:

**1.) Add a convention in which each Sink exposes an extension method such as `UseSinkName` or `RequireSinkName`**, that developers can optionally call when configuring the logger. e.g.

```
    Log.Logger = new LoggerConfiguration()
        .UsingRollingFileSink()
        .UsingSeqSink()
        .ReadFrom.AppSettings()
        .CreateLogger();
```

These `UsingSinkName` methods, of course, would be in each of the Sinks' assemblies, and therefore create the static dependency. They wouldn't do anything other than return the same `LoggerConfiguration` passed in.

[...]

PR #898, opened 2016-11-09
Add ConfigurationDependencyAttribute to help with referenced assemblies ...

caioproiete:
… not being copied over.

As per discussion on #880

### DIFF
--- a/src/Serilog/Settings/ConfigurationDependencyAttribute.cs
+++ b/src/Serilog/Settings/ConfigurationDependencyAttribute.cs
@@ -1,0 +1,27 @@
+﻿using System;
+
+namespace Serilog.Settings
+{
+    /// <summary>
+    /// Indicates that the marked assembly depends on the type that is specified in the constructor.
+    /// Typically used to force a compile-time dependency to the assembly that contains the type.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+    public sealed class ConfigurationDependencyAttribute : Attribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ConfigurationDependencyAttribute"/> class.
+        /// </summary>
+        /// <param name="dependencyType">A type from the assembly that is used dynamically.</param>
+        public ConfigurationDependencyAttribute(Type dependencyType)
+        {
+            DependencyType = dependencyType;
+        }
+
+        /// <summary>
+        /// Gets the dependent type reference.
+        /// </summary>
+        /// <value>The dependent type reference.</value>
+        public Type DependencyType { get; private set; }
+    }
+}
--- a/test/Serilog.Tests/Settings/ConfigurationDependencyAttributeTests.cs
+++ b/test/Serilog.Tests/Settings/ConfigurationDependencyAttributeTests.cs
@@ -1,0 +1,15 @@
+﻿using Xunit;
+using Serilog.Settings;
+
+namespace Serilog.Tests.Settings
+{
+    public class ConfigurationDependencyAttributeTests
+    {
+        [Fact]
+        public void DependencyTypePropertyContainsTypePassedInConstructor()
+        {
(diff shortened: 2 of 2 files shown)


src/Serilog/Build/ImplicitDependencyAttribute.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace Serilog.Build
```
nblumhardt: I still think `Serilog.Settings.ConfigurationDependencyAttribute` is a more self-describing name and namespace for the type. The attribute is only likely to be used in conjunction with XML or JSON settings, so focusing it on that scenario has some discoverability benefits even if it could be used for other purposes.

src/Serilog/Build/ImplicitDependencyAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+        public ImplicitDependencyAttribute(Type dependentType)
+        {
+            DependentType = dependentType;
+        }
+
+        /// <summary>
+        /// Gets the dependent type reference.
+        /// </summary>
+        /// <value>The dependent type reference.</value>
+        public Type DependentType { get; private set; }
```
nblumhardt: _Dependent_, to me, is the thing that "does-the-depending" (i.e. the inverse direction). Maybe `Dependency`?

src/Serilog/Build/ImplicitDependencyAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Indicates that the marked assembly depends on a type that is specified in the constructor.
+    /// Typically used to force a compile-time dependency to the assembly that contains the type.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+    public sealed class ImplicitDependencyAttribute : Attribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ImplicitDependencyAttribute"/> class.
+        /// </summary>
+        /// <param name="dependentType"></param>
```
nblumhardt: `A type from the assembly that is used dynamically.`

caioproiete:
@nblumhardt PR amended based on your comments

nblumhardt:
Looks good, thanks @caioproiete.

I wondered for a moment if `DependencyType` might be an instance of Hungarian notation, but on consideration, the `Type` may not be the actual `Dependency`. Any thoughts on this?

caioproiete:
> I wondered for a moment if DependencyType might be an instance of Hungarian notation, but on consideration, the Type may not be the actual Dependency. Any thoughts on this?

No strong feelings there. The `Type` suffix seems commonly used for property exposing a `Type` ([1](https://github.com/dotnet/corert/blob/26413145abcbdd49468db230a6f29784494f6f46/src/System.Private.CoreLib/src/System/Reflection/InterfaceMapping.cs#L9-L10), [2](https://github.com/dotnet/roslyn/blob/614299ff83da9959fa07131c6d0ffbc58873b6ae/src/Compilers/Core/Portable/Compilation/ScriptCompilationInfo.cs#L10-L11), [3](https://github.com/dotnet/corert/blob/26413145abcbdd49468db230a6f29784494f6f46/src/System.Private.CoreLib/src/System/Runtime/CompilerServices/StateMachineAttribute.cs#L12), [4](https://github.com/dotnet/roslyn/blob/614299ff83da9959fa07131c6d0ffbc58873b6ae/src/ExpressionEvaluator/Core/Test/ResultProvider/Debugger/Engine/DkmReportNonFatalWatsonExceptionAttribute.cs#L21), ...).

But yes, the `Type` is not really the dependency... The dependency is the assembly and in an ideal world, we would like to do something like `[ConfigurationDependencyAttribute(typeof(SinkName).Assembly)]` but, of course, that doesn't work on attributes.

nblumhardt:
👍 let's stick with `DependencyType`.

Need to investigate #880 a little further before merging, but LGTM here.

caioproiete:
Bug fixed with most recent version of ReSharper.

PR #902, opened 2016-11-16
Moved convention tests, and added support for other ILogger based classes

clarkis117:
As per mention in #896 convention tests for other classes based on `Serilog.Ilogger`

* Previous convention based tests were moved to their own file `MethodOverloadConventionTests.cs` 
* added support for other ILogger based classes: `Serilog.Log`, `Serilog.Ilogger`, and `Serilog.Core.Logger`
* added support for `Write` methods that are convention based


nblumhardt:
Ironically, looks like code coverage testing failed the build somehow... Kicking if off again to see whether it was some intermittent condition...

clarkis117:
CI randomly failing is odd, considering the changes didn't touch any of the noted failed tests according to the CI.

test/Serilog.Tests/MethodOverloadConventionTests.cs
```diff
@@ -0,0 +1,372 @@
...
+				paramList.AddRange(parameters);
+
+				parameters = paramList.ToArray();
+			}
+			else
+				Assert.True(Enum.TryParse(method.Name, out level));
+
+			if (method.IsStatic)
+			{
+				Log.Logger = logger;
```
nblumhardt: Here's the problem.

nblumhardt:
Unfortunately the test that's failing looks at `Log` in its uninitialized state. This means it needs to run ahead of the test here that assigns `Log.Logger`. Not sure how that should be achieved.. (The tests so far, for better or worse, have avoided the static logger since parallelization and so-on makes those kinds of tests fragile :-)).  Any ideas?

clarkis117:
I think the appropriate behavior would be to put all tests that touch `Serilog.Log` into the same test collection via xUnit's `CollectionAttribute`. This indicates to xUnit that the tests should not be run in parallel against each other, as xUnit defaults to not running tests in the same collection, in parallel, against each other.

nblumhardt:
👍 - that combined with some awareness of ordering should do it.

nblumhardt:
:+1: great stuff :-)

PR #903, opened 2016-11-18
ILogger API conformance fixes in Serilog.Log

clarkis117:
I found several differences in `Serilog.Log` versus the `Serilog.ILogger` interface. This PR brings `Serilog.Log` into parity with `Serilog.ILogger`.  

1.   `ForContext` method that takes multiple `ILogEventEnricher` takes an array versus `IEnumerable<ILogEventEnricher>` in `Serilog.ILogger`. This changes it so that it takes an `IEnumerable<ILogEventEnricher>`.

`ILogger ForContext(ILogEventEnricher[] enrichers)`  -->
  `ILogger ForContext(IEnumerable<IlogEventEnricher> enrichers)`

2. Adds a `BindProperty` method to `Serilog.Log` 

 `public static bool BindProperty(string propertyName, object value, bool destructureObjects, out LogEventProperty property)`

### DIFF
--- a/src/Serilog/Log.cs
+++ b/src/Serilog/Log.cs
@@ -81,7 +81,7 @@ namespace Serilog
         /// </summary>
         /// <param name="enrichers">Enrichers that apply in the context.</param>
         /// <returns>A logger that will enrich log events as specified.</returns>
-        public static ILogger ForContext(IEnumerable<ILogEventEnricher> enrichers)
+        public static ILogger ForContext(ILogEventEnricher[] enrichers)
         {
             return Logger.ForContext(enrichers);
         }
--- a/src/Serilog/Log.cs
+++ b/src/Serilog/Log.cs
@@ -81,7 +81,7 @@ namespace Serilog
         /// </summary>
         /// <param name="enrichers">Enrichers that apply in the context.</param>
         /// <returns>A logger that will enrich log events as specified.</returns>
-        public static ILogger ForContext(ILogEventEnricher[] enrichers)
+        public static ILogger ForContext(IEnumerable<ILogEventEnricher> enrichers)
         {
             return Logger.ForContext(enrichers);
         }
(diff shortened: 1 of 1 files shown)


nblumhardt:
Nice finds.

I find the `ForContext()` discrepancy a bit annoying, but I don't think we should take the breaking change required to bring these into line. Any code calling the `Log.ForContext(ILogEventEnricher[])` method would need re-compiling after the change, and in general we try to avoid breakage on the "consumer" side of the API (loggers, log context and so-on) at pretty much all costs.

2.0 was, I think, our last opportunity to fix this one, at least until our hand is forced and we have to take a breaking change for some other reason. Thankfully it's pretty minor. We could create an issue for it and bury it until some time down the track?

Happy adding the `BindProperty()` method; looking back I am not sure it made a whole lot of sense to add `BindMessageTemplate()` here, but since we have one, we might as well have the other 👍 

BTW, I'm liking the coverage reports from @codecov-io, but the success/fail status reporting I could do without. Wonder if there's any way to have one without the other?

clarkis117:
Gotcha, I'll add another commit that reverses the changes to the `ForContext` method and open an issue for it. Are there any labels specific to issues for the next major release? For Codecov, you should be able to set a threshold for max percent drop in code coverage allowed per commit. `threshold: number` in its config file. If the value is exceed then it will mark the checks as failed, at least that's what I interpreted from the docs (http://docs.codecov.io/docs/commit-status#section-project-status).

nblumhardt:
Thanks for the pointer, sounds like the way to go 👍 

It looks suspiciously like the test ordering issue is still causing some intermittent failure:

```
xUnit.net .NET CLI test runner (64-bit .NET Core win81-x64)
  Discovering: Serilog.Tests
  Discovered:  Serilog.Tests
  Starting:    Serilog.Tests
dotnet.exe :     Serilog.Tests.Core.LogEventPropertyCapturingTests.ProvidingLessParametersThanNamedPropertiesInTheTemplateWorksAndSelfLogs [FAIL]
At C:\projects\serilog\Build.ps1:45 char:5
+     & dotnet test -c Release
+     ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (    Serilog.Tes...SelfLogs [FAIL]:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
      System.InvalidOperationException : Collection was modified; enumeration operation may not execute.
      Stack Trace:
           at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
           at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
        C:\projects\serilog\test\Serilog.Tests\Core\LogEventPropertyCapturingTests.cs(91,0): at Serilog.Tests.Core.LogEventPropertyCapturingTests.ProvidingLessParametersThanNamedPropertiesInTheTemplateWorksAndSelfLogs()
```

https://ci.appveyor.com/project/serilog/serilog/build/752

Doesn't seem to have been triggered specifically by this PR; I'm guessing #902 may still need a tweak.

PR #905, opened 2016-11-23
Disable parallel test execution

nblumhardt:
It would be nicer to just find all tests that touch static state and put them in a single assembly, but unfortunately, `SelfLog` can be written to from anywhere and this would disrupt tests that check `SelfLog` behaviour.

Compacted and tidied a few small things while I was at it.

@clarkis117


clarkis117:
Just one thing, I don’t know what Xunit’s default collection behavior is, but to make explicit it this should be set. It should make sure that the entire assembly is executed serially.
`[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly)]`

nblumhardt:
Not sure about that - the default is _collection-per-class_, it may be just as confusing if we start re-specifying the defaults.... (The reader who knows what the default should be will then find themselves asking "Hey, isn't that the default? I wonder why that's there... Maybe an Xunit bug?" etc.)

merbla:
LGTM

clarkis117:
I ran 900 runs of all the unit tests without a single test failure, LGTM

PR #907, opened 2016-11-23
Dispose audit sinks on ILogger.Dispose()

skomis-mm:
Also added test to cover this and moved audit sink related tests to a separate collection.


nblumhardt:
Great catch 👍

PR #909, opened 2016-11-24
Added tests for the rest of ILogger and added support for SilentLogger to Method Overload tests

clarkis117:
This PR, as per #878, adds convention/reflection based tests for the remaining methods defined in `Serilog.ILogger` i.e. `ForContext` and `BindMessage/Property`. It also adds support for testing `Serilog.Core.Pipeline.SilentLogger`.

### DIFF
--- a/test/Serilog.Tests/MethodOverloadConventionTests.cs
+++ b/test/Serilog.Tests/MethodOverloadConventionTests.cs
@@ -219,10 +219,17 @@ namespace Serilog.Tests
 
             Assert.IsType(typeof(bool), result);
 
+            //silentlogger is always false
             if (loggerType == typeof(SilentLogger))
-                Assert.False(result as bool?);
-            else
-                Assert.True(result as bool?);
+                return;
+            
+            Assert.True(result as bool?);
+
+            //test null arg path
+            var falseResult = InvokeMethod(method, logger, new object[] { null, null, null, null });
+
+            Assert.IsType(typeof(bool), falseResult);
+            Assert.False(falseResult as bool?);
         }
 
         [Theory]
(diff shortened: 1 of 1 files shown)


clarkis117:
Seems, I forgot one method defined in `Serilog.ILogger` `IsEnabled`... adding another commit with a test for it.

test/Serilog.Tests/MethodOverloadConventionTests.cs
```diff
@@ -12,361 +13,864 @@
...
+                Log.CloseAndFlush();
+
+                Log.Logger = new LoggerConfiguration()
+                    .MinimumLevel.Verbose()
+                    .WriteTo.Sink(sink)
+                    .CreateLogger();
+
+                return null;
+            }
+            else
```
nblumhardt: Perhaps adding the `if (loggerType == typeof(ILogger) || loggerType == typeof(SilentLogger)` test here, and an `else` branch that throws `ArgumentException`, will help prevent issues creeping in?

clarkis117: @nblumhardt  Does this look better?
```C#
                return null;
            }
            else if (loggerType == typeof(SilentLogger))
                return new SilentLogger();
            else
                throw new ArgumentException($"Logger Type of {loggerType} is not supported");
```

clarkis117:
Interesting, looks like @codecov-io  now edits it's comments instead posting new ones. Also, I still have a few partial hits and misses to narrow down.

clarkis117:
@nblumhardt this should be ready for review now

nblumhardt:
:+1:

Issue #924, opened 2017-01-04
Dynamic/expression-based filtering


nblumhardt:
Filtering in Serilog is currently based on predicates - `Func<LogEvent, bool>` - and some helpers to generate common ones like `Matching.FromSource()`.

In some situations, writing a predicate over a log event can be hard. For example, excluding events with a particular property value:

```csharp
    .Filter.ByExcluding(e => {
        LogEventPropertyValue p;
        if (!e.Properties.TryGetValue("P", out p) ||
          !(p is ScalarValue) || !(((ScalarValue)p).Value is int!))
            return false;
        return 3 != (int)((ScalarValue)p).Value;
    })
```

...or something like that :-). The Serilog representation of structured data, and the need to exactly match or convert the boxed value types (casting to `long` would fail here) makes it extremely error prone.

Filters to do more complex things with structured data are pretty much out of the reach of casual users.

In other cases - XML or JSON configuration - it's not possible to supply a predicate, so other hoops need to be jumped through.

I've spiked an expression-based filtering system using bits and pieces of code from other projects. It supports an SQL-like predicate syntax (think `WHERE` clause body) and is pretty unsurprising:

```csharp
var expr = "@Level = 'Information' and AppId is not null and Items[?] like 'C%'";

Log.Logger = new LoggerConfiguration()
    .Enrich.WithProperty("AppId", 10)
    .Filter.ByIncludingOnly(expr)
    .WriteTo.LiterateConsole()
    .CreateLogger();

// Printed
Log.Information("Cart contains {@Items}", new[] { "Tea", "Coffee" });
Log.Information("Cart contains {@Items}", new[] { "Peanuts", "Chocolate" });

// Not printed
Log.Warning("Cart contains {@Items}", new[] { "Tea", "Coffee" });
Log.Information("Cart contains {@Items}", new[] { "Apricots" });

Log.CloseAndFlush();
```

This is wrapped up in a library tentatively called _Serilog.Filters.Expressions_, which is not yet finalized but can be seen at: https://github.com/nblumhardt/serilog-filters-expressions

Though it's a fairly large amount of somewhat hairy code, I'd like to propose we adopt this in the Serilog project and make it available alongside the other supported libraries. This would encourage us to get the work done to make it fit nicely into the configuration system, also.

```xml
  <add key="serilog:filter:ByExcluding.expression"
            value="SourceContext = 'MyApp.Program'" />
```

[...]

PR #925, opened 2017-01-04
Refactor key-value settings to consistently apply configuration methods

nblumhardt:
Enables `serilog:filter:Xyz` configuration directives. Cleaned up the code as much as possible in the process, so that `audit-to`, `write-to`, `enrich` and `filter` are consistently handled. Still some cleanup possible, but we're in better shape now :-)

Would support #924.


nblumhardt:
Belated reviews welcome :-) moving this one on so that it can get a workout in -dev.

PR #932, opened 2017-01-28
Enable builds for OSX on Travis #824

merbla:
Entire rework of Travis builds for #824.

* Changed to use Travis OSX example from dotnet/docs. See  [details](https://github.com/dotnet/docs/blob/master/.travis.yml) here.

Alternative is to slightly update previous build config as seen [here](https://github.com/merbla/serilog/blob/enable-osx-for-travis/.travis.yml).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,45 +1,14 @@
 language: csharp
 
-#dotnet cli require Ubuntu 14.04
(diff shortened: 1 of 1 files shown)


adamchester:
Much more clean 👍

PR #944, opened 2017-03-06
Support {Properties} in output templates

### LINKED ISSUE #825: Output template support for {Properties}
Via: http://stackoverflow.com/questions/38894310/serilog-saving-event-properties-added-with-forcontext-into-flat-files

When writing log files (and to the console), it's sometimes desirable to include various properties not formatted into the message, and not explicitly specified in the output template.

This might look like:

``` csharp
    .WriteTo.Console(outputTemplate: "{Level:u2} {Message} {Properties}{NewLine}")
    .CreateLogger() ;

Log.ForContext("Foo", 42)
    .Information("Hello from {Bar}!", "bar");

// -> INF Hello from "bar"! {Foo: 42}
```

Note that the properties written should exclude those already formatted into the template; it's a text logging scenario, if a full property listing is desired one of the JSON formatters will cover that case.

Marked `breaking` because this will interfere will logs that already use a static `{Properties}` element in their output templates, though in practice I can't see this appearing in many places.

Pliner:
Hi guys!

Here is a small addition to the Serilog 😉


nblumhardt:
Thanks for the PR, @Pliner!

It doesn't look like this implementation excludes the properties accounted for by the message template (see the example in #825).

Pliner:
@nblumhardt Thank you for the feedback. Without any doubts I missed that requirement 😞 

After brief investigation I found two ways to fix it:
1. Check MessageTemplate.NamedProperties
2. Filter MessageTemplate.Tokens to PropertyToken[] and check.

Could you tell me what way you prefer or maybe you can advice me another one? Thanks...

Pliner:
@nblumhardt Anyway, I pushed fix with inspecting MessageTemplate.NamedProperties.

src/Serilog/Formatting/Display/LogEventPropertiesValue.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        public override void Render(TextWriter output, string format = null, IFormatProvider formatProvider = null)
+        {
+            output.Write('{');
+
+            var delim = "";
+            foreach (var kvp in _properties)
+            {
+                if (_template.NamedProperties.Any(x => x.PropertyName == kvp.Key))
```
nblumhardt: Since `NamedProperties` is an array, this can use indexer-based access and avoid allocating a LINQ enumerator. (Keeping an eye on allocations, as they build up.)

nblumhardt:
Looking good! Just the minor comment.

@serilog/reviewers-core any thoughts on this one?

nblumhardt:
Cool! :-) Me too. I think we should just make the LINQ change then hit the button.

Pliner:
@nblumhardt I hope I fixed your comment 😉

src/Serilog/Formatting/Display/LogEventPropertiesValue.cs
```diff
@@ -0,0 +1,68 @@
...
+                output.Write(": ");
+                kvp.Value.Render(output, null, formatProvider);
+            }
+
+            output.Write('}');
+        }
+
+        static bool TemplateContainsPropertyName(MessageTemplate template, string propertyName)
+        {
+            foreach (var namedProperty in template.NamedProperties)
```
nblumhardt: @Pliner I actually had the ol' `for(var i = ...` in mind here.

nblumhardt:
Added a clarification. Thanks @Pliner!

Just stepping back for a second, there's one scenario where this might not be enough to achieve what we want.

I.e. in [Serilog.Extensions.Logging.File](https://github.com/serilog/serilog-extensions-logging-file) we use the output template:

```
{Timestamp:o} {RequestId,13} [{Level:u3}] {Message} ({EventId:x8}){NewLine}{Exception}
```

To take advantage of this change, we'd want something resembling:

```
{Timestamp:o} {RequestId,13} [{Level:u3}] {Message} ({EventId:x8}) {Properties}{NewLine}{Exception}
```

Although we've got `RequestId` and `EventId` going out on every line, they'll still be repeated in the `{Properties}` block.

Maybe not a show-stopper, but possibly worth a second thought?

Pliner:
@nblumhardt To be honest, log4net approach is ok for me: if %properties% is used in patternlayout, all properties from all contexts will be captured with any exclusions.

nblumhardt:
Thanks for the follow-up Yuri. I guess the difference here is that for log4net, the `%properties%` layout will not duplicate data that's in the log message, because log4net doesn't capture the arguments to the message format string the way Serilog does. With Serilog, there will be 100% duplication of these if we don't exclude them.

Noctis-:
Hey guys, what's the reasoning behind preferring the for loop to the foreach one ?

nblumhardt:
@Noctis- microbenchmarks :-) ... less allocation.

nblumhardt:
@serilog/reviewers-core any thoughts/opinions on https://github.com/serilog/serilog/pull/944#issuecomment-285579828 ?

adamchester:
I think `{Properties}` should not include those properties already rendered in the message template.

adamchester:
Another quick note that provides more evidence...  in Logary we also implemented it by ignoring the [fields already in the template](https://github.com/logary/logary/blob/0fbb7829b9d4b62f9ffd10d7a4b20d075c6cccc6/src/Logary.Facade/Facade.fs#L685).

merbla:
Agreed, we should not duplicate.

Pliner:
@nblumhardt @adamchester @merbla 
Thank you for the feedback 😉 

Will make another commit today or tomorow.

nblumhardt:
Thank you @Pliner :-) 👍

Pliner:
@nblumhardt I'm not sure about my approach, but now I haven't any ideas how to do it differently :cry:

nblumhardt:
Great! I think it's about as good as we are likely to get, here, without deeper changes (which we could still consider in the future).

:shipit: ?

src/Serilog/Formatting/Display/OutputProperties.cs
```diff
@@ -72,7 +72,7 @@ public static class OutputProperties
             result[TimestampPropertyName] = new ScalarValue(logEvent.Timestamp);
             result[LevelPropertyName] = new LogEventLevelValue(logEvent.Level);
             result[NewLinePropertyName] = new LiteralStringValue(Environment.NewLine);
-            result[PropertiesPropertyName] = new LogEventPropertiesValue(logEvent.MessageTemplate, logEvent.Properties);
+            result[PropertiesPropertyName] = new LogEventPropertiesValue(logEvent.MessageTemplate, logEvent.Properties, result);
```
johnduhart: I would just add a comment before this line to remind future explorers to make sure this value always comes last so it picks up all the existing properties.

Pliner:
@nblumhardt Sorry, but I have to rewrite implementation, because agait is doesn't do everything right.

I should not use output properties, but use output message template instead.

Pliner:
@adamchester Unfortunatly yes, because previous implementation is simply not working 😞

src/Serilog/Formatting/Display/OutputProperties.cs
```diff
@@ -51,11 +51,17 @@ public static class OutputProperties
...
+        public const string PropertiesPropertyName = "Properties";
+
+        /// <summary>
         /// Create properties from the provided log event.
         /// </summary>
         /// <param name="logEvent">The log event.</param>
+        /// <param name="outputTemplate">The output template.</param>
         /// <returns>A dictionary with properties representing the log event.</returns>
-        public static IReadOnlyDictionary<string, LogEventPropertyValue> GetOutputProperties(LogEvent logEvent)
+        public static IReadOnlyDictionary<string, LogEventPropertyValue> GetOutputProperties(LogEvent logEvent, MessageTemplate outputTemplate)
```
nblumhardt: Unfortunately, this will break the literate console sink binary:

https://github.com/serilog/serilog-sinks-literate/blob/dev/src/Serilog.Sinks.Literate/Sinks/Literate/LiterateConsoleSink.cs#L88

One obvious option would be to:

 1. Keep this overload
 2. Add back the one without the additional parameter, but make it forward calls to the full version (we could pass `MessageTemplate.Empty` as `outputTemplate`, so that the arguments could still be null-checked)
 3. Mark the old overload `[Obsolete("Pass the full output template using the other overload.")]`

An alternative, which might be worth considering, is to add the new overload as `internal`, mark the old one obsolete, and stop offering a public `GetOutputProperties()` at some time in the future. With the literate console being the only consumer, it may be defensible to simply duplicate the code there.

I like the latter alternative because it'd allow us to re-think how the `MessageTemplateTextFormatter` works in the future - hopefully to improve on the O(N) scans of the message templates - without then needing to break further public API surface.

Any thoughts?

nblumhardt:
@Pliner good spotting - thanks for the thorough check :-)

Pliner:
@nblumhardt I made an obvious changes for backward compatibility as you mentioned before.

nblumhardt:
🎉 thanks for the great contribution, Yuri!

adamchester:
img_0054

PR #946, opened 2017-03-09
Use dotnet core sdk 1.0.1

adamchester:
According to this link https://github.com/dotnet/cli/releases/latest, the current version is 1.0.1
image

If I do `File->New->Project(.NET Core: Console App)` then put `dotnet --version` in the pre-build event, I see `1.0.1`

image


nblumhardt:
Looks like the way to go 👍

Issue #134, opened 2014-05-05
Asynchronous support throughout Serilog


nathan-alden-sr:
Practically all of Serilog could benefit from asynchronous programming with .NET 4.0+ and  C# 5.0. For example, the `ILogger` interface currently lacks `*Async` methods. This means I have to write my own `Task.Run()` wrappers in order to get basic asynchronous support. Serilog interfaces with many external systems that really should use the TPL--ElasticSearch, file system, CouchDB, etc.

I would love to see a Serilog 2.0 that exposed all its existing interfaces asynchronously. Let's be sure we're using server threads as efficiently as possible!

Mystere:
I think a lack of async/await support is a mistake.  Particularly since it seems that you are basing this decision upon the implementation of http based sinks that use timers to ship logs.  Much has been made about how dangerous using timers for such recurring jobs can be (see: http://haacked.com/archive/2011/10/16/the-dangers-of-implementing-recurring-background-tasks-in-asp-net.aspx/) for example.  It's recommended that if you're using Timers, you should register them with IRegisteredObject.  I know you're subscribing to shutdown events, but i'm uncertain if that is truly wise.

Given the pervasiveness of async/await going forward in ASP.NET and general .NET development, especially in ASP.NET 5, I would urge you to reconsider adding native Async versions of the API rather than using the timer based batching that you currently use.  I would also urge you to implement some kind of contextual isolation of these logs, particularly for durable ones so that we don't have to disable overlapped recycling, and have less chances of log file locking with multi-user scenarios.

SimonCropp:
@nblumhardt so a year later and do u still think this is a bad idea?

i think this 

>  if we awaited a network call for each event written to one of these we'd bomb application performance

is not the approach you would use. you would actually return the task all the way up and let the root application decide to await or not based on how important they think the log message is and the context they are in

am i missing something?

nblumhardt:
Hey @SimonCropp ! ... I think it would be _really_ interesting to create an async Serilog fork as a PoC, but yes, I still expect we'd hit some really tough trade-offs.

For example, Serilog supports multiple sinks in a pipeline, all of which might be asynchronous. The code that dispatches a single event to multiple sinks would need to itself return a task, so you end up having to create an array of the `Task`s returned from each sink's `Emit()` and return `Task.WhenAll(...)` from the dispatcher. That's a lot of allocation and runtime machinery to get into the action for every single log event.

[...]

PR #950, opened 2017-03-18
Changed sinks and loggers to async

aKzenT:
I think it is time to consider reopening the discussion about async support raised in #134 

Just as a starting point for discussion I changed the sink and logger interfaces to return a Task, implementing the async pattern, so that you can easily do your own experiments, benchmarks, etc.

Unit tests are all green and I already did some performance tuning. Feedback appreciated!

### DIFF
--- a/src/Serilog/Core/CompletedTask.cs
+++ b/src/Serilog/Core/CompletedTask.cs
@@ -1,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace Serilog.Core
+{
+    class CompletedTask
+    {
+        public static readonly Task Instance = Task.FromResult((object) null);
+    }
+}
--- a/src/Serilog/Core/ILogEventSink.cs
+++ b/src/Serilog/Core/ILogEventSink.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using Serilog.Events;
+using System.Threading.Tasks;
 
 namespace Serilog.Core
 {
@@ -25,6 +26,6 @@ namespace Serilog.Core
         /// Emit the provided log event to the sink.
         /// </summary>
         /// <param name="logEvent">The log event to write.</param>
-        void Emit(LogEvent logEvent);
+        Task Emit(LogEvent logEvent);
     }
 }
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -18,6 +18,7 @@ using Serilog.Core.Enrichers;
 using Serilog.Debugging;
 using Serilog.Events;
 using Serilog.Parameters;
+using System.Threading.Tasks;
 
 #pragma warning disable Serilog004 // Constant MessageTemplate verifier
 
@@ -193,13 +194,15 @@ namespace Serilog.Core
         /// <param name="level">The level of the event.</param>
         /// <param name="messageTemplate">Message template describing the event.</param>
         [MessageTemplateFormatMethod("messageTemplate")]
-        public void Write(LogEventLevel level, string messageTemplate)
+        public Task Write(LogEventLevel level, string messageTemplate)
         {
             // Avoid the array allocation and any boxing allocations when the level isn't enabled
             if (IsEnabled(level))
             {
-                Write(level, messageTemplate, NoPropertyValues);
+                return Write(level, messageTemplate, NoPropertyValues);
             }
+
+            return CompletedTask.Instance;
         }
 
         /// <summary>
@@ -209,13 +212,15 @@ namespace Serilog.Core
         /// <param name="messageTemplate">Message template describing the event.</param>
         /// <param name="propertyValue">Object positionally formatted into the message template.</param>
         [MessageTemplateFormatMethod("messageTemplate")]
-        public void Write<T>(LogEventLevel level, string messageTemplate, T propertyValue)
+        public Task Write<T>(LogEventLevel level, string messageTemplate, T propertyValue)
         {
             // Avoid the array allocation and any boxing allocations when the level isn't enabled
             if (IsEnabled(level))
             {
-                Write(level, messageTemplate, new object[] { propertyValue });
+                return Write(level, messageTemplate, new object[] { propertyValue });
             }
+
+            return CompletedTask.Instance;
         }
 
         /// <summary>
@@ -226,13 +231,15 @@ namespace Serilog.Core
         /// <param name="propertyValue0">Object positionally formatted into the message template.</param>
         /// <param name="propertyValue1">Object positionally formatted into the message template.</param>
         [MessageTemplateFormatMethod("messageTemplate")]
-        public void Write<T0, T1>(LogEventLevel level, string messageTemplate, T0 propertyValue0, T1 propertyValue1)
+        public Task Write<T0, T1>(LogEventLevel level, string messageTemplate, T0 propertyValue0, T1 propertyValue1)
         {
             // Avoid the array allocation and any boxing allocations when the level isn't enabled
             if (IsEnabled(level))
             {
-                Write(level, messageTemplate, new object[] { propertyValue0, propertyValue1 });
+                return Write(level, messageTemplate, new object[] { propertyValue0, propertyValue1 });
             }
+
+            return CompletedTask.Instance;
         }
 
         /// <summary>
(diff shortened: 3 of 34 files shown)


nblumhardt:
@aKzenT thanks for taking the time to spike this.

There's a lot to discuss around how this could move forward without breaking existing applications, and so-on, but we should defer that discussion until we know what the benefit of making a change would be. (We can't really weigh up the cost of changes vs. their benefits, without having some idea of the benefits to go on).

To do that, we have to know what the performance advantages of this approach would be, and which kinds of apps and scenarios would benefit from them.

For example, file logging might achieve a higher throughput using `async` by eliminating blocked threads, but this will only become apparent at a certain level of output (events/kB/sec to the file), with a certain level of contention, and when the file is actually doing I/O rather than writing to an OS page buffer that's already in memory. The cost will be longer latency, so knowing that number would also be interesting.

HTTP logging (Elasticsearch, Seq, Splunk...) aren't likely to receive any benefit because to make calls efficiently under load, events already need to be sent in batches, meaning Serilog today never waits on any network I/O. The database sinks also tend to do batched writes as things stand.

Finally, when logging is going to an in-memory target, performance will be reliably worse, just because of the overhead of operating the `Task` machinery vs the absences of it today.

Do you have any numbers from your testing so far (or from your other experiences with Serilog) that illustrate what we could gain from this?

Thanks again for the input,
Nick

aKzenT:
Hey,

thank you for your comments, Nick.

I think .NET core based apps with many users would mostly benefit from this.

Already in .NET core many IO based APIs are async only. So writing a sink nowadays for these APIs forces you to deal with async methods.

Arguably most real world cases WILL use some sort of IO based sink in their pipeline. Since IO is magnitudes slower that the logging infrastructure itself most sinks provide a buffered approach to speed up writing of these events.

In my eyes async could provide a nice alternative for this. For many scenarios where the increased latency of a file write, etc. might not be important you would get higher throughput by not blocking worker threads, making them available to serve other requests in parallel. Without the need for any buffering.

Consider also that buffering sometimes might not be an option because you have to make sure an event is actually written (e.g. audit logs).

For cases where the latency is important you could still provider buffering, maybe as part of the framework itself, so that sinks don't have to care.

Now of course I understand that other cases where async is not needed should not pay the performance penalty of this. I believe however that it is possible to avoid most of this penalty by careful use of async within the framework itself.

If you take a look at the code I provided, you will see that an explicit async/await is only used in 3 places at the moment. Most of the other methods can simply return the Task created initially by the sink. For sinks that don't use IO directly, this can be the Task.Completed constant, so that no allocations have to be done. There are still some problems with things like the AggregateSink which might require this. But I think with careful engineering even this can be avoided except for cases where absolutely needed.

When I initially started, I used async/await extensively and the slowdown I experienced was sometimes as high as 70x using the Benchmarks included in Serilog (of course taking into consideration that we are still talking about nanoseconds here and any IO can easily be in the millisecond area). However using the Task based approach the slowdown is now in the area of 2-4x and I think it can go down even further.

Example for the LevelControlBenchmark for switch Off or LevelSwitchOff are approximately 4ns vs 8ns on my PC. This slowdown is probably caused by the intermediate SafeAggregateSink which still uses async/await for the error handling and aggregation of tasks.

aKzenT:
I updated the code to avoid async/await in the filter and aggregation sinks as well. Now, if there are no sinks that uses async in any way there are no additional allocations done in the method or anywhere else in the pipeline.

This is done by using the fact that a synchronous sink would always return a Task that is already completed. So we only do a Task.WhenAll if there is a sink that needs it.

To give you an idea how I would imagine this to added to Serilog in a compatible way would be to:
- Internally use Tasks in the library like demonstrated in the spike
- Support both an IAsyncLogEventSink and ILogEventSink for backward compatibility
- Write a simple Wrapper for ILogEventSink that implements IAsyncLogEventSink and simply returns Task.Completed at the end of Emit, so that the framework can work with both types of sinks without having to special case.

Now the difficult question would be what to do with the Logger-Interface. I see 3 options:
- Only provide an async API for logging. Clear disadvantage would be that it would break all existing code dependent on Serilog. Probably not a good idea.
- Provide both a synchronous and an asynchronous API for Logging, but require that to use the synchronous API you are not allowed to use asynchronous sinks (slowly moving users to the asynchronous API)
- Provide both a synchronous and an asynchronous API for Logging and do some magic to run the asynchronous version synchronously when needed (not recommended in my eyes, since it could lead to deadlocks if the asynchronous sinks are not carefully written)

What are your thoughts?

nblumhardt:
Thanks for the follow-up! I agree that there are many possible optimisations, and there are also a lot of interesting and subtle API/compatibility issues we'd need to consider (but are unlikely to be insurmountable).

What we still need to answer is - what is the gain? There's a lot of effort in discussing/building/testing/shipping/maintaining any feature, especially this kind. We really need to be able to weigh it up against specific benefits.

(I think the auditing example is interesting, but I don't think it would motivate a change like this. Since adding built-in reliable auditing support, no third-party sinks (to my knowledge) have taken advantage of it, so I don't think the audience for that one is huge. Happy to be proven wrong of course, just speculating here.)

To continue exploring this I think we really need a number: what's the increase in throughput, and for what scenario?

Cheers,
Nick

aKzenT:
Hey,

I would say the reasoning is the same as for all other async methods in the framework. Whenever a call can potentially block a thread, it is better to return a Task so that the thread is free to do other work while the call is being finished.

Now as for a concrete scenario with Serilog that I face in my company, which I think is not so uncommon: We write log events to an Azure Queue (of course would be the same for Entity Framework, Application Insights, ...) within a web application. The Microsoft provided API in this case does not even provide a synchronous way to write to the queue, so we have to explicitly wait in the Serlilog sink. In any case we are blocking the thread from doing any other work like serving other users.

Using background tasks in web applications is always a pain in the ass and risky as it is easy to loose events, when the app restarts, .etc., so buffering is also not a good option for us.

Now I can try to create some sort of sample that shows the increase in throughput in a scenario like this, but like I said at the end the argument is the same as for all other APIs. Microsoft spent a lof of effort to make the entire pipeline for MVC async capable exactly for increasing the throughtput and reducing the number of threads required to serve requests.

nblumhardt:
Thanks for the reply.

The event-at-a-time logging scenario with the queue is interesting and worth noting; the proportion of Serilog use with this model is low today, but if that style increases then optimising it with async would be on the table.

Getting clear on reasons _is_ a really important first step for any new feature. Just adding `Task` everywhere without measurements/scenarios would be dubious. As mentioned above, since Serilog offloads network access to background batching in most cases today, and page caching will be at play in others, the theoretical benefit isn't a real one in the common usage patterns we have at present.

I don't think that means we're against the feature, but the cost/benefit still doesn't seem convincingly in favour of doing it.

> Now I can try to create some sort of sample that shows the increase in throughput in a scenario like this, but like I said at the end the argument is the same as for all other APIs.

It's a fair assumption, and if the change resulted in no negative impact for other scenarios that would be fair, but since it _does_ have an impact on regular non-async logging (and on the "off" scenario) we'd absolutely need numbers for this. Perf work is always full of surprises.

> Microsoft spent a lof of effort to make the entire pipeline for MVC async capable

Except `ILogger` :-)

https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/ILogger.cs

aKzenT:
Hm, I think you make very good points. The reference to the Microsoft Logger implementation made me look a bit further and I found that they discussed async as well and rejected it. See: https://github.com/aspnet/Logging/issues/444

What I learn from all this is that the current recommended approach is for logging is that the actual sinks should usually do their IO in a background worker or tasks, like you also said is the approach taken by many of the current logging sinks.

So maybe additional effort should be focussed more on providing an easy and reliable way for sinks to implement an asynchronous writing of events without having to reinvent the wheel every time.

nblumhardt:
@aKzenT thanks for the follow-up and the pointer to the other discussion.

Sinks today don't generally reinvent the wheel here; there's _Serilog.Sinks.PeriodicBatching_ to base batched sinks on, and _Serilog.Sinks.Async_ to wrap event-at-a-time calls. There's room for improvement and unification of these, but they serve us pretty well.

Thanks again for the input and PR!

mattwoberts:
Hi, just landed here, since I'm in the same boat as @aKzenT - I've got a high-load asp.net core web app. I'm using this sink to write to azure table storage: https://github.com/serilog/serilog-sinks-azuretablestorage

Now - on investigating, I find that this third party sink is indeed blocking, because it has to wrap all the MS provided azure async API calls with a Wait - so we're "async over sync". Since it's not my sink, I need to spend some time checking if we're going to have problems

I just wanted to say thanks for the discussion thus far - it helped.

PR #955, opened 2017-03-27
Sink wrapping

johnduhart:
In our use of Serilog we'd like to be able to easily wrap existing sinks while also maintaining use of built-in extension methods. For example, we can create a decorator sink that adds the ability to disable sink output during runtime, and can be added to any existing sink.

This PR currently is a first pass on the API for this, so we can discuss the best approach for adding this in. The initial approach adds a new method to LoggerSinkConfiguration which accepts a callback to creating the wrapping Sink.

### DIFF
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -150,6 +150,21 @@ namespace Serilog.Configuration
         {
             if (logger == null) throw new ArgumentNullException(nameof(logger));
             return Sink(new SecondaryLoggerSink(logger, attemptDispose: false), restrictedToMinimumLevel);
-        } 
+        }
+
+        public static LoggerConfiguration Wrap(
+            LoggerSinkConfiguration parentSinkConfiguration,
+            Func<ILogEventSink, ILogEventSink> addSink,
+            Action<LoggerSinkConfiguration> sinkConfigAction)
+        {
+            var sinkConfiguration = new LoggerSinkConfiguration(
+                parentSinkConfiguration._loggerConfiguration,
+                s => parentSinkConfiguration.Sink(addSink(s)),
+                parentSinkConfiguration._applyInheritedConfiguration);
+
+            sinkConfigAction(sinkConfiguration);
+
+            return parentSinkConfiguration._loggerConfiguration;
+        }
     }
 }
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -169,9 +169,23 @@ namespace Serilog.Configuration
             if (wrapSink == null) throw new ArgumentNullException(nameof(wrapSink));
             if (configureWrappedSink == null) throw new ArgumentNullException(nameof(configureWrappedSink));
 
+            void WrapAndAddSink(ILogEventSink sink)
+            {
+                bool sinkIsDisposable = sink is IDisposable;
+
+                ILogEventSink wrappedSink = wrapSink(sink);
+
+                if (sinkIsDisposable && !(wrappedSink is IDisposable))
+                {
+                    SelfLog.WriteLine("Wrapping sink {0} does not implement IDisposable, but wrapped sink {1} does.", wrappedSink, sink);
+                }
+
+                loggerSinkConfiguration.Sink(wrappedSink);
+            }
+
             var capturingLoggerSinkConfiguration = new LoggerSinkConfiguration(
                 loggerSinkConfiguration._loggerConfiguration,
-                s => loggerSinkConfiguration.Sink(wrapSink(s)),
+                WrapAndAddSink,
                 loggerSinkConfiguration._applyInheritedConfiguration);
 
             configureWrappedSink(capturingLoggerSinkConfiguration);
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -152,19 +152,27 @@ namespace Serilog.Configuration
             return Sink(new SecondaryLoggerSink(logger, attemptDispose: false), restrictedToMinimumLevel);
         }
 
+        /// <summary>
+        /// Helper method for wrapping sinks.
+        /// </summary>
+        /// <param name="logggerSinkConfiguration">The parent sink configuration.</param>
+        /// <param name="sinkDecorator">A function that allows for wrapping <see cref="ILogEventSink"/>s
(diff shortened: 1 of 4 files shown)


nblumhardt:
Hi John, thanks for the suggestion. Have you checked out https://github.com/serilog/serilog-sinks-async ? This is a wrapper along these lines.

I think the API there is okay (`WriteTo.Async(w => ...`) - it would be great to see whether we could improve on the way it's implemented: https://github.com/serilog/serilog-sinks-async/blob/dev/src/Serilog.Sinks.Async/LoggerConfigurationAsyncExtensions.cs - thoughts?

johnduhart:
Hi Nick, wasn't aware of the async sink before, but I agree it's close to what I'm proposing here. My main concern with the implementation of `LoggerConfigurationAsyncExtensions` is that it requires a new logger pipeline to be instantiated. I'm not familiar with what performance impact that will have, but I would assume it's less expensive to decorate the sink itself.

With an API that makes defining wrapped sinks possible, the implementation of `WriteTo.Async` could be improved while retaining its existing API.

nblumhardt:
Thanks for the follow-up. Yes, simplifying the implementation of `WriteTo.Async()` was what I had in mind 👍 

`WrapSink()` is a helper method in this context; having it public on `LoggerSinkConfiguration` might add clutter that we can possibly avoid. I wonder if there's another way to expose this?

It's a little bit convoluted, but another option would be to make `Wrap` static:

```csharp
return LoggerSinkConfiguration.Wrap(loggerSinkConfiguration, s => new BackgroundWorkerSink(s, bufferSize), configure);
```

I.e. by making it static and explicitly passing `this`, we'd avoid anything showing up after `WriteTo.`. Thoughts?

johnduhart:
I totally agree, the extra method added to `WriteTo.` was one of the ugly parts of my original proposal. Making it a static method totally works around that though!

I'll update the PR with this approach.

softwaretecture:
To be honest, I'm slightly confused about a method named "Async". It sort of mixes the "async" concept with configuration of the logger. During configuration we don't really activate any I/O bound behaviors, so calling it Async seems very odd to me.

nblumhardt:
@johnduhart great, I'll take another look through ASAP :+1:

@softwaretecture thanks for the feedback! I'd rather not go into too much depth on this one here, since we'll derail the thread. I think there might be some discussion in an old ticket on the https://github.com/serilog/serilog-sinks-async repository. Naming things is hard, of course :-) Cheers!

johnduhart:
Taking the WIP tag off this, and added documentation to the method.

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -150,6 +150,29 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
         {
             if (logger == null) throw new ArgumentNullException(nameof(logger));
             return Sink(new SecondaryLoggerSink(logger, attemptDispose: false), restrictedToMinimumLevel);
-        } 
+        }
+
+        /// <summary>
+        /// Helper method for wrapping sinks.
+        /// </summary>
+        /// <param name="logggerSinkConfiguration">The parent sink configuration.</param>
```
nblumhardt: `ggg`

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -150,6 +150,29 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
...
+        /// <param name="sinkDecorator">A function that allows for wrapping <see cref="ILogEventSink"/>s
```
nblumhardt: Since it's an action, and the API uses the term _wrap_ rather than _decorate_, could we use `wrapSink`? Also missing a period/full-stop for this sentence.

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -150,6 +150,29 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
...
+        /// added in <paramref name="sinkConfigurationAction"/></param>
+        /// <param name="sinkConfigurationAction">An action that configures sinks to be wrapped in <paramref name="sinkDecorator"/></param>
+        /// <returns>Configuration object allowing method chaining.</returns>
+        public static LoggerConfiguration Wrap(
+            LoggerSinkConfiguration logggerSinkConfiguration,
+            Func<ILogEventSink, ILogEventSink> sinkDecorator,
+            Action<LoggerSinkConfiguration> sinkConfigurationAction)
```
nblumhardt: Ditto here, could be `configureWrappedSink`.

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -150,6 +150,29 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
...
+        {
+            var capturingLoggerSinkConfiguration = new LoggerSinkConfiguration(
```
nblumhardt: Missing argument null checks.

test/Serilog.Tests/LoggerConfigurationTests.cs
```diff
@@ -577,5 +578,19 @@ public string Property
                 get { throw new Exception("Boom!"); }
             }
         }
+
+        [Fact]
+        public void WrappedSinks()
```
nblumhardt: `WrappingDecoratesTheConfiguredSink()`?

test/TestDummies/DummyLoggerConfigurationExtensions.cs
```diff
@@ -42,5 +41,15 @@ public static LoggerConfiguration WithDummyThreadId(this LoggerEnrichmentConfigu
         {
             return loggerSinkConfiguration.Sink(new DummyRollingFileAuditSink(), restrictedToMinimumLevel);
         }
+
+        public static LoggerConfiguration DummyWrapping(
```
nblumhardt: `WriteTo.Dummy()`? :-)

nblumhardt:
Looks great - just a  bunch of nitpicks from me.

There's one further twist we might handle, though I'm not sure if there's a clean way. It's a little to easy to accidentally create a wrapper that doesn't implement `IDisposable` (and forward the `Dispose()` call to the wrapped sink).

Could we check dynamically to enforce that the wrapped sink (returned from the decorator function) implements `IDisposable` and write a warning to `SelfLog` if it does not?

johnduhart:
@nblumhardt Thanks, I'm terrible at writing method documentation.

That's an interesting point regarding IDisposable sinks. I can definitely add support to check for that.

nblumhardt:
Love it 👍

Issue #958, opened 2017-04-06
Output template processing improvements


nblumhardt:
#944 adds new functionality to output templates, in the form of `{Properties}` support.

_Serilog.Sinks.Literate_, the primary external consumer of `OutputProperties.GetOutputProperties()`, can't make use of the new functionality directly - in order to pretty-print the property object it needs to reimplement a lot of the logic there.

This leaves us at a good point to make some changes:

 - [x] Break the literate console sink's dependency on `OutputProperties.GetOutputProperties()` entirely, copying some implementation across if we have to
 - [x] Make the newly-added `GetOutputProperties()` overload internal
 - [x] Mark the old `GetOutputProperties()` method obsolete
 - [x] Rewrite `MessageTemplateTextFormatter` to avoid all of the unnecessary allocations that it currently makes via `GetOutputProperties()`; we can consider #956 in the process

nblumhardt:
This shipped in 2.5.

PR #960, opened 2017-04-07
Output template formatting harness

nblumhardt:
Still tweaking this, but will support the work in #958


nblumhardt:
Okay, think this one is as ready as it needs to be for now. On battery power, so can't get meaningful numbers.

Needs a squash-and-merge :-)

Issue #965, opened 2017-04-17
Support for c# 7 value tuple destructuring


jotabe-net:
If you try to destructure a C# 7 Value Tuple, by using `{@tupleValue}` it doesn't destructure it.

For example, when this code is run
```
var x =1;
var y = 7;
var point = (x,y);

Log.Information("The point is {@point}", point);
```

You get this message logged:
```The point is ValueTuple`2```
and this property:
```point   {_typeTag: "ValueTuple`2"}```

I tried to implement an `IDestructuringPolicy`, but it wasn't even invoked. I suppose that's because it's a struct (value type) instead of an object (reference type)

However I think it would be interesting to support this destructuring as tuples will become widely used sooner or later.

These are some clues for the possible implementation:
- it's a value type instead of a reference type
- it has fields, instead of properties
- the fileds have the names: `Item1`, `Item2` and so on
- it is supposed to automatically receive a `TupleElementNamesAttribute` with the friendly names of the fileds *in a pre-order depth-first traversal of a type's construction*, as stated in [this atrribute 's documentation](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.tupleelementnamesattribute.tupleelementnamesattribute(v=vs.110).aspx#Anchor_1) I tried to check this in the debugger, but the debugger is failing when dealing with this kind of type and Reflection.

In the meantime, you can solve this problem by manually converting it to an anonymus type, like so:
```
Log.Information("The point is {@point}", new { point.x, point.y});
```

nblumhardt:
Thanks for the note. Yes, this looks like a worthwhile improvement to explore.

I guess there are these implementation alternatives:

 1. Only support it for target frameworks that include `ValueTuple` natively (not sure yet which these are)
 2. Support it everywhere, by depending on the NuGet package carrying the `ValueTuple` type
 3. Support it everywhere, with an add-on package
 4. Support it everywhere, using reflection only (no direct dependency on `ValueTuple` itself)

It seems like a combination of 1 & 3 could work, i.e. use `#if VALUETUPLE` to conditionally support it in Serilog itself, and distribute a NuGet package version that would need to be activated via `Destructure.ValueTuples()` for use on other TFMs.

Anyone keen to investigate?

nblumhardt:
I've been looking into this some more; I think the best representation in Serilog for tuples is, in fact, arrays...

[...]

PR #976, opened 2017-05-30
ValueTuple to sequence value conversion

nblumhardt:
This PR for #965 converts tuples like `(1, 2, 3)` into arrays like `[1, 2, 3]` for consumption by sinks.

I don't think tuples are important enough, currently, to justify the effort of including another `LogEventPropertyValue` type to represent them. Arrays are a bit easier on the eyes than structures, and we don't lose any information since tuple member names aren't available dynamically.

The `@` operator is not required to capture a tuple. E.g.:

```csharp
Log.Information("{T}", (1, 2));
```

is fine. Specifying `@` will trigger deep serialization in the case of tuple members that are complex types.

The PR is a little ham-fisted (the change doesn't fit in as an `IScalarConversionPolicy`, unfortunately, so special treatment is needed). Only tuples up to length 7 are supported.

Since tuples are here to stay, and supported everywhere Serilog runs, I've introduced the _System.ValueTuple_ dependency for the current targets. A .NET 4.7-specific target would avoid it, as will .NET Core 2.0. We can add this fresh round of targets once core 2.0 ships, I think (4.7 is a pain right now as it requires a targeting pack).

### DIFF
--- a/src/Serilog/Parameters/PropertyValueConverter.cs
+++ b/src/Serilog/Parameters/PropertyValueConverter.cs
@@ -74,7 +74,7 @@ namespace Serilog.Parameters
                 new SimpleScalarConversionPolicy(BuiltInScalarTypes.Concat(additionalScalarTypes)),
                 new NullableScalarConversionPolicy(),
                 new EnumScalarConversionPolicy(),
-                new ByteArrayScalarConversionPolicy(),
+                new ByteArrayScalarConversionPolicy()
             };
 
             _destructuringPolicies = additionalDestructuringPolicies
@@ -188,6 +188,33 @@ namespace Serilog.Parameters
 
                 return new SequenceValue(
                     enumerable.Cast<object>().Take(_maximumCollectionCount).Select(o => limiter.CreatePropertyValue(o, destructuring)));
+            }
+
+            if (value is IStructuralEquatable)
+            {
+                var type = value.GetType();
+                if (type.IsConstructedGenericType)
+                {
+                    var definition = type.GetGenericTypeDefinition();
+                    if (definition == typeof(ValueTuple<>) || definition == typeof(ValueTuple<,>) ||
+                        definition == typeof(ValueTuple<,,>) || definition == typeof(ValueTuple<,,,>) ||
+                        definition == typeof(ValueTuple<,,,,>) || definition == typeof(ValueTuple<,,,,,>) ||
+                        definition == typeof(ValueTuple<,,,,,,>)) // Ignore the 8+ value case for now.
+                    {
+                        var elements = new List<LogEventPropertyValue>();
+                        foreach (var field in type.GetTypeInfo().DeclaredFields)
+                        {
+                            if (field.IsPublic && !field.IsStatic)
+                            {
+                                var fieldValue = field.GetValue(value);
+                                var propertyValue = limiter.CreatePropertyValue(fieldValue, destructuring);
+                                elements.Add(propertyValue);
+                            }
+                        }
+
+                        return new SequenceValue(elements);
+                    }
+                }
             }
 
             if (destructuring == Destructuring.Destructure)
--- a/src/Serilog/Parameters/PropertyValueConverter.cs
+++ b/src/Serilog/Parameters/PropertyValueConverter.cs
@@ -233,10 +233,20 @@ namespace Serilog.Parameters
             }
 
             var definition = valueType.GetGenericTypeDefinition();
+
+            // Ignore the 8+ value case for now.
+#if VALUETUPLE
             if (definition == typeof(ValueTuple<>) || definition == typeof(ValueTuple<,>) ||
                 definition == typeof(ValueTuple<,,>) || definition == typeof(ValueTuple<,,,>) ||
                 definition == typeof(ValueTuple<,,,,>) || definition == typeof(ValueTuple<,,,,,>) ||
-                definition == typeof(ValueTuple<,,,,,,>)) // Ignore the 8+ value case for now.
+                definition == typeof(ValueTuple<,,,,,,>))
+#else
+            var defn = definition.FullName;
+            if (defn == "System.ValueTuple`1" || defn == "System.ValueTuple`2" ||
+                defn == "System.ValueTuple`3" || defn == "System.ValueTuple`4" ||
+                defn == "System.ValueTuple`5" || defn == "System.ValueTuple`6" ||
+                defn == "System.ValueTuple`7")
+#endif
             {
                 var elements = new List<LogEventPropertyValue>();
                 foreach (var field in valueType.GetTypeInfo().DeclaredFields)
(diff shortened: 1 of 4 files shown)


src/Serilog/Parameters/PropertyValueConverter.cs
```diff
@@ -190,6 +190,33 @@ LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructur
                     enumerable.Cast<object>().Take(_maximumCollectionCount).Select(o => limiter.CreatePropertyValue(o, destructuring)));
             }
 
+            if (value is IStructuralEquatable)
```
nblumhardt: Attempting to keep the cost of the test down.

src/Serilog/Parameters/PropertyValueConverter.cs
```diff
@@ -190,6 +190,33 @@ LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructur
...
+                if (type.IsConstructedGenericType)
+                {
+                    var definition = type.GetGenericTypeDefinition();
+                    if (definition == typeof(ValueTuple<>) || definition == typeof(ValueTuple<,>) ||
+                        definition == typeof(ValueTuple<,,>) || definition == typeof(ValueTuple<,,,>) ||
+                        definition == typeof(ValueTuple<,,,,>) || definition == typeof(ValueTuple<,,,,,>) ||
+                        definition == typeof(ValueTuple<,,,,,,>)) // Ignore the 8+ value case for now.
+                    {
+                        var elements = new List<LogEventPropertyValue>();
+                        foreach (var field in type.GetTypeInfo().DeclaredFields)
```
nblumhardt: Assumes fields are ordered (I think this is safe).

nblumhardt:
Ready to go, if CI likes me :-)

src/Serilog/Serilog.csproj
```diff
@@ -23,25 +23,44 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'net45' ">
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
+    <PackageReference Include="System.ValueTuple" Version="4.3.1" />
```
skomis-mm: It's a pity. Why you didn't considered the fourth option (along with the first)? 
> 4.Support it everywhere, using reflection only (no direct dependency on  ValueTuple  itself)

Is that because of some performance concerns? I believe you could check `type.Name` to constants (`ValueTyple'1`, `ValueTuple'2` etc) like you did with `definition ==` comparisons

src/Serilog/Serilog.csproj
```diff
@@ -23,25 +23,44 @@
...
-    <PackageReference Include="System.Dynamic.Runtime" Version="4.0.11" />
-    <PackageReference Include="System.Globalization" Version="4.0.11" />
-    <PackageReference Include="System.Linq" Version="4.1.0" />
-    <PackageReference Include="System.Reflection" Version="4.1.0" />
-    <PackageReference Include="System.Reflection.Extensions" Version="4.0.1" />
-    <PackageReference Include="System.Runtime" Version="4.1.0" />
-    <PackageReference Include="System.Runtime.Extensions" Version="4.1.0" />
-    <PackageReference Include="System.Text.RegularExpressions" Version="4.1.0" />
-    <PackageReference Include="System.Threading" Version="4.0.11" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.3.0" />
```
khellang: The `netstandard1.x` targets should probably be updated to just depend on `NETStandard.Library` anyway, TBH.

src/Serilog/Serilog.csproj
nblumhardt: Thanks for the suggestion. I initially discounted this because it seemed a bit awkward, given tuples are a feature of C# that's here to stay. On second thought, it's not much worse to do the type name check than to do what we have here, so I'll switch it back over and `#if VALUETUPLE` conditionally-include the `Type`-based version 👍

src/Serilog/Serilog.csproj
nblumhardt: Thanks @khellang - you're probably right here. I guess the dust has settled enough for us to embrace the new order of things :-). Perhaps we make the change when .NET Core 2.0 comes along? I am tempted to leave it alone in this PR, just to avoid any more chasing of corner cases....

nblumhardt:
Alrighty! Avoided the package dependency and thus any dependency changes. Ready to go 🚀

test/Serilog.Tests/Parameters/PropertyValueConverterTests.cs
```diff
@@ -253,6 +255,55 @@ public void CSharpAnonymousTypesAreRecognizedWhenDestructuring()
...
+                (1, 2, 3, 4, 5, 6),
+                (1, 2, 3, 4, 5, 6, 7)
+            };
+
+            foreach (var t in tuples)
+                Assert.IsType<SequenceValue>(_converter.CreatePropertyValue(t));
+        }
+
+        [Fact]
+        public void EightPlusValueTupleElementsAreIgnored()
```
merbla: Nit `EightPlusValueTupleElementsAreIgnored` -> `EightPlusValueTupleElementsAreIgnoredWhenDestructuring`

Why the arbitrary 8?

test/Serilog.Tests/Parameters/PropertyValueConverterTests.cs
```diff
@@ -253,6 +255,55 @@ public void CSharpAnonymousTypesAreRecognizedWhenDestructuring()
...
+            Assert.Equal(new ScalarValue("A"), sequenceValue.Elements[1]);
+            var nested = Assert.IsType<SequenceValue>(sequenceValue.Elements[2]);
+            Assert.Equal(1, nested.Elements.Count);
+            Assert.Equal(new ScalarValue("B"), nested.Elements[0]);
+        }
+
+        [Fact]
+        public void AllTupleLengthsUpToSevenAreSupported()
+        {
+            var tuples = new object[]
```
merbla: Is an empty array test worthwhile?

merbla: Should we migrate to `InlineData` for use cases?

nblumhardt: Tried, but `ValueTuple` isn't regarded by csc as a compile-time constant, so no joy :-(

test/Serilog.Tests/Parameters/PropertyValueConverterTests.cs
nblumhardt: 👍 probably should be `WhenCapturing`, but good point.

The limitation is in how `ValueTuple` works, up to seven elements can be represented directly in one tuple, but to go 8+ they need to be nested. No major issue with implementing it down the road.

Issue #956, opened 2017-04-02
Allow "string quoting" to be overridden at the output template level


nblumhardt:
From: https://github.com/serilog/serilog-extensions-logging-file/issues/7

When Serilog renders data, it uses a JSON-like representation, so strings end up `"like this"` while numbers are `123` and arrays `[1, 2, 3]` etc.

_(Personal note - I can see the downside for some, but I still love this feature as it makes clear that there's more type information behind the scenes, when you can't use a sink with structured data support.)_

In some cases, especially when trying to match legacy file output, it'd be nice to switch this off and just pretend we're a fancy `string.Format()`. Opting out at the level of individual tokens (`{0:l}`) doesn't really achieve what's desired here. It'd be nicer to override the rendering at a higher level.

One option for achieving this would be to modify `MessageTemplateTextFormatter`'s handling of `{Message}` so that the `:l` is applied to the tokens within the message; e.g.:

```csharp
"{Timestamp:o} {RequestId,13} [{Level:u3}] {Message:l} ({EventId:x8}){NewLine}{Exception}"
```

Implementation strategy needs some investigation, but doesn't seem like it will be too complicated.

nblumhardt:
This shipped in 2.5.

Issue #839, opened 2016-08-30
Destructure -> Capture? (Discussion)


nblumhardt:
We might benefit from a bit of a terminology clean-up; dragging along the term "destructuring" in the API makes the library more confusing than it needs to be.
- `@` - _structure-capturing operator_
- `ICapturingPolicy` and `TryCapture()`
- `Capture.AsScalar()`, `Capture.ByTransforming()`
- `Capturing.Default`, `Capturing.Stringify`, `Capturing.Structured`

I think this can be done without too many breaking changes: the old interfaces and types can stick around, and where they are plugged in, we can adapt them to the newer types/names. E.g.:

``` csharp
    .Destructure.With<T>()
```

would, under the hood, create the `IDestructuringPolicy` and wrap it in an `ICapturingPolicy` adapter and delegate to `Capture.With(wrapper)`.

The old types and members could be marked obsolete and non-browsable for the next couple of major releases.

The most painful-looking changes such as the parameter rename on `ForContext` from `destructureObjects` to `captureStructure` would be source-breaking changes, but not binary-breaking ones.

What's everyone think? Feasible? Worth some inevitable headaches? This would be a 3.0 kind of change, and only something we'd do if other subtle breaking changes come into 3.0, for example I am working on a performance-oriented changeset that slightly modifies `TextToken` and `PropertyToken` constructors. "2017" kinds of things :-)

adamchester:
Totally agree that your proposed terminology is much better.

Totally unsure if it's worth the pain :smile: 

My vote is :+1:

tsimbalar:
Maybe it's just me, but I kinda liked *destructuring*. Even if the terminology is not great, it is kind of consistent with terms like *structured logging*. 

Also, while "structure capturing" is probably better semantically, it's kind of a mouthful :) , and using "capturing" alone is kind of vague .

Totally subjective opinion, of course ;)

PR #977, opened 2017-05-31
Output and rendering improvements

nblumhardt:
Bundling a whole range of output and template rendering improvements for #958 here.

We won't be able to remove all of the obsolete code until the literate console sink has been weaned off of `OutputProperties.GetOutputProperties()`, but when that's done we should be able to clear out a lot of gunk here.

Currently just making the first few moves, will streamline things a bit before this is "done".


nblumhardt:
Oh, and I'll fix the tests ;-)

nblumhardt:
Now includes #958, #956 and the optional JSON output style suggested in https://github.com/serilog/serilog-sinks-console/issues/7.

This is, unfortunately, a bit of a monster changeset. Digging in to output formatting unearthed a lot of code that needed cleanup - I ended up fixing a few things like unused `using`s globally to get us back to "ReSharper Green".

Here are the main points:

 * Output templates can now include one or both of two optional formats, e.g.
  - `{Message:l}` will omit `"quotes"` and not perform any escaping on strings embedded in the message
  - `{Message:j}` will use JSON, rather than Serilog's friendly formatting, for embedded data (this includes numbers, structures, arrays and so-on)
  - `{Message:lj}` is as for JSON, but with unquoted/unescaped string values
 * Output formatting now makes many less allocations. Benchmarks coming shortly.
 * `OutputProperties.GetOutputProperties()` and the original allocatey output formatting types are still around for API compatibility but marked `[Obsolete]`; it's probably only `LiterateConsoleSink` really depends on them, can remove in a future major release.
 * Rendering of message templates is now separate from the `MessageTemplate` class; in general, the two places where our data structures implement their own rendering (`MessageTemplate` and `LogEventPropertyValue`) are at odds with the need for multiple different styles.
   - The new `MessageTemplateRenderer` can be used as the basis for some of the console sink improvements, where coloring etc. make other reuse methods awkward.
   - Planning to take this further and move `LogEventPropertyValue` rendering into a visitor implementation at a later date.
   - For now, I've left the public API surface the same.
 * Obsoleted `RawFormatter` - discovered we have no tests for it, and I'm not sure there's any real need for it.
 * In light of the terminology we tend to use these days I've moved some internal types into `Serilog.Capturing` and `Serilog.Rendering` namespaces; #839 might take this further.
 * `Padding.Apply()` no longer allocates for padding <= 80 chars.

Now to bench... :-)

nblumhardt:
Having trouble with the benchmark runner, will need to come back to this when it's sorted out :-(

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
```diff
@@ -58,42 +61,70 @@ public void Format(LogEvent logEvent, TextWriter output)
             if (logEvent == null) throw new ArgumentNullException(nameof(logEvent));
             if (output == null) throw new ArgumentNullException(nameof(output));
 
-            // This could be lazier: the output properties include
-            // everything from the log event, but often we won't need any more than
-            // just the standard timestamp/message etc.
-            var outputProperties = OutputProperties.GetOutputProperties(logEvent, _outputTemplate);
```
nblumhardt: The changes in this method are the meat of the "less allocations" change - `GetOutputProperties()` above would previously allocate objects for `{Level}`, `{Message}`, `{Exception}` etc. Instead now we just iterate and examine property names, and write the output inline.

src/Serilog/Rendering/MessageTemplateRenderer.cs
```diff
@@ -0,0 +1,111 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Serilog.Events;
+using Serilog.Formatting.Json;
+using Serilog.Parsing;
+
+namespace Serilog.Rendering
+{
+    static class MessageTemplateRenderer
```
nblumhardt: This class is the remainder - formatting of message templates is externalized from the `MessageTemplate` and related token classes here.

I expect there will be a small perf hit because dispatch is done by switching on type, rather than by virtual dispatch.

johnduhart:
After a quick look, it seems that a few class namespaces have changed. Is this considered a breaking change?

nblumhardt:
Thanks for checking it out, @johnduhart. It would be if the namespaces contained public types, but these are all internal types so there shouldn't be externally visible effects this time around.

...unless I've overlooked something public in there, of course :-)

johnduhart:
Oh, right, I guess I overlooked that :sweat_smile:

nblumhardt:
I am really itching to clean up some of the public type names and namespaces - they've been frozen for compatibility reasons for three years or so now, and a few like #839 would be worthwhile improvements. Someday!

merbla:
@nblumhardt can you link to your Resharper profile/settings?  Just want to be on the same page.

nblumhardt:
@merbla I just exported them in their entirety and included the dotsettings file in this PR: https://github.com/serilog/serilog/pull/977/files#diff-a08cec6edf37e99e02d322888266d97d

:-)

merbla:
@nblumhardt a bit of love is needed post #976

nblumhardt:
### OutputTemplateRenderingBenchmark

Before:

 |         Method |      Mean |    StdDev |
 |--------------- |---------- |---------- |
 | FormatToOutput | 2.5465 us | 0.0497 us |

After:

 |         Method |      Mean |    StdDev |
 |--------------- |---------- |---------- |
 | FormatToOutput | 1.4546 us | 0.0188 us |

This doesn't take the garbage reduction into account, so I think it's nice to get a ~40% perf boost along the way.

### MessageTemplateRenderingBenchmark

Before:

 |                       Method |        Mean |    StdDev |
 |----------------------------- |------------ |---------- |
 |     TemplateWithNoProperties |          NA |        NA |
 | TemplateWithVariedProperties | 524.7727 ns | 9.0085 ns |

After:

 |                       Method |        Mean |    StdDev |
 |----------------------------- |------------ |---------- |
 |     TemplateWithNoProperties |          NA |        NA |
 | TemplateWithVariedProperties | 576.5388 ns | 9.5544 ns |

So, the switch from virtual to type-driven dispatch looks like it will cost about 10%. Not awesome, but also not too worrying.

The broken benchmark turned up some problems with this one in general, namely, the code intended to reset the output `StringWriter` before each test was only being run once. Having a look at the implications now.

nblumhardt:
Fixed the message template rendering benchmark; this is the "after":

 |                       Method |       Mean |     Error |    StdDev |
 |----------------------------- |-----------:|----------:|----------:|
 |     TemplateWithNoProperties |   6.773 ns | 0.0551 ns | 0.0488 ns |
 | TemplateWithVariedProperties | 408.450 ns | 1.8831 ns | 1.7614 ns |

Knocks ~25% off the overall time, meaning the differences in the earlier run are probably understated by about that much.

BDN 0.10.6 doesn't get the .NET Framework target running successfully either, unfortunately.

src/Serilog/Rendering/Padding.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (alignment.Value.Direction == AlignmentDirection.Left)
+                output.Write(value);
+
+            if (pad <= PaddingChars.Length)
+            {
+                output.Write(PaddingChars, 0, pad);
+            }
+            else
+            {
+                output.Write(new string(' ', pad));
```
nblumhardt: We now only allocate if the padding applied to a token is > 80 chars. (We allocated for every render, previously.)

nblumhardt:
I think this'd benefit from a look over, now. Might be some future improvements to make, but I'm happy that it makes enough progress to be a worthwhile merge.

nblumhardt:
Ran these again, with the benchmark changes backported to my local `dev` branch for the "before" numbers.

``` ini
BenchmarkDotNet=v0.10.6, OS=Windows 10 Redstone 2 (10.0.15063)
Processor=Intel Core i7-7700HQ CPU 2.80GHz (Kaby Lake), ProcessorCount=8
Frequency=2742186 Hz, Resolution=364.6726 ns, Timer=TSC
dotnet cli version=1.0.4
  [Host]     : .NET Core 4.6.25211.01, 64bit RyuJIT
  DefaultJob : .NET Core 4.6.25211.01, 64bit RyuJIT
```

### OutputTemplateRenderingBenchmark

Before:

 |         Method |     Mean |     Error |    StdDev |  Gen 0 | Allocated |
 |--------------- |---------:|----------:|----------:|-------:|----------:|
 | FormatToOutput | 2.438 us | 0.0189 us | 0.0167 us | 0.3662 |   1.13 KB |

After:

 |         Method |     Mean |     Error |    StdDev |   Gen 0 | Allocated |
 |--------------- |---------:|----------:|----------:|---------:|----------:|
 | FormatToOutput | 1.433 us | 0.0344 us | 0.0992 us | 0.1545 |     488 B |

### MessageTemplateRenderingBenchmark

Before:

 |                       Method |       Mean |     Error |    StdDev |  Gen 0 | Allocated |
 |----------------------------- |-----------:|----------:|----------:|-------:|----------:|
 |     TemplateWithNoProperties |   5.232 ns | 0.0798 ns | 0.0708 ns |      - |       0 B |
 | TemplateWithVariedProperties | 406.996 ns | 3.5597 ns | 3.3298 ns | 0.0300 |      96 B |

After:

 |                       Method |       Mean |     Error |    StdDev |  Gen 0 | Allocated |
 |----------------------------- |-----------:|----------:|----------:|-------:|----------:|
 |     TemplateWithNoProperties |   8.287 ns | 0.2085 ns | 0.2401 ns |      - |       0 B |
 | TemplateWithVariedProperties | 437.724 ns | 2.9852 ns | 2.7924 ns | 0.0300 |      96 B |

I'm not unhappy with these, but I don't think `TemplateWithNoProperties` should show such a difference. Trying a couple of other tweaks.

nblumhardt:
Alright, figured out the regression :-) This is the "after" now:

```
 |                       Method |       Mean |     Error |    StdDev |  Gen 0 | Allocated |
 |----------------------------- |-----------:|----------:|----------:|-------:|----------:|
 |     TemplateWithNoProperties |   5.591 ns | 0.1466 ns | 0.1907 ns |      - |       0 B |
 | TemplateWithVariedProperties | 418.376 ns | 2.6112 ns | 2.3148 ns | 0.0300 |      96 B |
```

The additional cost was a couple of argument checks. The numbers are so small I think it probably makes sense to leave (at least some of) the extra checks in. I'll see what I can come up with.

nblumhardt:
|                       Method |       Mean |     Error |     StdDev |  Gen 0 | Allocated |
 |----------------------------- |-----------:|----------:|-----------:|-------:|----------:|
 |     TemplateWithNoProperties |   5.934 ns | 0.0778 ns |  0.0690 ns |      - |       0 B |
 | TemplateWithVariedProperties | 411.162 ns | 8.2177 ns | 13.0341 ns | 0.0300 |      96 B |

Ready to go.

There's **one further sneaky change here** - the classes `PropertyToken` and `TextToken` are now `sealed`. This is technically a breaking change, but I don't believe there are any valid scenarios for deriving from these, and I've never seen it done in the wild. There are two reasons for sealing - first, it makes `Render()` on these types potentially statically dispatchable through a subclass reference, but also, it makes clear that these aren't designed for inheritance anyway.

nblumhardt:
Is there a chance anyone has some time to look over this one? :-)

src/Serilog/Formatting/Display/OutputProperties.cs
```diff
@@ -62,7 +65,7 @@ public static class OutputProperties
         /// </summary>
         /// <param name="logEvent">The log event.</param>
         /// <returns>A dictionary with properties representing the log event.</returns>
-        [Obsolete("Pass the full output template using the other overload.")]
+        [Obsolete("These implementation details of output formatting will not be exposed in a future version.")]
         public static IReadOnlyDictionary<string, LogEventPropertyValue> GetOutputProperties(LogEvent logEvent)
```
merbla: Just confirming this is the main consumer we are worried about? https://github.com/serilog/serilog-sinks-literate/blob/dev/src/Serilog.Sinks.Literate/Sinks/Literate/LiterateConsoleSink.cs#L88

src/Serilog/Formatting/Display/OutputProperties.cs
```diff
@@ -74,7 +77,7 @@ public static class OutputProperties
         /// <param name="logEvent">The log event.</param>
         /// <param name="outputTemplate">The output template.</param>
         /// <returns>A dictionary with properties representing the log event.</returns>
-        public static IReadOnlyDictionary<string, LogEventPropertyValue> GetOutputProperties(LogEvent logEvent, MessageTemplate outputTemplate)
+        internal static IReadOnlyDictionary<string, LogEventPropertyValue> GetOutputProperties(LogEvent logEvent, MessageTemplate outputTemplate)
```
merbla: Do we call this out as a breaking change in 2.5?

src/Serilog/Serilog.csproj
```diff
@@ -18,6 +18,8 @@
     <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
     <!-- Don't reference the full NETStandard.Library -->
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
+    <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
```
merbla: GOLD :1st_place_medal:

src/Serilog/Parsing/TextToken.cs
```diff
@@ -16,13 +16,14 @@
...
 using Serilog.Events;
+using Serilog.Rendering;
 
 namespace Serilog.Parsing
 {
     /// <summary>
     /// A message template token representing literal text.
     /// </summary>
-    public class TextToken : MessageTemplateToken
+    public sealed class TextToken : MessageTemplateToken
```
merbla: Another note for 2.5 release.

merbla:
@nblumhardt all good from me, just some questions to make sure these changes are noted in release.

merbla:
Lets get this to `dev` to measure the impact on sinks

src/Serilog/Formatting/Display/OutputProperties.cs
nblumhardt: I think it would make sense to list it prominently. Super surprised if it breaks anything out there today, though.

src/Serilog/Formatting/Display/OutputProperties.cs
nblumhardt: That's the one 👍

nblumhardt:
Thanks for the review @merbla 👍

Issue #993, opened 2017-07-01
Release build error on UWP: Cannot generate zero initialization code for type 'System.String'.


JasperE84:
After having upgraded to Serilog 2.5.0, my release builds (with .net optimalization) fail with the following error (VS2017).

3>C:\Users\myuser\.nuget\packages\microsoft.net.native.compiler\1.6.2\tools\Microsoft.NetNative.targets(691,5): error : Internal compiler error: Internal compiler error: Cannot generate zero initialization code for type 'System.String'.

After downgrading to 2.4.0 everything is fine again.

skomis-mm:
.net native toolchain doesn't like [this](https://github.com/serilog/serilog/blob/dev/src/Serilog/Rendering/Padding.cs#L22) line. Not sure what's wrong with it according to the [official guide](https://docs.microsoft.com/en-us/dotnet/framework/net-native/migrating-your-windows-store-app-to-net-native), but seems like a compiler issue. I wonder if CI can run this toolchain as a part of the build process.

nblumhardt:
We've held off adding a specific UWP target to the build, because of the slightly annoying build-time tooling dependency, but if these kinds of issues are present it seems worthwhile 👍 

I'm not sure what kind of workaround might get us past the compiler bug in the meantime. I don't suppose there are any VS updates/tooling patches available for you to install @JasperE84 ?

JasperE84:
I'm using the latest tooling updates (microsoft.net.native.compiler v1.6.2)

nblumhardt:
@JasperE84 https://www.nuget.org/packages/Serilog/2.5.1-dev-00869 includes the workaround from @skomis-mm - any chance you can test this package version and confirm the issue fixed? Thanks!

JasperE84:
@nblumhardt  I can confirm a succesful build with toolchain enabled. Thanks @nblumhardt  & @skomis-mm !

MichalStrehovsky:
.NET Native dev here - yes, this was a compiler issue. Great job root causing it to a single line! Coincidentally, we just fixed this very recently and the fix will ship with the next major release of .NET Native.

Here's another possible workaround: https://github.com/dotnet/corefx/blob/32c4013a5d750c1c6a5b240d37a689991a479097/src/System.Private.Xml/src/System/Xml/Core/XmlTextWriter.cs#L145

nblumhardt:
Great, thanks for the confirmation @MichalStrehovsky 👍

[...]

PR #996, opened 2017-07-03
Enable .NET Native on AppVeyor + workaround on compiler error for #993

skomis-mm:
Basically adds an empty UWP project (SDK = [10.0.10586](https://www.appveyor.com/docs/build-environment/#visual-studio-2017)) to the `test` folder and references `Serilog` to activate .NET Native on build. I didn't include this project to the solution to avoid tooling dependency. For the same reason build script was moved to a separate file. Not sure if its correct though.


skomis-mm:
First pass reported incorrect `success`, needs some tweaking

skomis-mm:
second pass reported correct [fail](https://ci.appveyor.com/project/serilog/serilog/build/865). Now let's try the workaround for [#993](https://github.com/serilog/serilog/issues/993)

nblumhardt:
LGTM; just one nit - could we call the UWP project _Serilog.UwpTest_, (_UwpTest**s**_?) or, put it under _sample/UwpDemo_? The name jumps out at me as "an implementation of Serilog for UWP", otherwise.

skomis-mm:
- renamed to Serilog.UwpTests to match other project names in that folder
- changed project type to `unit test project` so ideally tests can also be added and run on that runtime (for now its impossible because Visual Studio 2017 image runs on Windows Server 2016 on AppVeyor,)

PR #913, opened 2016-11-29
Include AssemblyInformationalVersion attribute to the with commit hash

skomis-mm:
- fixes [#784](https://github.com/serilog/serilog/issues/784), related ES'[#49](https://github.com/serilog/serilog-sinks-elasticsearch/issues/49)

By default `--version-suffix` of `dotnet-pack` command flows to `--version-suffix` of `dotnet-build` command.

 To overcome this `pr`'s propose  is to use `--no-build` switch on `dotnet-pack` so `--version-suffix` of `dotnet-pack` will affect only package/nuspec version. And `--version-suffix` of `dotnet-build` will only affect desired `AssemblyInformationalVersion`. 

The result is that
- when building on `master`  `AssemblyInformationalVersion` will look like `2.4.0-master-g3ff4af` and nuspec version will be **2.4.0**
- when building on `dev` `AssemblyInformationalVersion` will look like `2.4.0-dev-000042-g3ff4af` and nuspec version will be **2.4.0-dev-000042**

(see `AppVeyor`'s [artifacts](https://ci.appveyor.com/project/serilog/serilog/build/764/artifacts))

Also tested on `msbuild`-alpha tools - works as expected so this should work on a long run..

PR #1018, opened 2017-08-30
Include source in package builds

nblumhardt:
Can't see any reason not to use the build implicit in `dotnet pack`, please correct me if I'm missing anything.

### DIFF
--- a/Build.ps1
+++ b/Build.ps1
@@ -23,10 +23,12 @@ foreach ($src in ls src/*) {
 
 	echo "build: Packaging project in $src"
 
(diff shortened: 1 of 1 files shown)


merbla:
Makes sense

skomis-mm:
It was  for setting a custom `AssemblyInformationalVersion` while keeping nuget package version independent from it. See this [pr](https://github.com/serilog/serilog/pull/913)

nblumhardt:
Ah, thanks, fixing now 👍

nblumhardt:
...and, I see I even managed to push the branch to the wrong remote. Stepping away from the keyboard...

PR #1020, opened 2017-09-01
Support for LevelSwith in KeyValuePairSettings

tsimbalar:
In order to support serilog/serilog-settings-appsettings#7

### DIFF
--- a/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
+++ b/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
@@ -109,7 +109,7 @@ namespace Serilog.Settings.KeyValuePairs
             string minimumLevelControlledByLevelSwitchName;
             if (directives.TryGetValue(MinimumLevelControlledByDirective, out minimumLevelControlledByLevelSwitchName))
             {
-                var globalMinimumLevelSwitch = LookUpSwitchByNameOrThrow(minimumLevelControlledByLevelSwitchName, declaredLevelSwitches);
+                var globalMinimumLevelSwitch = LookUpSwitchByName(minimumLevelControlledByLevelSwitchName, declaredLevelSwitches);
                 loggerConfiguration.MinimumLevel.ControlledBy(globalMinimumLevelSwitch);
             }
 
@@ -125,7 +125,7 @@ namespace Serilog.Settings.KeyValuePairs
                 }
                 else
                 {
-                    var overrideSwitch = LookUpSwitchByNameOrThrow(minimumLevelOverrideDirective.Value, declaredLevelSwitches);
+                    var overrideSwitch = LookUpSwitchByName(minimumLevelOverrideDirective.Value, declaredLevelSwitches);
                     loggerConfiguration.MinimumLevel.Override(namespacePrefix, overrideSwitch);
                 }
             }
@@ -207,7 +207,7 @@ namespace Serilog.Settings.KeyValuePairs
             return new ReadOnlyDictionary<string, LoggingLevelSwitch>(namedSwitches);
         }
 
-        static LoggingLevelSwitch LookUpSwitchByNameOrThrow(string switchName, IReadOnlyDictionary<string, LoggingLevelSwitch> declaredLevelSwitches)
+        static LoggingLevelSwitch LookUpSwitchByName(string switchName, IReadOnlyDictionary<string, LoggingLevelSwitch> declaredLevelSwitches)
         {
             if (declaredLevelSwitches.TryGetValue(switchName, out var levelSwitch))
             {
@@ -221,7 +221,7 @@ namespace Serilog.Settings.KeyValuePairs
         {
             if (type == typeof(LoggingLevelSwitch))
             {
-                return LookUpSwitchByNameOrThrow(valueOrSwitchName, declaredSwitches);
+                return LookUpSwitchByName(valueOrSwitchName, declaredSwitches);
             }
             return SettingValueConversions.ConvertToType(valueOrSwitchName, type);
         }
--- a/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
+++ b/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
@@ -30,6 +30,7 @@ namespace Serilog.Settings.KeyValuePairs
         const string AuditToDirective = "audit-to";
         const string WriteToDirective = "write-to";
         const string MinimumLevelDirective = "minimum-level";
+        const string MinimumLevelControlledByDirective = "minimum-level:controlled-by";
         const string EnrichWithDirective = "enrich";
         const string EnrichWithPropertyDirective = "enrich:with-property";
         const string FilterDirective = "filter";
@@ -48,6 +49,7 @@ namespace Serilog.Settings.KeyValuePairs
             AuditToDirective,
             WriteToDirective,
             MinimumLevelDirective,
+            MinimumLevelControlledByDirective,
             EnrichWithPropertyDirective,
             EnrichWithDirective,
             FilterDirective
@@ -118,14 +120,38 @@ namespace Serilog.Settings.KeyValuePairs
             var matchObjectInstantiations = new Regex(ObjectInstantiationDirectiveRegex);
 
             var objectInstantiationDirectives = (from wt in directives
-                                                where matchObjectInstantiations.IsMatch(wt.Key)
-                                                let match = matchObjectInstantiations.Match(wt.Key)
-                                                select new
-                                                {
-                                                    ObjectType = InstantiableObjectTypes[match.Groups["objectTypeAlias"].Value],
-                                                    VariableName = match.Groups["variableName"].Value,
-                                                    ConstructorParamValue = wt.Value
-                                                }).ToList();
+                                                 where matchObjectInstantiations.IsMatch(wt.Key)
+                                                 let match = matchObjectInstantiations.Match(wt.Key)
+                                                 select new
+                                                 {
+                                                     ObjectType = InstantiableObjectTypes[match.Groups["objectTypeAlias"].Value],
+                                                     VariableName = match.Groups["variableName"].Value,
+                                                     ConstructorParamValue = wt.Value
+                                                 }).ToList();
+
+            var variablesInScope = new Dictionary<string, object>();
+            foreach (var objectInstantiationDirective in objectInstantiationDirectives)
+            {
+                if (TryInstantiate(objectInstantiationDirective.ObjectType, objectInstantiationDirective.ConstructorParamValue, out var paramValue))
+                {
+                    variablesInScope[objectInstantiationDirective.VariableName] = paramValue;
+                }
+            }
+
+            string minimumLevelControlledByLevelSwitchName;
+            if (directives.TryGetValue(MinimumLevelControlledByDirective, out minimumLevelControlledByLevelSwitchName))
+            {
+                var levelSwitch = variablesInScope
+                    .Where(kvp => kvp.Key == minimumLevelControlledByLevelSwitchName)
+                    .Select(kvp => kvp.Value)
+                    .OfType<LoggingLevelSwitch>()
+                    .FirstOrDefault();
+
+                if (levelSwitch != null)
+                {
+                    loggerConfiguration.MinimumLevel.ControlledBy(levelSwitch);
+                }
+            }
 
             var matchCallables = new Regex(CallableDirectiveRegex);
 
@@ -147,14 +173,7 @@ namespace Serilog.Settings.KeyValuePairs
             {
                 var configurationAssemblies = LoadConfigurationAssemblies(directives);
 
-                var variablesInScope = new Dictionary<string, object>();
-                foreach (var objectInstantiationDirective in objectInstantiationDirectives)
-                {
-                    if (TryInstantiate(objectInstantiationDirective.ObjectType, objectInstantiationDirective.ConstructorParamValue, out var paramValue))
-                    {
-                        variablesInScope[objectInstantiationDirective.VariableName] = paramValue;
-                    }
-                }
+
 
 
                 foreach (var receiverGroup in callableDirectives.GroupBy(d => d.ReceiverType))
--- a/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
+++ b/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
@@ -41,7 +41,7 @@ namespace Serilog.Settings.KeyValuePairs
         const string MinimumLevelOverrideDirectivePrefix = "minimum-level:override:";
 
         const string CallableDirectiveRegex = @"^(?<directive>audit-to|write-to|enrich|filter):(?<method>[A-Za-z0-9]*)(\.(?<argument>[A-Za-z0-9]*)){0,1}$";
-        const string ObjectInstantiationDirectiveRegex = @"^(?<objectTypeAlias>level-switch):(?<variableName>[A-Za-z0-9]+)$";
+        const string LevelSwitchDeclarationDirectiveRegex = @"^level-switch:(?<switchName>[A-Za-z0-9]+)$";
 
         static readonly string[] _supportedDirectives =
         {
@@ -54,11 +54,6 @@ namespace Serilog.Settings.KeyValuePairs
             EnrichWithPropertyDirective,
             EnrichWithDirective,
             FilterDirective
-        };
-
-        static readonly Dictionary<string, Type> InstantiableObjectTypes = new Dictionary<string, Type>
-        {
-            ["level-switch"] = typeof(LoggingLevelSwitch),
         };
 
         static readonly Dictionary<string, Type> CallableDirectiveReceiverTypes = new Dictionary<string, Type>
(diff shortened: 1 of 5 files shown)


test/Serilog.Tests/Settings/KeyValuePairSettingsTests.cs
```diff
@@ -152,5 +153,89 @@ public void TestMinimumLevelOverrides()
...
+            };
+
+            LogEvent evt = null;
+
+            var log = new LoggerConfiguration()
+                .ReadFrom.KeyValuePairs(settings)
+                .WriteTo.Sink(new DelegatingSink(e => evt = e))
+                .CreateLogger();
+
+            // TODO : maybe add Logger.LevelSwitch as an internal readonly property to avoid reflection here ?
```
tsimbalar: I'm using reflection to access a private field `_levelSwitch` of class `Logger` .... 
I'm not super happy about that, but 
- I believe accessing that member is the only way to test that the Level switch is really applied (I have no "pointer" to the actual object because it was defined from config....)
- I was not sure whether it was ok to make the field `internal` to make it visible from the test (or insite an `internal` get-only property)

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -106,6 +118,19 @@ public void Configure(LoggerConfiguration loggerConfiguration)
                 }
             }
 
+            var declaredVariables = ParseVariableDeclarationDirectives(directives);
+
+            string minimumLevelControlledByLevelSwitchName;
+            if (directives.TryGetValue(MinimumLevelControlledByDirective, out minimumLevelControlledByLevelSwitchName))
+            {
+                var levelSwitch = LookUpVariable(minimumLevelControlledByLevelSwitchName, typeof(LoggingLevelSwitch), declaredVariables) as LoggingLevelSwitch;
+                // TODO : maybe throw an exception to notify missing variable declaration ?
```
tsimbalar: I'm not sure what is the *policy* around exception in Serilog. 
I know the idea is that a logging library shouldn't crash the app, but is it ok to throw exceptions during configuration, to help the developer fix the incorrect config ? 
I believe it would be more user friendly to throw an exception that states what is exactly wrong (undeclared variable / wrong name etc), but I was not sure :p

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -185,6 +252,36 @@ internal static IEnumerable<Assembly> LoadConfigurationAssemblies(Dictionary<str
             return configurationAssemblies.Distinct();
         }
 
+        internal static bool TryInstantiate(Type type, string constructorParamAsString, out object createdInstance)
```
tsimbalar: I'm wondering if I did not over-engineer that part ... Seeing how the "callable directives" relied on reflection to find the proper method to call, I did the same to try and find a constructor for the type we are trying to build .... but considering the only type we need to build is always `LoggingLevelSwitch`, that might be overkill ... 
If you think this should be simplified in order not to rely on reflection, that's fine, I'll do it :p

tsimbalar:
I added some comments in places where I was unsure about the proper way of doing things to be consistent with the rest of the code base ...

test/Serilog.Tests/Settings/KeyValuePairSettingsTests.cs
```diff
@@ -152,5 +153,90 @@ public void TestMinimumLevelOverrides()
...
+            Assert.Equal(expectedSwitchInitialLevel, ((LoggingLevelSwitch)actual).MinimumLevel);
+        }
+
+        [Fact]
+        public void LoggingLevelSwitchIsConfigured()
+        {
+            var settings = new Dictionary<string, string>
+            {
+                ["level-switch:Switch1"] = "Information",
+                ["minimum-level:controlled-by"] = "Switch1",
```
nblumhardt: `<add key="serilog:level-switch:Switch1" value="Information" />` contains the same information but also enforces that if a switch is declared, we set its level. Separating the two means we can declare a switch without using it, or try to use a nonexistent switch. Do we gain anything by splitting the declaration and usage across two directives?

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
nblumhardt: https://github.com/serilog/serilog/wiki/Reliability should have some clues around this (it's a little vague, but sets out the general idea).

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
nblumhardt: I'm just doing a quick first-pass over to try to provide some feedback, so haven't give it a huge amount of thought yet, but my instinct says just building something explicit to support `LevelSwitch` would be reasonable right now. 

We're likely to address general object construction through the requirement that sinks be able to receive complex parameters to their configuration methods - `ColumnOptions` for the MSSQL sink, for example. My bet is we'll revisit general construction rules then.

test/Serilog.Tests/Settings/KeyValuePairSettingsTests.cs
nblumhardt: Could we just infer the level by testing with a couple of different events? May be a bit clunky, but should stay stable over the long term via the public API.

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
tsimbalar: thanks ! I knew I had read something along those lines somewhere, I just couldn't remember :)
I'll change the code to throw a more explicit exception

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
tsimbalar: ok, go for YAGNI then :)

test/Serilog.Tests/Settings/KeyValuePairSettingsTests.cs
tsimbalar: yes, that would prove that the initial value for the "level switch" was indeed passed, but not that it was assigned .... but I guess it is covered by the full test-case (passing the switch also to a sink)

test/Serilog.Tests/Settings/KeyValuePairSettingsTests.cs
tsimbalar: Sorry, I'm not sure I understand your comment U_U 

I just based my implementation on your proposal from here : https://github.com/serilog/serilog-settings-appsettings/issues/7#issuecomment-325182308 

The switch needs to be used twice, both for the : 
- `.MinimumLevel.ControlledBy(LoggingLevelSwitch ls)`
- `.WriteTo.Seq( LoggingLevelSwitch controlLevelSwitch)`
... so we need to find a way to express that both parameters refer to the same instance of a LoggingSwitch , or did I miss something ?

nblumhardt: Sorry, yes, you're entirely correct, I misread the code in my attempt to skim through quickly.

tsimbalar:
I think I'm done with this PR. 
Please tell me if anything needs changing !

tsimbalar:
I have added code to handle referencing a `LoggingLevelSwitch` also from a `MinimumLevel overrife`

nblumhardt:
Nice catch on the `Override`s 👍 

This does have me thinking, when I see:

```
["minimum-level:override:System"] = "specificSwitch"
```

Alongside:

```
["minimum-level:override:System"] = "Warning"
```

I'm reminded again of the possibility of disambiguating by forcing some kind of 'variable reference'-style syntax:

```
["minimum-level:override:System"] = "$specificSwitch"
```

I don't want to take the discussion around in circles :-), but what are your thoughts about this case?

tsimbalar:
I do like that there is an indication that it is a reference to a variable :) . This also help explicitly look up a value by name instead of first trying to look up a value by name and then fall back to conversion. For consistence, I would also add a specific symbol when declaring the variable (`php`anyone ? :p) , that would make it easier to visually recognize occurrences of the variable in the config file.

If we do not want to go too deep into variables and stuff like that and really just support `LoggingLevelSwitch`es for now , maybe an option is to :
- change the "regex" for switch declaration to enforce that the switch name must start with a `$` in `level-switch` directives (thus forbidding to declare a switch named `Information` or `Error`
```xml
<add key="serilog:level-switch:$systemSwitch" value="Warning" />
```
- when calling configuration methods and trying to find a value to pass for each argument, keep on looking for a switch declared with name `$switch` ... and fall back to conversion .... without risk of interpreting it as a LoggingLevel .

(I think we still need to fallback to conversion because of possible Sink configuration methods that would accept a string, and in which the string `"$this is my string"` is valid (for instead file paths on file loggers ?) ... )

To support this, the only change needed is to change the regex for switch-level declarations, the rest would work as is, I believe.

This allows to support the use case for now, without closing too many doors for a possible more generic implementation in the future.

The final syntax would then be :
```xml
<!-- declare the switch -->
<add key="serilog:level-switch:$Switch1" value="Debug" />
<!-- use it to control the root logger -->
<add key="serilog:minimum-level:controlled-by" value="$Switch1" />
<add key="serilog:using:Seq" value="Serilog.Sinks.Seq" />
<add key="serilog:write-to:Seq.serverUrl" value="http://localhost:5341" />
<!-- give the sink access to the switch -->
<add key="serilog:write-to:Seq.controlLevelSwitch" value="$Switch1" />
```

and for the overrides

```xml
<add key="serilog:level-switch:$systemSwitch" value="Warning" />
<add key="serilog:minimum-level:override:System" value="$systemSwitch" />
<add key="serilog:using:Seq" value="Serilog.Sinks.Seq" />
<add key="serilog:write-to:Seq.serverUrl" value="http://localhost:5341" />
<!-- give the sink access to the switch -->
<add key="serilog:write-to:Seq.controlLevelSwitch" value="$systemSwitch" />
```

nblumhardt:
@tsimbalar that looks good to me 👍 (sorry about the slow reply).

tsimbalar:
@nblumhardt no problem ;)  OK, I'll change the regex to accept only switches in the format `$switchName` ... and if I find the time, I'll start also issuing pull requests for documentation updates

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -81,6 +88,9 @@ public void Configure(LoggerConfiguration loggerConfiguration)
                 .Where(k => _supportedDirectives.Any(k.StartsWith))
                 .ToDictionary(k => k, k => _settings[k]);
 
+            var declaredLevelSwitches = ParseNamedLevelSwitchDeclarationDirectives(directives);
+
+
```
nblumhardt: Nit: extra whitespace

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -37,13 +41,16 @@ class KeyValuePairSettings : ILoggerSettings
         const string MinimumLevelOverrideDirectivePrefix = "minimum-level:override:";
 
         const string CallableDirectiveRegex = @"^(?<directive>audit-to|write-to|enrich|filter):(?<method>[A-Za-z0-9]*)(\.(?<argument>[A-Za-z0-9]*)){0,1}$";
+        const string LevelSwitchDeclarationDirectiveRegex = @"^level-switch:(?<switchName>\$[A-Za-z0-9]+)$";
```
nblumhardt: Instead of enforcing the presence of `$` here, would it be friendlier to accept anything, but emit a warning or throw when the directive is found but the name is invalid?

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -135,12 +159,63 @@ select new
...
+                if (string.IsNullOrEmpty(switchDeclarationDirective.InitialSwitchLevel))
+                {
+                    newSwitch = new LoggingLevelSwitch();
+                }
+                else
+                {
+                    var initialLevel = (LogEventLevel)SettingValueConversions.ConvertToType(switchDeclarationDirective.InitialSwitchLevel, typeof(LogEventLevel));
+                    newSwitch = new LoggingLevelSwitch(initialLevel);
+                }
+                namedSwitches[switchDeclarationDirective.SwitchName] = newSwitch;
```
nblumhardt: Can we assume keys are unique, and so use `namedSwitches.Add()` here to express/enforce that expectation?

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -135,12 +159,63 @@ select new
...
+
+            }
+            return new ReadOnlyDictionary<string, LoggingLevelSwitch>(namedSwitches);
+        }
+
+        static LoggingLevelSwitch LookUpSwitchByNameOrThrow(string minimumLevelControlledByLevelSwitchName, IReadOnlyDictionary<string, LoggingLevelSwitch> declaredLevelSwitches)
+        {
+            if (!declaredLevelSwitches.ContainsKey(minimumLevelControlledByLevelSwitchName))
```
nblumhardt: How about:

```csharp
if (declaredLevelSwitches.TryGetValue(minimumLevelControlledByLevelSwitchName, out var levelSwitch)
    return levelSwitch;

throw new InvalidOperationException($"No `LoggingLevelSwitch` has been declared with name \"{minimumLevelControlledByLevelSwitchName}\". You might be missing a key \"{LevelSwitchDirective}:{minimumLevelControlledByLevelSwitchName}\"");
```

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -135,12 +159,63 @@ select new
...
+                    var initialLevel = (LogEventLevel)SettingValueConversions.ConvertToType(switchDeclarationDirective.InitialSwitchLevel, typeof(LogEventLevel));
+                    newSwitch = new LoggingLevelSwitch(initialLevel);
+                }
+                namedSwitches[switchDeclarationDirective.SwitchName] = newSwitch;
+
+            }
+            return new ReadOnlyDictionary<string, LoggingLevelSwitch>(namedSwitches);
+        }
+
+        static LoggingLevelSwitch LookUpSwitchByNameOrThrow(string minimumLevelControlledByLevelSwitchName, IReadOnlyDictionary<string, LoggingLevelSwitch> declaredLevelSwitches)
```
nblumhardt: Could we just call the argument `switchName`? The result might not always be used in a `minimum-level:controlled-by` directive, so the name might be a bit too specific.

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -135,12 +159,63 @@ select new
...
+            }
+            var levelSwitch = declaredLevelSwitches[minimumLevelControlledByLevelSwitchName];
+            return levelSwitch;
+        }
+
+        static object ConvertOrLookupByName(string valueOrSwitchName, Type type, IReadOnlyDictionary<string, LoggingLevelSwitch> declaredSwitches)
+        {
+            if (type == typeof(LoggingLevelSwitch))
+            {
+                return LookUpSwitchByNameOrThrow(valueOrSwitchName, declaredSwitches);
```
nblumhardt: The naming scheme here isn't consistent - since `ConvertOrLookupByName()` could throw, I'd expect it to also carry the `OrThrow()` suffix if we're using that convention. Perhaps we could drop the `OrThrow()` from `LookUpSwitchByNameOrThrow()`, with the intention that we'd provide a `TryXxxx()` variant for the non-throwing version?

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
tsimbalar: Yes that would be friendlier indeed :)
I would say "throw" in here, the same way that right now for instance setting an invalid url in the Seq Sink throws a FormatException somewhere down the line.

I think throwing a `FormatException` with a detailed message would make sense. I don't think creating a custom exception type would add any value.

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
tsimbalar: your OCD might be worse than mine ! ;)

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
tsimbalar: yep :) That's a leftover of R# *Extract Method* ;)

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
tsimbalar: makes sense

tsimbalar:
Thanks for the review, I'll fix'em when I get a moment.

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
nblumhardt: Agreed :+1:

tsimbalar:
I've updated my PR to include the suggested changes. Please tell me if I need to do anything else (for instance, should I squash / rebase my commits, or is it ok to see the sausage-making ?)

nblumhardt:
Squashed and merged, thanks!

Issue #967, opened 2017-05-04
MinimumLevel.Override not working inside child logger (WriteTo.Logger())


AR1ES:
Serilog version 2.4.0. Simple demo:
```
namespace SerilogOverrideErrorDemo {
	class Program {
		static void Main(string[] args) {
			var logBuilder = new LoggerConfiguration()
				.MinimumLevel.Information()
				.WriteTo.Logger(l => l.MinimumLevel.Information().MinimumLevel.Override("SerilogOverrideErrorDemo.Class2", LogEventLevel.Error).WriteTo.LiterateConsole()
				);
			Log.Logger = logBuilder.CreateLogger();
			var l1 = Log.ForContext<Class1>();
			var l2 = Log.ForContext<Class2>();
			l1.Information("Info 1");
			l2.Information("Info 2");
			l1.Error("Error 1");
			l2.Error("Error 2");
			Console.ReadLine();
		}
	}
	class Class1 {
	}
	class Class2 {
	}
}
```
Output:
```
[14:06:12 INF] Info 1 - expected
[14:06:13 INF] Info 2 - unexpected
[14:06:13 ERR] Error 1 - expected
[14:06:13 ERR] Error 2 - expected
```

It works fine if "Override" is placed in rootLoggerConfiguration:
```
			var logBuilder = new LoggerConfiguration()
				.MinimumLevel.Information()
				.MinimumLevel.Override("SerilogOverrideErrorDemo.Class2", LogEventLevel.Error)
				.WriteTo.LiterateConsole();
```

nblumhardt:
Thanks for the report 👍

johnduhart:
I did some looking on this, and it's a bit tricky. Here's a failing test case I assembled using the original report:

```csharp
        [Fact]
        public void Issue967()
        {
            var sink = new CollectingSink();
            var logger = new LoggerConfiguration()
                .MinimumLevel.Information()
                .WriteTo.Logger(lc => lc
                    .MinimumLevel.Information()
                    .MinimumLevel.Override("Serilog.Tests", LogEventLevel.Error)
                    .WriteTo.Sink(sink))
                .CreateLogger();

            logger.Write(Some.InformationEvent());
            logger.Write(Some.LogEvent(level: LogEventLevel.Error));

            var contextLogger = logger.ForContext<LoggerConfigurationTests>();
            contextLogger.Write(Some.InformationEvent());
            contextLogger.Write(Some.LogEvent(level: LogEventLevel.Error));

[...]

PR #1027, opened 2017-09-17
Support for Overrides in Child loggers - issue #967

tsimbalar:
Issue #967 

It shouldn't ! 
It does introduce new types / new constructor overrides, but they are maked as internal.

**Other information**:
All the tests pass, but there is definitely room for improvement :)

### DIFF
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -108,7 +108,7 @@ namespace Serilog.Core
                         enricher,
                         null,
                         _levelSwitch,
-                        OverrideMap);
+                        _overrideMap);
         }
 
         /// <summary>
@@ -148,11 +148,11 @@ namespace Serilog.Core
 
             var minimumLevel = _minimumLevel;
             var levelSwitch = _levelSwitch;
-            if (OverrideMap != null && propertyName == Constants.SourceContextPropertyName)
+            if (_overrideMap != null && propertyName == Constants.SourceContextPropertyName)
             {
                 var context = value as string;
                 if (context != null)
-                    OverrideMap.GetEffectiveLevel(context, out minimumLevel, out levelSwitch);
+                    _overrideMap.GetEffectiveLevel(context, out minimumLevel, out levelSwitch);
             }
 
             return new Logger(
@@ -162,7 +162,7 @@ namespace Serilog.Core
                 enricher,
                 null,
                 levelSwitch,
-                OverrideMap);
+                _overrideMap);
         }
 
         /// <summary>
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -88,6 +88,9 @@ namespace Serilog.Core
             _enricher = enricher;
         }
 
+        internal bool HasOverrides => _overrideMap != null;
+        internal LevelOverrideMap OverrideMap => _overrideMap;
+
         /// <summary>
         /// Create a logger that enriches log events via the provided enrichers.
         /// </summary>
@@ -105,7 +108,7 @@ namespace Serilog.Core
                         enricher,
                         null,
                         _levelSwitch,
-                        _overrideMap);
+                        OverrideMap);
         }
 
         /// <summary>
@@ -145,11 +148,11 @@ namespace Serilog.Core
 
             var minimumLevel = _minimumLevel;
             var levelSwitch = _levelSwitch;
-            if (_overrideMap != null && propertyName == Constants.SourceContextPropertyName)
+            if (OverrideMap != null && propertyName == Constants.SourceContextPropertyName)
             {
                 var context = value as string;
                 if (context != null)
-                    _overrideMap.GetEffectiveLevel(context, out minimumLevel, out levelSwitch);
+                    OverrideMap.GetEffectiveLevel(context, out minimumLevel, out levelSwitch);
             }
 
             return new Logger(
@@ -159,7 +162,7 @@ namespace Serilog.Core
                 enricher,
                 null,
                 levelSwitch,
-                _overrideMap);
+                OverrideMap);
         }
 
         /// <summary>
@@ -273,11 +276,11 @@ namespace Serilog.Core
         /// <returns>True if the level is enabled; otherwise, false.</returns>
         public bool IsEnabled(LogEventLevel level)
         {
-            if ((int) level < (int) _minimumLevel)
+            if ((int)level < (int)_minimumLevel)
                 return false;
 
             return _levelSwitch == null ||
-                   (int) level >= (int) _levelSwitch.MinimumLevel;
+                   (int)level >= (int)_levelSwitch.MinimumLevel;
         }
 
         /// <summary>
@@ -1359,7 +1362,7 @@ namespace Serilog.Core
                 return false;
             }
 
-            property =_messageTemplateProcessor.CreateProperty(propertyName, value, destructureObjects);
+            property = _messageTemplateProcessor.CreateProperty(propertyName, value, destructureObjects);
             return true;
         }
 
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -88,8 +88,11 @@ namespace Serilog.Core
             _enricher = enricher;
         }
 
-        internal bool HasOverrides => _overrideMap != null;
-        internal LevelOverrideMap OverrideMap => _overrideMap;
+        internal bool TryGetOverrideMap(out LevelOverrideMap overrideMap)
+        {
+            overrideMap = _overrideMap;
+            return _overrideMap != null;
+        }
 
         /// <summary>
         /// Create a logger that enriches log events via the provided enrichers.
--- a/src/Serilog/Core/Sinks/OverrideRestrictingSink.cs
+++ b/src/Serilog/Core/Sinks/OverrideRestrictingSink.cs
@@ -1,0 +1,66 @@
+﻿// Copyright 2013-2015 Serilog Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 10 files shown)


tsimbalar:
There are a few things I did that may or may not be considered acceptable :) : 
- added an `internal` secondary constructor to `LoggerConfiguration`
- added a new `internal` type `OverrideRestrictingSink` which is a decorator ... not super happy about the name :)
- exposed as `internal` property the field `Logger._overrideMap`
- I have reused the type `LevelOverrideMap` which I think is fine, but maybe there's a better way :p
- I have unconsciously reformatted some files (`Ctrl+K,D` with ReSharper) ... I just noticed it now

I'd be happy to get and review and change whatever needs to be changed :)

src/Serilog/Core/Sinks/OverrideRestrictingSink.cs
```diff
@@ -0,0 +1,74 @@
...
+using System;
+using System.Collections.Generic;
+using Serilog.Events;
+
+namespace Serilog.Core.Sinks
+{
+    class OverrideRestrictingSink : ILogEventSink
+    {
+        readonly ILogEventSink _sink;
+        LevelOverrideMap _overrideMap;
```
nblumhardt: Looks like this one could be `readonly`

src/Serilog/Core/Sinks/OverrideRestrictingSink.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public OverrideRestrictingSink(ILogEventSink sink, Dictionary<string, LoggingLevelSwitch> overrides)
+        {
+            if (sink == null) throw new ArgumentNullException(nameof(sink));
+            if (overrides == null) throw new ArgumentNullException(nameof(overrides));
+            _sink = sink;
+            _overrideMap = new LevelOverrideMap(overrides, LevelAlias.Minimum, null);
+
```
nblumhardt: Nit: whitespace

nblumhardt:
@tsimbalar thanks! The PR looks thorough 👍 

I will need some time to properly wrap my head around the implementation (sorry turnaround time is a bit slow at the moment).

(The more eyes on this, the better, in case anyone else watching can take a look, too.)

src/Serilog/Core/Sinks/OverrideRestrictingSink.cs
tsimbalar: yup, good catch :)

src/Serilog/Core/Sinks/OverrideRestrictingSink.cs
tsimbalar: I forgot a `Ctrl+K,D` somewhere here :)

tsimbalar:
Btw, I didn't see it anywhere, and I don't think it is covered in the `editor.config` file, but is it ok to use syntactic candy from the latest versions of C# in the code base (namely `out var` and *pseudo pattern-matching* like `if foo is Logger x`) ? mostly it's ReSharper's fault (and some OCD of mine) : I find it hard to ignore R# squigglies and suggestions to use the newer syntax stuff ...

src/Serilog/LoggerConfiguration.cs
```diff
@@ -36,13 +36,34 @@ public class LoggerConfiguration
...
         readonly Dictionary<string, LoggingLevelSwitch> _overrides = new Dictionary<string, LoggingLevelSwitch>();
+        readonly Action<string, LoggingLevelSwitch> _addOverrideCallback;
         LogEventLevel _minimumLevel = LogEventLevel.Information;
         LoggingLevelSwitch _levelSwitch;
         int _maximumDestructuringDepth = 10;
         int _maximumStringLength = int.MaxValue;
         int _maximumCollectionCount = int.MaxValue;
         bool _loggerCreated;
 
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
```
nblumhardt: Todo - doc

src/Serilog/LoggerConfiguration.cs
```diff
@@ -36,13 +36,34 @@ public class LoggerConfiguration
...
+        public LoggerConfiguration()
+#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
+            : this(null)
+        {
+
```
nblumhardt: Nit: whitespace

src/Serilog/Core/Sinks/OverrideRestrictingSink.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright 2013-2015 Serilog Contributors
```
nblumhardt: Nit: date

src/Serilog/Core/Sinks/OverrideRestrictingSink.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Serilog.Core.Sinks
+{
+    class OverrideRestrictingSink : ILogEventSink
+    {
+        readonly ILogEventSink _sink;
+        readonly LevelOverrideMap _overrideMap;
+
+        public OverrideRestrictingSink(ILogEventSink sink, Dictionary<string, LoggingLevelSwitch> overrides)
+        {
+            if (sink == null) throw new ArgumentNullException(nameof(sink));
```
nblumhardt: Can use the `?? throw`-expression idiom here.

src/Serilog/Core/Logger.cs
```diff
@@ -88,6 +88,9 @@ public sealed class Logger : ILogger, ILogEventSink, IDisposable
             _enricher = enricher;
         }
 
+        internal bool HasOverrides => _overrideMap != null;
+        internal LevelOverrideMap OverrideMap => _overrideMap;
```
nblumhardt: Could we use the `bool TryGetOverrideMap(out OverrideMap)` pattern to avoid potentially exposing a null-valued property?

nblumhardt:
I think this is a reasonable approach. I like that it's only activated in the scenario that we're aiming to fix 👍.

A couple of considerations -

 * It seems like it should work recursively (sub-loggers of sub-loggers) but I don't think it's tested.
 * Would also be good to do a before/after perf run, and add a benchmark to see what impact it has on writing through sub-loggers.
 * Finally, there's one sneaky failure mode - if enrichers modify `SourceContext` the feature won't work correctly; it's probably not worth our while trying to code around that possibility, but something to look out for.

There's only one thing that troubles me; level overrides can be used as filters, but they're different from filters because they can make logging more verbose than the default, rather than just restrict it.

In this version of the feature, I think we only end up supporting the filtering case; the outer logger's level will always restrict the events that reach the sub-logger. (Please correct me if I'm missing something).

If that's all the feature can support, would we be better off just pushing people to use filters on sub-loggers? It seems like we could use the mechanism introduce here to rig `MinimumLevel.Override()` to throw when called on a sub-logger, with a message suggesting the use of a filter instead...

The advantage of simply not supporting overrides on sub-loggers would be avoiding inconsistent behavior. What do you think?

tsimbalar:
While implementing, I did wonder if it really was worth the pain :p

In my head, the general rule for sub-loggers is : ~~you don't talk about sub-loggers~~ *a sub-logger will **never** see an event that its parent logger rejected* , and I think this should also be true about overrides on sub-loggers.

That is, in the following code : 
```csharp
new LoggerConfiguration()
     .MinimumLevel.Verbose()
     .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
     .WriteTo.Logger(lc => 
          lc.MinimumLevel.Override("Microsoft", LogEventLevel.Debug)
          WriteTo.Whatever()
     )
```
... the *Whatever* sink will never see any event with level `< Information` even though the overrides on the child logger accepts *Debug*. 
This, as you said, ends up being just a very specific kind of `Filter` that is based on the `SourceContext`, with the difference that it can also use a `LoggingLevelSwitch` but it is probably there by coincidence. 

The only problem I see is that, because of the way the Configuration API is designed, people might already be using calls like the one above in the wild, and suddenly throwing in those cases might be problematic ... Maybe just writing to `SelfLog` would be enough ? 

So yes, I agree, discouraging people from using Overrides in subloggers is probably better for everyone :)

Here is a suggestion of what we could do to help people not using an unsupported feature : 
- when calling `WriteTo.Logger(ILogger)` : if the `ILogger` is a `Logger` with overrides, write to the `SelfLog` that this is not supported and that the overrides will be ignored
- for the `WriteTo.Logger(Action<LoggerConfiguration>)` : actually change it to `WriteTo.Logger(Action<SubLoggerConfiguration>)` , where `SubLoggerConfiguration` is a subclass of `LoggerConfiguration` for which `MinimumLevel.Overrides()` is hidden / is `Obsolete` / is `NonBrowsable` / throws `NotSupported` / writes to `SelfLog` . I believe this would not break stuff on the calling side, given that we are just making the API more specific. Being that specific would allow to hide stuff that we do not want to support in sub-loggers.


So in the end, it might worth doing a completely different PR about **officially not supporting Minimum Level Overrides in sub-loggers** that would contain : 
- mentioning it in the docs somewhere
- doing everything we can so that the option does not show up in Intellisense
- writing to `SelfLog` that this will be ignored when it is used anyway (or `throw` if we really want to make it clear)

Does that make sense ?

tsimbalar:
Actually, thinking about it a bit more, I believe there are several scenarios that would probably not need to be supported on sub-loggers ... and ideally, if they don't make sense, they should not even appear as an option in the configuration API .... 
I will create a more general issue about this topic

tsimbalar:
The code contains the suggested changes, but still implements a limited version of sub-logger overrides... 
I guess it's better to not support it at all rather that supporting a limited/broken version
I'll issue another PR that completely removes support .

nblumhardt:
Let's close this for now in favor of #1033, we can re-open and pick this up another time if we decide to.

Issue #1032, opened 2017-09-25
Drop support for some features in sub-loggers


tsimbalar:
While working on an issue in order to *honor* overrides configuration in *sub-loggers* (`Logger.WriteToLogger(...)` methods see #967 ) and going around the API and use cases (see https://github.com/serilog/serilog/pull/1027#issuecomment-331361672) , it turns out that maybe such a feature should not be supported.

I may be wrong, but I see at least 2 features that do not make much sense in sub-loggers : 
- the whole `LoggerConfiguration.ReadFrom`, because I can not imagine a scenario where the config from the sublogger would be loaded from somewhere else
- the `LoggerConfiguration.MinimumLevel.Overrides()` methods considering that `Filters` would probably be more appropriate

Now, the question is, if the feature is not supported, how should we make it obvious to the user that such a feature is not supported ? 

I can see several options : 
1. making the call throw a `UnsupportedException` when issued on a *sub-logger*
2. change the call to be a *no-op* but write something to the `SelfLog` to tell the user that the feature is not supported and its configuration will not be taken into account
3. marking the method as `Obsolete` and advising another workaround
4. marking it as not `EditorBrowsable` so that it is not suggested in *intellisense*
5. removing the calls from the API

- **Option 1** seems a bit extreme, and will cause some codebases to start failing at runtime when everything was just fine so far ...
- **Option 2** is a bit better, but there's a chance that nobody will ever see the message in the `SelfLog`
- **Option 3** would be nice in for a new *minor* version as it would give the users some time to migrate but would not break code straight away 
- **Option 4** is compatible with option 3
- **Option 5** would be nice for the next *major* version, it's definitely a **breaking** change

So I would like to suggest to follow **option 3** + **option 4** in the next minor versions in order to prepare for **option 5** in the next major version.

What do you think ? Are there more features that should not be exposed on sub-loggers ? 

*I do have some kind of ugly proof of concept of what a code-compatible but not binary-compatible version would look like in order to hide `MinimumLevel.Overrides` from sub-loggers here : https://github.com/tsimbalar/serilog/commit/21f9d503a9e83cc06c7c0dc3493c6dcf3f9e72a1*

nblumhardt:
Thanks for writing up the options :+1:

I think `ReadFrom()` could appear in valid sub-logger scenarios, but it'd be esoteric usage. If it functions correctly, I think it'd be fine to leave on the API unless it costs us in complexity.

[...]

PR #1033, opened 2017-09-28
Make it clear that Sub-logger MinimumLevel.Override is not supported

tsimbalar:
Related to #967 
This is a second PR  after comments on #1027 and discussion in #1032

The idea is that we no longer try to support `MinimumLevel.Override` in sub-loggers, but write to the `SelfLog` that it is not supported nor recommended, and using `.Filter` is better, or event better, the overrides of the root logger. 

No, it actually just makes it explicit that `MinimumLevel.Override` is not supported on sub-loggers

**Other information**:
I'm not super convinced about the message written to SelfLog :) Review from a native English speaker would be nice :)

### DIFF
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -32,12 +32,9 @@ namespace Serilog.Configuration
 
         internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action<ILogEventSink> addSink, Action<LoggerConfiguration> applyInheritedConfiguration)
         {
-            if (loggerConfiguration == null) throw new ArgumentNullException(nameof(loggerConfiguration));
-            if (addSink == null) throw new ArgumentNullException(nameof(addSink));
-            if (applyInheritedConfiguration == null) throw new ArgumentNullException(nameof(applyInheritedConfiguration));
-            _loggerConfiguration = loggerConfiguration;
-            _addSink = addSink;
-            _applyInheritedConfiguration = applyInheritedConfiguration;
+            _loggerConfiguration = loggerConfiguration ?? throw new ArgumentNullException(nameof(loggerConfiguration));
+            _addSink = addSink ?? throw new ArgumentNullException(nameof(addSink));
+            _applyInheritedConfiguration = applyInheritedConfiguration ?? throw new ArgumentNullException(nameof(applyInheritedConfiguration));
         }
 
         /// <summary>
@@ -128,9 +125,17 @@ namespace Serilog.Configuration
             var lc = new LoggerConfiguration();
 
             _applyInheritedConfiguration(lc);
-
             configureLogger(lc);
-            return Sink(new SecondaryLoggerSink(lc.CreateLogger(), attemptDispose: true), restrictedToMinimumLevel, levelSwitch);
+
+            var subLogger = lc.CreateLogger();
+            if (subLogger.HasOverrideMap)
+            {
+                SelfLog.WriteLine("Minimum level overrides are not supported on sub-loggers " +
+                                  "and may be removed completely in a future version.");
+            }
+
+            var secondarySink = new SecondaryLoggerSink(subLogger, attemptDispose: true);
+            return Sink(secondarySink, restrictedToMinimumLevel, levelSwitch);
         }
 
         /// <summary>
@@ -149,7 +154,15 @@ namespace Serilog.Configuration
             LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum)
         {
             if (logger == null) throw new ArgumentNullException(nameof(logger));
-            return Sink(new SecondaryLoggerSink(logger, attemptDispose: false), restrictedToMinimumLevel);
+
+            if (logger is Logger concreteLogger && concreteLogger.HasOverrideMap)
+            {
+                SelfLog.WriteLine("Minimum level overrides are not supported on sub-loggers " +
+                                  "and may be removed completely in a future version.");
+            }
+
+            var secondarySink = new SecondaryLoggerSink(logger, attemptDispose: false);
+            return Sink(secondarySink, restrictedToMinimumLevel);
         }
 
         /// <summary>
--- a/src/Serilog/LoggerConfiguration.cs
+++ b/src/Serilog/LoggerConfiguration.cs
@@ -80,7 +80,8 @@ namespace Serilog
             get
             {
                 return new LoggerMinimumLevelConfiguration(this,
-                    l => {
+                    l =>
+                    {
                         _minimumLevel = l;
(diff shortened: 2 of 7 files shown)


src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -128,9 +128,21 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
...
+
+            var subLogger = lc.CreateLogger();
+            if (subLogger.HasOverrideMap)
+            {
+                // Overrides in sub-loggers are not supported and will be ignored
+                SelfLog.WriteLine("We have detected that you are using .MinimumLevel.Override() from a sub-logger (.WriteTo.Logger(Action<LoggerConfiguration>)). " +
+                                  "Overrides in a sub-logger are not supported and are therefore ignored. " +
+                                  "We recommend applying .MinimulLevel.Override on the root logger or " +
+                                  "using .Filter in your sub-loggers. " +
+                                  "Note that support for sub-loggers' minimum level overrides may be removed in the next major version.");
```
nblumhardt: Reads very nicely 👍

We could probably cut it down a bit to just:

`Minimum level overrides are not supported on sub-loggers and may be removed completely in a future version.`

since it's googleable back to this issue :-) Happy to go with the longer version if you think it's preferable.

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -128,9 +128,21 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
...
+            }
+
+            ILogEventSink secondaryLoggerSink = new SecondaryLoggerSink(subLogger, attemptDispose: true);
```
nblumhardt: Can use `var`. Variable with the same purpose is called `secondarySink` in the method below.

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -149,7 +161,21 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
             LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum)
         {
             if (logger == null) throw new ArgumentNullException(nameof(logger));
-            return Sink(new SecondaryLoggerSink(logger, attemptDispose: false), restrictedToMinimumLevel);
+            ILogEventSink secondarySink = new SecondaryLoggerSink(logger, attemptDispose: false);
+
+            if (logger is Logger concreteLogger)
```
nblumhardt: `if (logger is Logger concreteLogger && concreteLogger.HasOverrideMap)` ?

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -128,9 +128,21 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
...
 
             _applyInheritedConfiguration(lc);
-
             configureLogger(lc);
-            return Sink(new SecondaryLoggerSink(lc.CreateLogger(), attemptDispose: true), restrictedToMinimumLevel, levelSwitch);
+
+            var subLogger = lc.CreateLogger();
+            if (subLogger.HasOverrideMap)
+            {
+                // Overrides in sub-loggers are not supported and will be ignored
```
nblumhardt: The comment here is probably redundant, given the content of the message we are writing to `SelfLog`.

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -149,7 +161,21 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
...
+            if (logger is Logger concreteLogger)
+            {
+                if (concreteLogger.HasOverrideMap)
+                {
+                    // Overrides in sub-loggers are not supported and will be ignored
+                    SelfLog.WriteLine("We have detected that you are using .MinimumLevel.Override() from a sub-logger (.WriteTo.Logger(Logger)). " +
+                                      "Overrides in a sub-logger are not supported and are therefore ignored. " +
+                                      "We recommend applying .MinimulLevel.Override on the root logger or " +
+                                      "using .Filter in your sub-loggers. " +
+                                      "Note that support for sub-loggers' minimum level overrides may be removed in the next major version.");
```
nblumhardt: Dropping the comment and using the shorter message would reduce the sense of duplication here.

src/Serilog/LoggerConfiguration.cs
```diff
@@ -69,6 +69,11 @@ void ApplyInheritedConfiguration(LoggerConfiguration child)
         /// </remarks>
         public LoggerAuditSinkConfiguration AuditTo => new LoggerAuditSinkConfiguration(this, s => _auditSinks.Add(s), ApplyInheritedConfiguration);
 
+        void AddOverride(string source, LoggingLevelSwitch levelSwitch)
```
nblumhardt: I don't know that this method is as useful now given the change in direction; inlining the assignment in the next method can save us five lines.

test/Serilog.Tests/Support/TemporarySelfLog.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using Serilog.Debugging;
+
+namespace Serilog.Tests.Support
+{
+    public class TemporarySelfLog : IDisposable
```
nblumhardt: We've needed something like this for a while 👍 

(A future version might modify this slightly to only siphon off messages for one logical call context/thread at a time, as this will have interleaving problems when we introduce more tests using it that might run in parallel.)

src/Serilog/LoggerConfiguration.cs
tsimbalar: yup, I'll put that back to its original state

test/Serilog.Tests/Support/TemporarySelfLog.cs
tsimbalar: Not quite sure what that would look like... `ThreadStatic` comes to mind, but not sure that would help in that case.

tsimbalar:
Thanks for the review ! 
I'll make the changes, and probably rebase/squash this PR, because it's starting to contain lots of commits going in very different directions and the history would be confusing

tsimbalar:
@nblumhardt not sure what is going wrong with the Travis CI build https://travis-ci.org/serilog/serilog/builds/285388771
... something about expired signature keys somewhere , related to mongodb.... 
looks like I'm not the only one, see https://github.com/travis-ci/travis-ci/issues/8554

tsimbalar:
Closed and re-opened to trigger a new build

nblumhardt:
Looks great :-)

PR #1034, opened 2017-09-29
Add log level conditioned 'ForContext'

### LINKED ISSUE #1002: Levelled `ForContext()`
`ForContext()` is widely-used to add properties to events independently of the message template.

This has some allocation overhead when logging is switched off, since `ForContext()` will allocate a few objects such as the return value and a property enricher.

The overhead can be avoided today by checking `IsEnabled()` and only conditionally adding properties when the level is known to be "on", but, it's quite clunky to do it.

An extension method like the one below would make it easier to add properties conditionally:

```csharp
static ILogger ForContext<TValue>(
    this ILogger logger,
    LogEventLevel level,
    string name, 
    TValue value,
    bool destructureObjects = false)
{
    if (!logger.IsEnabled(level))
        return logger;

    return logger.ForContext(name, value, destructureObjects);
}
```

Usage:

```csharp
log.ForContext(LogEventLevel.Debug, "OrderId", order.Id)
    .Information("Here is an event that may or may not be enriched");
```

Is it worth shipping this in Serilog somewhere?

pardahlman:
Hello 👋 
First time contributor to the Serilog project. Very excited to help out, and more than willing to adjust the PR based on feedback.

This issue fixes #1002 by following the suggested implementation.

No, it merely adds a convenience method for conditional enrich log context with provided property. 

**Other information**:

### DIFF
--- a/src/Serilog/Core/ForContextExtension.cs
+++ b/src/Serilog/Core/ForContextExtension.cs
@@ -30,9 +30,9 @@ namespace Serilog
             if (logger == null)
                 throw new ArgumentNullException(nameof(logger));
 
-            return !logger.IsEnabled(level)
-                ? logger
-                : logger.ForContext(propertyName, value, destructureObjects);
+            return logger.IsEnabled(level)
+                ? logger.ForContext(propertyName, value, destructureObjects)
+                : logger;
         }
     }
 }
--- a/src/Serilog/Core/ForContextExtension.cs
+++ b/src/Serilog/Core/ForContextExtension.cs
@@ -1,4 +1,5 @@
-﻿using Serilog.Events;
+﻿using System;
+using Serilog.Events;
 
 namespace Serilog
 {
@@ -26,6 +27,9 @@ namespace Serilog
             TValue value,
             bool destructureObjects = false)
         {
+            if (logger == null)
+                throw new ArgumentNullException(nameof(logger));
+
             return !logger.IsEnabled(level)
                 ? logger
                 : logger.ForContext(propertyName, value, destructureObjects);
--- a/src/Serilog/Core/ForContextExtension.cs
+++ b/src/Serilog/Core/ForContextExtension.cs
@@ -1,0 +1,34 @@
+﻿using Serilog.Events;
+
+namespace Serilog
+{
+    /// <summary>
+    /// Extension method 'ForContext' for ILogger.
+    /// </summary>
+    public static class ForContextExtension
+    {
+        /// <summary>
+        /// Create a logger that enriches log events with the specified property based on log event level.
+        /// </summary>
+        /// <typeparam name="TValue"> The type of the property value. </typeparam>
+        /// <param name="logger">The logger</param>
+        /// <param name="level">The log event level used to determine if log is enriched with property.</param>
+        /// <param name="propertyName">The name of the property. Must be non-empty.</param>
+        /// <param name="value">The property value.</param>
+        /// <param name="destructureObjects">If true, the value will be serialized as a structured
+        /// object if possible; if false, the object will be recorded as a scalar or simple array.</param>
+        /// <returns>A logger that will enrich log events as specified.</returns>
+        /// <returns></returns>
+        public static ILogger ForContext<TValue>(
+            this ILogger logger,
+            LogEventLevel level,
+            string propertyName,
+            TValue value,
+            bool destructureObjects = false)
+        {
+            return !logger.IsEnabled(level)
+                ? logger
+                : logger.ForContext(propertyName, value, destructureObjects);
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


nblumhardt:
This looks great, thanks! I should have a chance to review early in the week, unless I'm beaten to it :-)

src/Serilog/Core/ForContextExtension.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <returns>A logger that will enrich log events as specified.</returns>
+        /// <returns></returns>
+        public static ILogger ForContext<TValue>(
+            this ILogger logger,
+            LogEventLevel level,
+            string propertyName,
+            TValue value,
+            bool destructureObjects = false)
+        {
+            return !logger.IsEnabled(level)
```
nblumhardt: Need to check for a null `logger`, probably to throw `ArgumentNullException`, as this will throw `NullReferenceException` otherwise.

nblumhardt: Can invert the condition, no need for the `!` "not".

pardahlman: I agree with you on both points.  Fix on it's way!

pardahlman:
Fixes in place ✔️ . Let me know if you want me to squash the commits (I didn't as I'm not sure if that affects the review process here on Github).

nblumhardt:
Great, thanks 👍 Just letting you know this hasn't fallen off the radar - a lot on the plate, should be back to this one soon. Cheers!

src/Serilog/Core/ForContextExtension.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using Serilog.Events;
+
+namespace Serilog
+{
+    /// <summary>
+    /// Extension method 'ForContext' for ILogger.
+    /// </summary>
+    public static class ForContextExtension
```
nblumhardt: It's possible we'll add more extensions on `ILogger` in the future; shall we name this one `LoggerExtensions`?

src/Serilog/Core/ForContextExtension.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        /// <summary>
+        /// Create a logger that enriches log events with the specified property based on log event level.
+        /// </summary>
+        /// <typeparam name="TValue"> The type of the property value. </typeparam>
+        /// <param name="logger">The logger</param>
```
nblumhardt: Nitpick, missing `.`.

src/Serilog/Core/ForContextExtension.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using Serilog.Events;
+
+namespace Serilog
+{
+    /// <summary>
+    /// Extension method 'ForContext' for ILogger.
```
nblumhardt: `Extends <see cref="ILogger" /> with additional methods.` ?

src/Serilog/Core/ForContextExtension.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public static class ForContextExtension
+    {
+        /// <summary>
+        /// Create a logger that enriches log events with the specified property based on log event level.
```
nblumhardt: `Create a logger that enriches log events when the specified level is enabled.` ?

test/Serilog.Tests/Core/ForContextExtensionsTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        [Theory]
+        [InlineData(LogEventLevel.Debug, LogEventLevel.Debug)]
+        [InlineData(LogEventLevel.Debug, LogEventLevel.Information)]
+        [InlineData(LogEventLevel.Debug, LogEventLevel.Error)]
+        [InlineData(LogEventLevel.Debug, LogEventLevel.Fatal)]
+        [InlineData(LogEventLevel.Debug, LogEventLevel.Warning)]
+        public void ShouldEnrichLogEventWhenLevelIsSameOrHigherThanMinLevel(LogEventLevel logMinLevel, LogEventLevel propertLogLevel)
+        {
+            // Arrange
```
nblumhardt: I know there are other tests in the codebase commented with `// Arrange` etc., but we're not consistent - I'd prefer we drop these as after a time they become boilerplate.

pardahlman:
Top of the morning ☕️ ,

I'll look at this later today!

src/Serilog/Core/ForContextExtension.cs
pardahlman: Sure - my idea here was to create one extension class for each method name/overload and create more classes as more "types" of extension methods was introduces. But, it makes sense to have one class to start with and break it out to different classes when the need arises.

test/Serilog.Tests/Core/ForContextExtensionsTests.cs
pardahlman: Roger that - I'm renaming the test class name to reflect the name change of the extension method class.

nblumhardt:
Great, thank you!

Issue #1038, opened 2017-10-11
Proposal : Granular Log configuration combination


tsimbalar:
*Help needed to find a better issue title :p*

In this issue I propose a **new feature** in which we could extend the way the `LoggerConfiguration.ReadFrom.*` work in order to support finer control over which settings are declared "in the code" and which settings are declared "in app settings" (or any other source of `KayValuePair`s).

## Objective
What I would like is a way to define most of the logging configuration in code (with the .NET fluent API), but still let the **possibility to change/override some parts of this logging configuration at runtime**, for instance, reading a configuration file, or another source.

For instance, I would like to be able to define a *baseline* logging configuration as such in C# : 

```csharp
new LoggerConfiguration()
    .MinimumLevel.Debug()
    .MinimulLevel.Override("MyApp.Controllers", LogEventLevel.Warning)
    .WriteTo.RollingFile(@"C:\temp\whatever.log", restrictedToMinimumLevel:LogEventLevel.Information)
    .CreateLogger()
```
... but be able to declare the *log file path* in a config file, or override the *MinimumLevel* in some environments.
```xml
<appSettings>
  <add key="serilog:minimum-level" name="Warning" />
  <add key="serilog:write-to:RollingFile.pathFormat" name="D:\another\location.log" />
</appSettings>
```

## Current Limitation
The current way to provide parameters at *runtime* is to rely on some of the `ReadFrom.*` methods, but those can only *add* on top of a `LoggerConfiguration`, but not change the configuration defined in code (namely override parameters of the sinks for instance)

The workaround would be, in that case, to explicitely write the code that will read a specific `appSetting` and pass the runtime value to the `LoggerConfiguration` construction calls.

## Proposed solution(s)
What I am trying to achieve is to have more fine-grained access to every parameter that ends up being used in the `LoggerConfiguration` method invocations. This is pretty difficult to do when dealing with method calls, but the `KeyValuePair` already pretty much divides every parameter in individual chunk, so I believe we could manage to take advantage of that as an extension point.

Taking inspiration from the way *AspNetCore Configuration* are additive (first json, then environment variables etc ...), the idea would be, to : 
1. define the baseline configuration in terms of individual `KeyValuePair`s (like `"minimum-level"="Warning"` or `"write-to:RollingFile.pathFormat"="D:\another\location.log"`)
2. add to this list another set of `KeyValuePair`s that can add new values, or override the values of the previous
3. possibly add yet another source of settings
[...]

PR #1041, opened 2017-10-14
Allow combination of settings from multiple sources

tsimbalar:
This is a first step regarding the combination of multiple sources of key-value settings, to allow finer control over those settings

No. No existing public signature was changed (only additions)

**Other information**:
First part towards #1038 
- Introduces ISettingsSource to define a source of key-value settings
- Updates KeyValuePairSettings to rely on ISettingsSource
- Adds ReadFrom.Source / ReadFrom.Sources
- Introduces ISettingsSourceBuilder
- Introduces ReadFrom.Combined(Func<ISettingsSourceBuilder, ISettingsSourceBuilder>) to allow for fluent combination of sources
- Introduces SettingsSourceBuilderExtensions to support AddKeyValuePairs / AddKeyValuePair / AddFunc


tsimbalar:
This is a first step and contains only the core infrastructure to allow combining sources of settings.

## New types / methods 
- interface `ISettingsSource` represents a source of key-value settings, somehow a simplified form of [`Microsoft.Extensions.Configuration`](https://github.com/aspnet/Configuration/blob/dev/src/Config.Abstractions/IConfigurationProvider.cs)
- method `LoggerConfiguration.ReadFrom.Source(ISettingsSource source)` : that allows loading settings from an arbitrary `ISettingsSource`
- method `LoggerConfiguration.ReadFrom.Sources(params ISettingsSource[] sources)` : that allows loading settings from several sources, the key-value settings from sources specified later *win*
- method `LoggerConfiguration.ReadFrom.Combined(Func<ISettingsSourceBuilder, ISettingsSourceBuilder> builder)` : similar to the previous one, but based on a more chainable fluent API
- interface `ISettingsSourceBuilder` exists mostly to support a fluent API
- extension methods on `ISettingsSourceBuilder` `AddKeyValuePair`, `AddKeyValuePairs`, `AddFunc` that allow to write something like the following pseudo-code : 
```csharp
.ReadFrom.Combined(builder => builder
    .AddKeyValuePairs(IEnumerable<KeyValuePairs<string,string>>)
    .AddKeyValuePair(KeyValuePair<string,string>)
    .AddKeyValuePair(string key, string value)
    .AddFunc(() => IEnumerable<KeyValuePair<string, string>>)
)
- class `SettingsSource` : a concrete *not sealed* class that can be used as a base for other implementations of settings source. The idea is that we could use this base class to add *less stringy* methods to help implementers properly format the strings (for instance AddEnrichProperty(string propertyName, string propertyValue)) 
```

## Extension points : 
The idea is that external code can plug into it by  : 
- creating new implementations of `ISettingsSource` (for instance AppSettings, or Configuration, or Command-line ...) directly, or inheriting `SettingsSource`
- adding extension methods to `ISettingsSourceBuilder` 

## Feedback needed !
I would love feedback about the code itself, but also about the names of the classes and methods (I am myself not convinced about the choice, but could not figure out better names). 

In particular, I have doubts about : 
- namespaces : I've put most classes under `Serilog.Settings` , but the extension methods on `ISettingsSourceBuilder` are in the root `Serilog` namespace so that they show up when using `LoggerConfiguration` ... is that fine ? 
- naming : I'm not specially enthusiastic about the names `SettingsSource`,  `Source` and derivative, and they are a bit vague, but I couldn't find a better name. Same goes with `Combined` ...
- after second thought, I'm not so sure it makes sense to provide `.AddFunc(() => IEnumerable<KeyValuePair<string, string>>)` , specially with that name ...
- I'm not sure it makes sense to provider a base `SettingsSource` class, mostly because it may encourage inheritance for not so much added value 
- I'm not a native English speaker, so the `xml doc` comments may need reviewing/rewriting

tsimbalar:
I had second thoughts, and I simplified a bit : 
- all the overloads on `ReadFrom.` are now called `KeyValuePairs` , because that's what it is after all
- removed methods or classes that I felt were not really needed after all


As a results we have : 

- interface `ISettingsSource` represents a source of key-value settings, somehow a simplified form of [`Microsoft.Extensions.Configuration`](https://github.com/aspnet/Configuration/blob/dev/src/Config.Abstractions/IConfigurationProvider.cs)
- method `LoggerConfiguration.ReadFrom.KeyValuePairs(params ISettingsSource[] sources)` : that allows loading settings from one or several sources, the key-value settings from sources specified later *win*
- method `LoggerConfiguration.ReadFrom.KeyValuePairs(Func<ISettingsSourceBuilder, ISettingsSourceBuilder> builder)` : similar to the previous one, but based on a more chainable fluent API
- interface `ISettingsSourceBuilder` exists mostly to support a fluent API
- extension methods on `ISettingsSourceBuilder` `AddKeyValuePair`, `AddKeyValuePairs`, ~~`AddFunc`~~ that allow to write something like the following pseudo-code : 
```csharp
.ReadFrom.Combined(builder => builder
    .AddKeyValuePairs(IEnumerable<KeyValuePairs<string,string>>)
    .AddKeyValuePair(KeyValuePair<string,string>)
    .AddKeyValuePair(string key, string value)
)
```
- ~~class `SettingsSource` : a concrete *not sealed* class that can be used as a base for other implementations of settings source. The idea is that we could use this base class to add *less stringy* methods to help implementers properly format the strings (for instance AddEnrichProperty(string propertyName, string propertyValue))~~

tsimbalar:
This was probably a bit too early :) 
I'll close this PR, and issue another one once we know a bit more how we want to do it .... 

There are still discussions happening in #1038 and #1043

PR #1042, opened 2017-10-16
Make Resharper green again

tsimbalar:
Get rid of errors and warnings reported by ReSharper Solution-wide analysis

Nope

- *N/A* Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:
Before : 
before-resharper-cleanup

After : 
after-resharper-cleanup

There are still some warnings; but I was not sure how to fix them : 
```
Solution Serilog.sln
    Project Serilog
      src\Serilog\Capturing\PropertyValueConverter.cs:250 Possible 'System.NullReferenceException'
      src\Serilog\Capturing\PropertyValueConverter.cs:250 Possible 'System.NullReferenceException'
      src\Serilog\Context\LogContext.cs:257 Possible 'System.NullReferenceException'
      src\Serilog\Core\Pipeline\MessageTemplateCache.cs:54 The field is sometimes used inside synchronized block and sometimes used without synchronization
      src\Serilog\Core\Pipeline\MessageTemplateCache.cs:54 The field is sometimes used inside synchronized block and sometimes used without synchronization
      src\Serilog\Core\Pipeline\MessageTemplateCache.cs:54 The field is sometimes used inside synchronized block and sometimes used without synchronization
```

### DIFF
--- a/Serilog.sln.DotSettings
+++ b/Serilog.sln.DotSettings
@@ -50,6 +50,7 @@
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=PossibleNullReferenceException/@EntryIndexedValue">WARNING</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=PossibleUnintendedReferenceComparison/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=PrivateFieldCanBeConvertedToLocalVariable/@EntryIndexedValue">ERROR</s:String>
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=RedundantArgumentDefaultValue/@EntryIndexedValue">SUGGESTION</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=RedundantAssignment/@EntryIndexedValue">ERROR</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=RedundantBaseConstructorCall/@EntryIndexedValue">ERROR</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=RedundantBoolCompare/@EntryIndexedValue">ERROR</s:String>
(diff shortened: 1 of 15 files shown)


nblumhardt:
Great! Will have a look in detail ASAP.

```
      src\Serilog\Capturing\PropertyValueConverter.cs:250 Possible 'System.NullReferenceException'
```

This is a false positive, can be ignored.

```
      src\Serilog\Context\LogContext.cs:257 Possible 'System.NullReferenceException'
```

I am not 100% confident this can never occur, could ignore with a comment to that effect.

```
      src\Serilog\Core\Pipeline\MessageTemplateCache.cs:54 The field is sometimes used inside synchronized block and sometimes used without synchronization
```

This is by design and can be ignored.

tsimbalar:
Thanks, I've updated with your comments :)

ReSharper status : 
image

image

Out of curiosity, I looked up `ObjectHandle`, the base class of `DisposableObjectHandle` used in `LogContext`, and the [reference source contains](https://referencesource.microsoft.com/#mscorlib/system/runtime/remoting/objecthandle.cs,76) : 

```csharp
public override Object InitializeLifetimeService()
{
    BCLDebug.Trace("REMOTE", "ObjectHandle.InitializeLifetimeService");

    //
    // If the wrapped object has implemented InitializeLifetimeService to return null,
    // we don't want to go to the base class (which will result in a lease being
    // requested from the MarshalByRefObject, which starts up the LeaseManager,
    // which starts up the ThreadPool, adding three threads to the process.
    // We check if the wrapped object is a MarshalByRef object, and call InitializeLifetimeServices on it
    // and if it returns null, we return null. Otherwise we fall back to the old behavior.
    //

    MarshalByRefObject mbr = WrappedObject as MarshalByRefObject;
    if (mbr != null) {
        Object o = mbr.InitializeLifetimeService();
        if (o == null)
            return null;
    }
    ILease lease = (ILease)base.InitializeLifetimeService();
    return lease;
}
```

.... so **it may indeed return null**.

Maybe 

https://github.com/serilog/serilog/blob/8528ae34434a51db2f3494643891ca6abcc87a4b/src/Serilog/Context/LogContext.cs#L254-L259

could be changed to become: 

```csharp
public override object InitializeLifetimeService()
{
    var lease = base.InitializeLifetimeService() as ILease;
    lease?.Register(LifeTimeSponsor);
    return lease;
}
```

nblumhardt:
That would be a fair change. I don't think we ever create a handle to a `MarshalByRefObject` here, but the clarity would be improved anyway.

Will merge as-is so that we don't drift into merge-conflict territory in the meantime :-)

Issue #1043, opened 2017-10-16
Granular configuration using Expression Trees


tsimbalar:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [x] New Feature

This issue proposes an extension on top of the initial proposal described in #1038 .

The objective with this one is to be able to define a `LoggerConfiguration` using the existing fluent API, but still being able to override some specific parameters with key-value settings from another source.

The proposed implementation would look like this :

```csharp
// supposedly, appsetting contains 
// <add key="serilog:write-to:RollingFile.pathFormat" value="anotherPath.log" />
// <add key="serilog:minimum-level:override:Microsoft" value="Warning" />
var logger = new LoggerConfiguration()
    .ReadFrom.Combined(x => x
        .AddExpression(lc => lc  
            .MinimumLevel.Debug
            .WriteTo.RollingFile("default-path-{Date}.txt"))
        .AddAppSettings())  
    .CreateLogger();

// AddExpression takes as parameter a : 
// Expression<Func<LoggerConfiguration, LoggerConfiguration>>
```

that is, a new extension method for `ISettingsSourceBuilder` that accepts a parameter of type `Expression<Func<ISettingsSourceBuilder, ISettingsSourceBuilder>>`.

That extension method would analyze the method calls, and generate the correspounding *key-value* settings that would allow to accomplish such a configuration.

**Limitations**
*Expression-trees* have a few limitations (see https://github.com/dotnet/roslyn/issues/2060 ), among others : 
- you cannot call methods using named arguments (only using positional arguments work)
- you cannot call methods using *default* arguments (you have to call the method passing the argument event if it is the default value)

At least in the first version, I think **not all** configuration methods will / should be handled ... 

### Supported features
Here is what I think makes sense to implement based on my personal usage of Serilog, or the technical limitations (and the fact that they translate rather well to the stringy key-value settings) :
- the `MinimumLevel.Debug()`, `MinimumLevel.Information()` etc family of methods 
- the `MinimumLevel.Override("System", LogEventLevelWarning)`
- `Enrich.WithProperty(propName, propValue)`
-  `Enrich.WithExtensionMethod()` , that is also work with extension methods from *Serilog* or *external* (`Enrich.WithThreadId()`, `Enrich.FromLogContext()` etc)
- *Sinks* : all the `WriteTo.Something(...)` or `AuditTo.SomeThing()` should work

### Unsupported features
Here is what I think we may not need to implement :
[...]

PR #1044, opened 2017-10-16
Experiments/readfrom expression

tsimbalar:
issue #1043 itself related to #1038 

no, only new APIs

**Other information**:
First implementation of *expression trees* as a source of *key-value* settings.
It is based on the commits currently in #1041


tsimbalar:
This was a bit too early for that, I think :) 
I'll close this PR and open the new one when we know a bit better what the "combined" settings scenario will look like

PR #1047, opened 2017-10-19
Hardening of KeyValuePairSettings

tsimbalar:
Harden / Refactor `KeyValuePairSettings` as a preparation for adding extensions on top of it

- Made KeyValuePairSettings consume the `IEnumerable<KeyValuePair<string,string>>` only when `Configure` is called
- Ignored KeyValuePairs with a `null` key
- Added tests about exception that can be thrown

No

**Other information**:

### DIFF
--- a/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
+++ b/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
@@ -164,6 +164,7 @@ namespace Serilog.Settings.KeyValuePairs
         internal static IReadOnlyDictionary<string, string> ExtractDirectives(IEnumerable<KeyValuePair<string, string>> keyValuePairs)
         {
             var directives = keyValuePairs
+                .Where(kvp => kvp.Key != null)
                 .Where(kvp => _supportedDirectives.Any(kvp.Key.StartsWith))
                 .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
 
--- a/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
+++ b/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
@@ -83,11 +83,8 @@ namespace Serilog.Settings.KeyValuePairs
         public void Configure(LoggerConfiguration loggerConfiguration)
         {
             if (loggerConfiguration == null) throw new ArgumentNullException(nameof(loggerConfiguration));
-
-            var settings = _settings.ToDictionary(s => s.Key, s => s.Value);
-            var directives = settings.Keys
-                .Where(k => _supportedDirectives.Any(k.StartsWith))
-                .ToDictionary(k => k, k => settings[k]);
+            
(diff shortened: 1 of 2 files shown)


nblumhardt:
I'm not sure about making the use of the `IEnumerable` lazy, I have tended to go the other direction anywhere it's possible.

```csharp
var pairs = xxx;
var settings = new KeyValueSettings(pairs);

// Later somewhere...
pairs.Add("A", "b");

// Later still somewhere else...
settings.Configure(loggerConfiguration);
```

It's no less correct, but it seems to introduce some unpredictability that isn't present in the current code. The Serilog codebase currently considers any `IEnumerable` passed to a method or constructor to be valid only during the call, and copies it if the values are needed later on (there may be some rare perf-sensitive cases where this isn't true ;-))

All things being equal I'd tend to leave this one, but interested to hear your thoughts if there's an advantage I'm missing in the alternative approach. Cheers!

tsimbalar:
Gosh, I thought I had lots of super-convincing arguments about this one ... but it turns out that I can't come up with any good one right now 😀 

The only thing I have is an opinion that accepting a `IEnumerable` as a parameter is a bit deceptive in that case : 
- technically it means I could pass in an *infitinite* enumerable (but who would do that? 😛 )
- also the key-value pairs are checked for unicity straight away by `ToDictionary()`, so I would expect the constructor to only accept a `Dictionary` in here to respect the *principle of least astonishment*

(same for `ReadFrom.KeyValuePairs()` I guess)

So anyway, would it be OK to:
- change the signature of the *constructor* from `KeyValuePairSettings(IEnumerable<KeyValuePair<string, string>>)` to `KeyValuePairSettings(Dictionary<string, string>)`
- to the `enumerable.ToDictionary()` in the `ReadFrom.KeyValuePairs()` method
?

(*no* is a perfectly fine answer ;) )

nblumhardt:
I think that makes sense - though perhaps `IReadOnlyDictionary<string, string>` rather than the concrete `Dictionary` (guessing this is what you mean).

Another option we could consider is to allow duplicates in the keys, but have later values override earlier ones. It doesn't win anything in the current implementation, but would be consistent with how I imagine "overrides" might work in _Serilog.Settings.Combined_ ...?

tsimbalar:
> I think that makes sense - though perhaps IReadOnlyDictionary<string, string> rather than the concrete Dictionary (guessing this is what you mean).

Actually I meant `IDictionary<string, string>`, but `ReadOnlyDictionary` sounds good ... and well, the class is `internal` so I guess we are free to do one thing and change it later. 

> allow duplicates in the keys, but have later values override earlier ones.

I have mixed feelings about this .... I think the class `KeyValuePairSettings` is already doing quite some work :) And I believe the setting combination should happen somewhere else so that we don't *cut off* some scenarios where *the latest wins* is not valid anymore.

nblumhardt:
:+1: I'm fine with that, sounds good :-)

tsimbalar:
I created PR #1051 instead. 
Closing this one.

PR #1050, opened 2017-10-23
Fix last ReSharper warning

tsimbalar:
Fix the last ReSharper warning that was left after #1042 according to discussion https://github.com/serilog/serilog/pull/1042#issuecomment-337764490


nblumhardt:
👍

PR #1051, opened 2017-10-23
KeyValuePairSettings - IReadOnlyDictionary / latest setting win

tsimbalar:
new PR based on discussion https://github.com/serilog/serilog/pull/1047#issuecomment-338544876 .

Make `KeyValuePairSettings` accept a `IReadOnlyDictionary<string, string>` instead of a `IEnumerable<KeyValuePair<string, string>>` to make it more obvious that : 
- it is not consuming the key-value pairs lazily
- it expects unique keys
- it will not touch the contents of the provided dictionary

This has no impact on the public API given that class `KeyValuePairSettings` is `internal`.

This PR also adds a few tests around the edge cases

### DIFF
--- a/src/Serilog/Configuration/LoggerSettingsConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSettingsConfiguration.cs
@@ -14,7 +14,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.Linq;
 using Serilog.Settings.KeyValuePairs;
 
--- a/src/Serilog/Configuration/LoggerSettingsConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSettingsConfiguration.cs
@@ -50,11 +50,21 @@ namespace Serilog.Configuration
         /// </summary>
         /// <param name="settings">A list of key-value pairs describing logger settings.</param>
         /// <returns>Configuration object allowing method chaining.</returns>
+        /// <remarks>In case of duplicate keys, the last value for the key is kept and the previous ones are ignored.</remarks>
         public LoggerConfiguration KeyValuePairs(IEnumerable<KeyValuePair<string, string>> settings)
         {
             if (settings == null) throw new ArgumentNullException(nameof(settings));
+            var uniqueSettings = new Dictionary<string, string>();
+            foreach (var kvp in settings)
+            {
+                uniqueSettings[kvp.Key] = kvp.Value;
+            }
+            return KeyValuePairs(uniqueSettings);
+        }
 
-            return Settings(new KeyValuePairSettings(settings.ToDictionary(x => x.Key, x => x.Value)));
+        LoggerConfiguration KeyValuePairs(IReadOnlyDictionary<string, string> settings)
+        {
+            return Settings(new KeyValuePairSettings(settings));
         }
     }
 }
--- a/src/Serilog/Configuration/LoggerSettingsConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSettingsConfiguration.cs
@@ -14,6 +14,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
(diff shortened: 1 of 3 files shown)


src/Serilog/Configuration/LoggerSettingsConfiguration.cs
```diff
@@ -52,7 +54,9 @@ public LoggerConfiguration Settings(ILoggerSettings settings)
         public LoggerConfiguration KeyValuePairs(IEnumerable<KeyValuePair<string, string>> settings)
         {
             if (settings == null) throw new ArgumentNullException(nameof(settings));
-            return Settings(new KeyValuePairSettings(settings));
+
+            var settingsDictionary = new ReadOnlyDictionary<string, string>(settings.ToDictionary(x => x.Key, x => x.Value));
+            return Settings(new KeyValuePairSettings(settingsDictionary));
```
nblumhardt: I think `Dictionary` implements `IReadOnlyDictionary` in recent framework versions - I don't think the creation of the `ReadOnlyDictionary` should be needed here.

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
```diff
@@ -164,12 +161,21 @@ select new
             }
         }
 
+        internal static IReadOnlyDictionary<string, string> ExtractDirectives(IReadOnlyDictionary<string, string> settings)
+        {
+            var directives = settings
+                .Where(kvp => _supportedDirectives.Any(kvp.Key.StartsWith))
+                .ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
+
+            return new ReadOnlyDictionary<string, string>(directives);
```
nblumhardt: Also here, if we don't create a `ReadOnlyDictionary` explicitly, we can get this back to a one-liner and avoid the additional method.

(Nitpick - the codebase doesn't use explicit `internal` modifiers elsewhere.)

tsimbalar: Actually the method exists for the sake of tests ... but I don't know if the test 

```csharp
+        [Fact]
 +        public void IrrelevantKeysAreIgnored()
 +        {
 +            var irrelevantSettings = new Dictionary<string, string>()
 +            {
 +                ["whatever:foo:bar"] = "willBeIgnored",
 +                ["irrelevant"] = "willBeIgnored",
 +            };
 +
 +            var directives = KeyValuePairSettings.ExtractDirectives(irrelevantSettings);
 +
 +            Assert.False(directives.Any());
 +        }
```

provides much value ...

(about `internal` : dammit, I thought I had removed them all !)

src/Serilog/Configuration/LoggerSettingsConfiguration.cs
tsimbalar: Ah yes, that's true. It's one of those strange things where a *writable* dictionary actually implements *read-only* dictionary .... 

I'm not sure what the "best practices" are in that regard, but I guess in internal classes there is no need to be a control-freak :) 

I'll change that.

src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
nblumhardt: Makes sense, I didn't notice the test dependency, thanks 👍

tsimbalar:
`ReadFrom.KeyValuePairs(IEnumerable<KeyValuePair<string, string>>)` now keeps the last one in case of duplicate keys. This will allow *Serilog.Settings.Combined* to rely on that behavior.

Maybe it should also be mentioned in the documentation somewhere ... (or included in  #1053 )

For now, I have left an overload `ReadFrom.KeyValyePairs(IReadOnlyDictionary<string, string>)` as `private` (I'm trying to resist the itch to make it `public` :p )

nblumhardt:
👍

PR #1059, opened 2017-11-08
Add support for abstract class parameters in settings

### LINKED ISSUE #1056: KeyValuePairSettings : support a fully qualified type name for arguments whose type is abstract
For symmetry with serilog/serilog-settings-configuration#71 
For consistency with the way `interface`s parameter are handled

For parameters of configuration methods whose type is an `abstract class`, the key-value settings format should accept as value the fully-qualified type name of a subclass of this abstract class that has a default constructor.

The typical use-case for this enhancement is *Serilog.Sinks.Console* whose configuration method looks like this : 

```csharp
public static LoggerConfiguration Console(
	this LoggerSinkConfiguration sinkConfiguration,
	LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum,
	string outputTemplate = DefaultConsoleOutputTemplate,
	IFormatProvider formatProvider = null,
	LoggingLevelSwitch levelSwitch = null,
	LogEventLevel? standardErrorFromLevel = null,
	ConsoleTheme theme = null)
```
where `ConsoleTheme` is defined as `public abstract class ConsoleTheme` .

Given a class 
```csharp
namespace MyCustom.NameSpace
{
    public class MyCustomTheme:ConsoleTheme
    {
        public MyCustomTheme(){
        }

        // snip ...
    }
}
```

We should be able to configure the console sink like so : 

```xml
<configuration>
  <appSettings>
    <add key="serilog:using:Console" value="Serilog.Sinks.Console" />
    <add key="serilog:write-to:Console.theme" value="MyCustom.NameSpace.MyCustomTheme" />
```

I'm not sure how "qualified" the type name needs to be here, or whether adding `using:` statements are needed, but the behavior should be similar to the way interfaces are processed

( serilog/serilog-sinks-console#28 seems to mean that name must be *assembly-qualified* at least in *Serilog.Settings.Configuration* )

tsimbalar:
When a parameter is an abstract class, now accept the type name of a concrete subclass that has a default constructor ( the same way interfaces are supported)

Resolves #1056 
This is the equivalent of serilog/serilog-settings-configuration#74

Nope

**Other information**:

### DIFF
--- a/src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
+++ b/src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
@@ -51,7 +51,7 @@ namespace Serilog.Settings.KeyValuePairs
             if (convertor != null)
                 return convertor(value);
 
-            if (toTypeInfo.IsInterface && !string.IsNullOrWhiteSpace(value))
+            if ((toTypeInfo.IsInterface || toTypeInfo.IsAbstract) && !string.IsNullOrWhiteSpace(value))
             {
                 var type = Type.GetType(value.Trim(), throwOnError: false);
                 if (type != null)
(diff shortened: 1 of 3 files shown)


skomis-mm:
👍

nblumhardt:
👏

PR #1063, opened 2017-11-15
Move LoggerExtension so that namespace and folder matches

### LINKED ISSUE #1054: LoggerExtensions.cs is in the wrong source folder
https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/LoggerExtensions.cs

Does not match namespace, should be in the root of the project.

pardahlman:
This PR fixes #1054

No, just moved file to match namespace

- [x ] The commit follows our [guidelines](https://github.com/serilog/serilog/blob/dev/CONTRIBUTING.md)
- [ x] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:
My apologies for placing the file in the wrong directory. I noticed that other LoggerExtensions (sinks repos etc) used `Serilog` as namespace, so I thought it was a convention to allow for file location and namespace to mismatch for these classes. I had a closer look and realized that the sinks use `Serilog` as root namespace :flushed:


merbla:
LGTM 👍

PR #1064, opened 2017-11-17
Add support for reference to public static properties in settings

### LINKED ISSUE #1057: KeyValuePairSettings : support for static members via `::`
Related to discussion going on in serilog/serilog-sinks-console#16

It would be nice to support **static members** for configuration methods that accept a complex type .

The typical use case is *Serilog.Sinks.Console* with its configuration method : 
```csharp
public static LoggerConfiguration Console(
	this LoggerSinkConfiguration sinkConfiguration,
	LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum,
	string outputTemplate = DefaultConsoleOutputTemplate,
	IFormatProvider formatProvider = null,
	LoggingLevelSwitch levelSwitch = null,
	LogEventLevel? standardErrorFromLevel = null,
	ConsoleTheme theme = null)
```
where `ConsoleTheme` is an `abstract class`, but static members exist (`ConsoleTheme.None`, `SystemConsoleTheme.Literate`, `AnsiConsoleTheme.Grayscale` [etc](https://github.com/serilog/serilog-sinks-console#themes) )

One possible syntax that would be do access static members via `::` (like in [Powershell](https://docs.microsoft.com/en-us/powershell/scripting/getting-started/cookbooks/using-static-classes-and-methods?view=powershell-5.1#displaying-static-properties-of-systemenvironment) or [PHP](http://php.net/manual/en/language.oop5.static.php#language.oop5.static.properties)), for instance 
```xml
<configuration>
  <appSettings>
    <add key="serilog:using:Console" value="Serilog.Sinks.Console" />
    <add key="serilog:write-to:Console.theme" value="Serilog.Sinks.SystemConsole.Themes.AnsiConsoleTheme::Grayscale, Serilog.Sinks.Console" />
```

would use `static` property `Grayscale` of type `Serilog.Sinks.SystemConsole.Themes.AnsiConsoleTheme` in assembly `Serilog.Sinks.Console`.

As in #1056 the idea is that we should be able to specify the minimum information about the type to make the configuration less verbose (for instance, no need for the assembly name, Version, PublicKeyToken etc if we can avoid it).

Other use-case where this would help : *Serilog.Sinks.Email* that accepts a `EmailConnectionInfo` for advanced configuration

tsimbalar:
as a value for parameters, using the syntax `NameSpace.To.ConcreteType::StaticProperty, AssemblyName`

resolves #1057 

It shouldn't, although technically, there is a chance that "string" settings whose value looks like `"Namespace.To.TypeName::PropertyName"` may now be interpreted as a reference to a static property, even though I've done my best to avoid ambiguities.  

**Other information**:
When / if accepted, similar code should be needed in *Serilog.Settings.Configuration* to handle issue serilog/serilog-settings-configuration#73 .

### DIFF
--- a/src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
+++ b/src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
@@ -16,11 +16,17 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
 
 namespace Serilog.Settings.KeyValuePairs
 {
     class SettingValueConversions
     {
+        // should match "The.NameSpace.TypeName::MemberName" optionnally followed by
+        // usual assembly qualifiers like : 
+        // ", MyAssembly, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"
+        static Regex StaticMemberAccessorRegex = new Regex("^(?<shortTypeName>[^:]+)::(?<memberName>[A-Za-z][A-Za-z0-9]*)(?<typeNameExtraQualifiers>[^:]*)$");
+
         static Dictionary<Type, Func<string, object>> ExtendedTypeConversions = new Dictionary<Type, Func<string, object>>
             {
                 { typeof(Uri), s => new Uri(s) },
@@ -29,6 +35,22 @@ namespace Serilog.Settings.KeyValuePairs
 
         public static object ConvertToType(string value, Type toType)
         {
+            if (TryParseStaticMemberAccessor(value, out var accessorTypeName, out var memberName))
+            {
+                var accessorType = Type.GetType(accessorTypeName, throwOnError: true);
+                var publicStaticPropertyInfo = accessorType.GetTypeInfo().DeclaredProperties
+                    .Where(x => x.Name == memberName)
+                    .Where(x => x.GetMethod != null)
+                    .Where(x => x.GetMethod.IsPublic)
+                    .FirstOrDefault(x => x.GetMethod.IsStatic);
+
+                if (publicStaticPropertyInfo == null)
+                {
+                    throw new InvalidOperationException($"Could not find public static property `{memberName}` on type `{accessorTypeName}`");
+                }
+                return publicStaticPropertyInfo.GetValue(null); // static property, no instance to pass
+            }
+
             var toTypeInfo = toType.GetTypeInfo();
             if (toTypeInfo.IsGenericType && toType.GetGenericTypeDefinition() == typeof(Nullable<>))
             {
@@ -71,6 +93,30 @@ namespace Serilog.Settings.KeyValuePairs
             }
 
             return Convert.ChangeType(value, toType);
+        }
+
+        internal static bool TryParseStaticMemberAccessor(string input, out string accessorTypeName, out string memberName)
+        {
+            if (input == null)
+            {
+                accessorTypeName = null;
+                memberName = null;
+                return false;
+            }
+            if (StaticMemberAccessorRegex.IsMatch(input))
+            {
+                var match = StaticMemberAccessorRegex.Match(input);
+                var shortAccessorTypeName = match.Groups["shortTypeName"].Value;
+                var rawMemberName = match.Groups["memberName"].Value;
+                var extraQualifiers = match.Groups["typeNameExtraQualifiers"].Value;
+
+                memberName = rawMemberName.Trim();
+                accessorTypeName = shortAccessorTypeName.Trim() + extraQualifiers.TrimEnd();
+                return true;
+            }
+            accessorTypeName = null;
+            memberName = null;
+            return false;
         }
     }
 }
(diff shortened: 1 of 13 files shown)


src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
```diff
@@ -29,6 +35,22 @@ class SettingValueConversions
 
         public static object ConvertToType(string value, Type toType)
         {
+            if (TryParseStaticMemberAccessor(value, out var accessorTypeName, out var memberName))
```
nblumhardt: Perhaps this could move down to the `if ((toTypeInfo.IsInterface || toTypeInfo.IsAbstract) && !string.IsNullOrWhiteSpace(value))`  block? If it's first up, here, we might hit issues where this is ambiguous, e.g. the target `Type` is `string`.

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
```diff
@@ -29,6 +35,22 @@ class SettingValueConversions
...
+            {
+                var accessorType = Type.GetType(accessorTypeName, throwOnError: true);
```
skomis-mm: Probably, for a better UX could we use full type names without specifying assembly? We could look for the target type in configuration assemblies (`using` section) something like:
```csharp
var accessorType = Type.GetType(typeName) ??
            _assemblies.SelectMany(a => a.ExportedTypes)
                        .FirstOrDefault(t => t.FullName == typeName);
```

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
```diff
@@ -29,6 +35,22 @@ class SettingValueConversions
...
+                var publicStaticPropertyInfo = accessorType.GetTypeInfo().DeclaredProperties
+                    .Where(x => x.Name == memberName)
```
skomis-mm: Perhaps we could also add support for the public static fields at the same time?

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
nblumhardt: If we do that for members, could/should we do the same thing for instances?

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
tsimbalar: That was my initial implementation, but then I thought it would be strange to limit it to only `abstract class`es and `interface`s, because after all this can also be useful in other cases, like for sinks that accept *complex* parameters that are not `abstract` nor `interface` (for instance : `EmailConnectionInfo` in *Serilog.Sinks.Email* , `ColumnOptions` for *Serilog.Sinks.MSSqlServer* ...) 

This also gives a workaround for configuration that canot be easily defined with the keay-value settings syntax, by allowing to define it as code in a static property.

Also, I've tried to make the *regex* for static-property accessor as specific as possible to avoid false positives.  

What do you think ?

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
tsimbalar: I have mixed feelings about accessing arbitrary properties, but I guess if they are `public`, they are part of the public API so technically accessible. I'll include public properties as well

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
tsimbalar: For now I've tried to not deviate too much from the way arbitrary sub-types (for interfaces and abstract classes) are referenced, that is with full assembly-qualified type names. Maybe lifting that restriction would be nice, but I would keep it as a separate issue / PR.

nblumhardt: Sounds good

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
tsimbalar: @nblumhardt any opinion on the subject? :)

nblumhardt: I think it's a fair consideration - just not sure if it's behavior we want to commit to yet in this PR. (We'd also then need to consider whether assembly-qualified type names should be supported in more places, since subclassing is a possible workaround for other scenarios, plus look at how this would interact with some kind of "structure" syntax for populating fields of complex configuration parameters down the track).

TL;DR: I don't think it's a bad idea - just might be scope creep for this PR, since we haven't really dug through all the possible implications :-) .. what do you think?

tsimbalar: That makes sense 👍 
Easier to make it support a few use cases initially and extend it later on if needed.

Thanks !

tsimbalar:
I've updated my PR with the suggested changes.

skomis-mm:
👍

PR #1068, opened 2017-11-27
Fixes #1067

### LINKED ISSUE #1067: Using int.Parse is unsafe
For example this template message "{fdsf,4611686018427387903:sdf}" throws here with integer overflow.

https://github.com/serilog/serilog/blob/f284f85ce585fe59c26fc349df3043c90fb542b8/src/Serilog/Parsing/MessageTemplateParser.cs#L137

nikolaybobrovskiy:
Fixes #1067


nblumhardt:
Thanks @nikolaybobrovskiy, looks good 👍

PR #1091, opened 2018-01-23
Improvements to {Properties} handling in output templates

nblumhardt:
Fixes serilog/serilog-sinks-file#45 - support {Properties:j} in output templates; fixes #1088 - handle positional properties when computing {Properties}.

Brings the built-in message template formatting in line with _Serilog.Sinks.Console_.

Originally, `:j` wasn't supported on `{Properties}` because the formatting APIs here required some additional allocations in order to implement it, but in retrospect, better to be consistent.

PR #1075, opened 2017-12-29
#1073 Overload Wrap() to accept LogEventLevel and LoggingLevelSwitch.

joostas:
**What issue does this PR address?**
#1073.

**Does this PR introduce a breaking change?**
No

**Please check if the PR fulfills these requirements**
- [x] The commit follows our [guidelines](https://github.com/serilog/serilog/blob/dev/CONTRIBUTING.md)
- [x] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:

PR #1094, opened 2018-01-28
Release 2.7.1

nblumhardt:
* #1075 - overload of `LoggerSinkConfiguration.Wrap()` accepting `LoggingLevelSwitch`
 * #1083 - update package metadata, including `https://` project and icon URLs
 * #1091 - make `{Properties:j}` work consistently with the console sink
 * #1110 - expose an efficient `Logger.None`
 * #1113 - (tooling) improve tests for `LogContext`
 * #1136 - (tooling) fix `build.sh` exit code
 * #1141 - fix logging of `byte[]` smaller than 1 KB
 * #1157 - (tooling) update _.gitignore_
 * #1158 - (tooling) re-enable macOS builds
 * #1163 - SourceLink v2 support
 * #1165 - fix a number of issues with `LoggerSinkConfiguration.Wrap()` affecting `WriteTo.Async()`
 * #1134 - (tooling) add README badges for downloads/stats
 * #1166 - allow `LoggerConfiguration` to be garbage collected after logger is created
 * #1160 - (tooling) update NuGet.org API key
 * (tooling) Update to use the latest .NET SDK


tsimbalar:
I'm not sure why, but the current pre-release (dev) nuget packages are versioned v2.7.1 ... so I guess this release should be v2.7.1, to avoid going through decreasing versions ?

adamchester:
Why isn't the Travis build status part of the github checks here?

nblumhardt:
I've updated the release notes above and think we've got enough here to warrant merging/moving forward.

Has anyone had an opportunity to put 2.7-dev through its paces in a more demanding setting? I should have some time in the next few days to update various apps, but it'd be great if more people can smoke test, too.

tinytech-delivery:
LGTM, as noted I guess this will be `2.7.1.x`

merbla:
LGTM, as noted it will be `2.7.1` and with existing packages on [NuGet](https://www.nuget.org/packages/Serilog/) not sure it is worth bumping it back.

nblumhardt:
This is now published! 🎉 

One small caveat encountered while updating _Serilog.Sinks.Async_ post-publishing, the implicit reference to `NETStandard.Library` that's now used for the .NET Standard targets implicitly bumps the versions of framework dependencies up. The change will fix DLL hell for some, and require others to upgrade or explicitly specify dependency versions. The world we live in...

Binary compatibility is not impacted, and neither are .NET Framework builds, so fingers crossed this'll all come out fine 😅 .

tsimbalar:
Updated in https://github.com/tsimbalar/serilog-settings-comparison , all tests are green (not that I expected otherwise :P ).

tsimbalar:
Also, it turns out this release was [**build number 999**](https://ci.appveyor.com/project/serilog/serilog/build/999) for AppVeyor ... I hope the next build will not break everything *à la* Y2K bug :)

PR #1108, opened 2018-02-18
Fixes #1105 - adds static Log.None property with appropriate tests.  …

### LINKED ISSUE #1105: Expose `SilentLogger` as `Logger.None`
- [ ] Bug
- [x ] New Feature

2.6.0

**What is the target framework and operating system? See [target frameworks]
- [x ] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

There is no mechanism for creating a Serilog.ILogger that does nothing.

I sometimes use an optional pattern for logging in libraries such as:

````csharp
        public SomeClass(ILogger logger = null)
        {
            m_logger = (logger ?? new SilentLogger).ForContext<SomeClass>();
        }
````

This allows consumers of the class that aren't using logging to use the abstraction without logging, while those that are can pass in a logger without having all log statements require a condition method operator.  Being able to instantiate a SilentLogger makes this cheap.  

Is there a better pattern with Serilog for accomplishing this?

### PR DESCRIPTION
**Pull request by IanYates**
Adds a static `Log.None` property as per the discussion in #1105 

No

**Other information**:
Basic change to stash a `SilentLogger` instance in a private static variable on the `Log` class. That instance is used as the initial value of the `Logger` property.  The same instance is then exposed via the `None` property. `CloseAndFlush()` was modified to reset back to this instance rather than creating yet another `SilentLogger` instance.

Comments were updated where appropriate and code style was preserved.

I've added tests to ensure that the `None` property is initialised properly.  I've also ensured via a test that initialisation of the `Logger` property doesn't affect the `None` property. 
Finally I've ensured that `CloseAndFlush` doesn't corrupt the logger stored and accessible via the `None` property by checking that it remains a `SilentLogger` instance and that we can invoke it without raising an exception.

I think there are lines of words here than lines of code changed in the PR.  All tests (on my machine) seem to pass in the various framework versions.

Thanks 💯

### DIFF
--- a/src/Serilog/Log.cs
+++ b/src/Serilog/Log.cs
@@ -40,7 +40,15 @@ namespace Serilog
     /// </remarks>
     public static class Log
     {
-        static ILogger _logger = new SilentLogger();
+        /// <summary>
+        /// Initialise a SilentLogger to be the constant value of _noneLogger
+        /// </summary>
+        static readonly ILogger _noneLogger = new SilentLogger();
+
+        /// <summary>
+        /// Use the constant _noneLogger as the initial value
+        /// </summary>
+        static ILogger _logger = _noneLogger;
 
         /// <summary>
         /// The globally-shared logger.
@@ -61,7 +69,7 @@ namespace Serilog
         /// </summary>
         public static void CloseAndFlush()
         {
-            ILogger logger = Interlocked.Exchange(ref _logger, new SilentLogger());
+            ILogger logger = Interlocked.Exchange(ref _logger, _noneLogger);
 
             (logger as IDisposable)?.Dispose();
         }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by IanYates** [comment]
Hmmm - AppVeyor failed in an area that I don't *think* I affected.  `ContextPropertiesCrossAsyncCalls`...  Travis CI seems happy (as did my PC).

**Comment by nblumhardt** [comment]
Thanks, Ian!

Since this is more of a plumbing feature, than something that will be used regularly by callers of `Log`, I'd thought perhaps `Logger.None` rather than `Log.None` might keep it tucked away?

File: src/Serilog/Log.cs [hunk 1]
```diff
@@ -1201,5 +1209,16 @@ public static bool BindProperty(string propertyName, object value, bool destruct
         {
             return Logger.BindProperty(propertyName, value, destructureObjects, out property);
         }
+
+        /// <summary>
+        /// An <see cref="ILogger"/> instance that silently ignores all log messages
```
nblumhardt: Nitpick, could use a `.` here.

Perhaps we could try `"that efficiently ignores all method calls"`?

IanYates: 👍 
Shall do

IanYates:
Not sure why I ended up on Log.None...  Logger.None makes more sense.  Shall amend this evening. Thanks

IanYates:
Closing in favour of PR #1110 instead

PR #1110, opened 2018-02-19
Fixes #1105 - add a static None property to the Logger class

### LINKED ISSUE #1105: Expose `SilentLogger` as `Logger.None`

IanYates:
Addresses issue #1105.  Replaces PR #1108 

No

**Other information**:
Sorry for the new PR.  Given the changes in the old PR were in the completely wrong file, it seemed easier to just close that PR and open a new one.

Issue #1105 asked for an easy way to get at a constant instance of a `SilentLogger`. The original PR added this on the `Log` class and slightly adjusted that class' internals to make use of this instance.
This new PR instead modifies the `Logger` class to add a static `Logger.None` property that returns the same `SilentLogger`.

One new allocation overall.  I suspect we can live with that (?).  Former PR didn't have that issue.  A lazy init sounds like overkill for this but I'm happy to be told I'm wrong.

Only one new unit test - verifying that the property is a silent logger.  There's not a lot of machinery or moving parts to verify anymore compared with the earlier PR.

Feedback most welcome. Thanks 💯

### DIFF
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -1373,21 +1373,10 @@ namespace Serilog.Core
         {
             _dispose?.Invoke();
         }
-
+        
         /// <summary>
-        /// Backing property for the exposed None property
+        /// An <see cref="ILogger"/> instance that efficiently ignores all method calls.
         /// </summary>
-        static readonly ILogger _noneLogger = new SilentLogger();
-
-        /// <summary>
-        /// An <see cref="ILogger"/> instance that efficiently ignores all method calls
-        /// </summary>
-        public static ILogger None
-        {
-            get
-            {
-                return _noneLogger;
-            }
-        }
+        public static ILogger None { get; } = new SilentLogger();
     }
 }
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -16,6 +16,7 @@ using System;
 using System.Collections.Generic;
 using Serilog.Capturing;
 using Serilog.Core.Enrichers;
+using Serilog.Core.Pipeline;
 using Serilog.Debugging;
 using Serilog.Events;
 
@@ -1371,6 +1372,22 @@ namespace Serilog.Core
         public void Dispose()
         {
             _dispose?.Invoke();
+        }
+
+        /// <summary>
+        /// Backing property for the exposed None property
+        /// </summary>
+        static readonly ILogger _noneLogger = new SilentLogger();
+
+        /// <summary>
+        /// An <see cref="ILogger"/> instance that efficiently ignores all method calls
+        /// </summary>
+        public static ILogger None
+        {
+            get
+            {
+                return _noneLogger;
+            }
         }
     }
 }
--- a/test/Serilog.Tests/Core/LoggerTests.cs
+++ b/test/Serilog.Tests/Core/LoggerTests.cs
@@ -125,5 +125,13 @@ namespace Serilog.Tests.Core
         {
             Assert.IsType<SilentLogger>(Logger.None);
         }
+
+        [Fact]
+        public void TheNoneLoggerIsAConstant()
+        {
+            var firstCall = Logger.None;
+            var secondCall = Logger.None;
+            Assert.Equal(firstCall, secondCall);
+        }
 	}
 }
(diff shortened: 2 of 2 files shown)


src/Serilog/Core/Logger.cs
```diff
@@ -1372,5 +1373,21 @@ public void Dispose()
...
+
+        /// <summary>
+        /// Backing property for the exposed None property
+        /// </summary>
+        static readonly ILogger _noneLogger = new SilentLogger();
+
+        /// <summary>
+        /// An <see cref="ILogger"/> instance that efficiently ignores all method calls
+        /// </summary>
+        public static ILogger None
```
nblumhardt: We could get away without a backing property by declaring this as:

```csharp
public static ILogger None { get; } = new SilentLogger();
```

src/Serilog/Core/Logger.cs
```diff
@@ -1372,5 +1373,21 @@ public void Dispose()
...
             _dispose?.Invoke();
         }
+
+        /// <summary>
+        /// Backing property for the exposed None property
+        /// </summary>
+        static readonly ILogger _noneLogger = new SilentLogger();
+
+        /// <summary>
+        /// An <see cref="ILogger"/> instance that efficiently ignores all method calls
```
nblumhardt: Nitpick - missing `.`.

nblumhardt:
Not sure why the AppVeyor CI build has started producing flaky results around `LogContext` - will keep an eye on it, but pretty sure it's not related to this PR :+1:

src/Serilog/Core/Logger.cs
IanYates: No worries - wasn't sure if you were happy with newer syntax

tsimbalar:
Maybe some unit tests to add : 
- Asking for Logger.None twice returns the same instance (i.e. it's not a `new`each time)
- the result of `new LoggerConfiguration().CreateLogger()` is `Logger.None`

IanYates:
@tsimbalar - I've addressed your first suggestion.
The second though is a little more work and may warrant some discussion with @nblumhardt first.  

This test
```

        [Fact]
        public void EmptyLoggerConfigurationCreatesNoneLogger()
        {
            var emptyConfigLogger = new LoggerConfiguration().CreateLogger();
            var noneLongger = Logger.None;

            Assert.Equal(emptyConfigLogger, noneLongger);
        }
```

fails.

We could modify `CreateLogger()` to see that there's no zero configuration and deliberately return Logger.None.  It doesn't do that now.  `SilentLogger` itself is barely used in the Serilog codebase.  Certainly there's no way you could get one from `CreateLogger()` as the code stands now  (which surprised me given the chat in issue #1105)

@nblumhardt - should we do this at all?  If so, want it in a separate PR?  I've only had a cursory look at LoggerConfiguration.cs and I can think of a couple of approaches - one to check for any methods/setters being called and set a flag so we know not to short-circuit return Logger.None.  The other, which is less invasive but may not be quite correct, would be to have `CreateLogger()` see if everything is still at the defaults (empty sinks list, etc), and if so, just return Logger.None.

Con for doing it...  If you really want a silent logger, use the .None property rather than the hacky 'empty configuraton' approach.

tsimbalar:
@IanYates oh my bad, I thought that was the actual behavior ... I may be mixed up ... The test that should pass as of now, though, is that `Log.Logger` should be a `SilentLogger` by default, if not assigned otherwise ... should it actually be `Logger.None` ?

nblumhardt:
Great suggestions @tsimbalar 👍 

In v1.x, `CreateLogger()` did return the `SilentLogger`. But, in v2, we switched the return type of `CreateLogger` from `ILogger` to `Logger` so that the need to dispose the result was communicated by the API.

I think this is ready to go; no idea why the CI build is suddenly unhappy, but guess that's something to figure out another day :-)

Issue #1112, opened 2018-02-20
LogContext CI test failure


nblumhardt:
From #1110; recently a test for `LogContext` has started failing in CI:

https://ci.appveyor.com/project/serilog/serilog/build/955

```
[xUnit.net 00:00:03.9255834]       Assert.NotEqual() Failure
dotnet.exe : [xUnit.net 00:00:03.9252507]     Serilog.Tests.Context.LogContextTests.ContextPropertiesCrossAsyncCalls [FAIL]
At C:\projects\serilog\Build.ps1:43 char:5
+     & dotnet test -c Release
+     ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: ([xUnit.net 00:0...yncCalls [FAIL]:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
[xUnit.net 00:00:03.9256298]       Expected: Not 15
[xUnit.net 00:00:03.9256638]       Actual:   15
[xUnit.net 00:00:03.9263832]       Stack Trace:
[xUnit.net 00:00:03.9294312]         Context\LogContextTests.cs(200,0): at Serilog.Tests.Context.LogContextTests.<ContextPropertiesCrossAsyncCalls>d__7.MoveNext()
[xUnit.net 00:00:03.9295603]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:03.9297617]            at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[xUnit.net 00:00:03.9298184]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[xUnit.net 00:00:03.9298684]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:03.9299341]            at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[xUnit.net 00:00:03.9299947]            at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[xUnit.net 00:00:03.9300433]         --- End of stack trace from previous location where exception was thrown ---
[xUnit.net 00:00:03.9300952]            at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #1113, opened 2018-02-21
`ContextPropertiesCrossAsyncCalls` test failure fix on CI:

skomis-mm:
fixes [1112](https://github.com/serilog/serilog/issues/1112)

- no

- [ ] Unit Tests for the changes have been added (for bug fixes / features)

### DIFF
--- a/test/Serilog.Tests/Context/LogContextTests.cs
+++ b/test/Serilog.Tests/Context/LogContextTests.cs
@@ -3,15 +3,13 @@ using Serilog.Context;
 using Serilog.Events;
 using Serilog.Core.Enrichers;
 using Serilog.Tests.Support;
-#if APPDOMAIN
-using System;
-#endif
 #if REMOTING
 using System.Runtime.Remoting;
 using System.Runtime.Remoting.Lifetime;
 using System.Runtime.Remoting.Services;
 using System.Runtime.Remoting.Messaging;
 #endif
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Serilog.Core;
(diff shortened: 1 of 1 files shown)


test/Serilog.Tests/Context/LogContextTests.cs
```diff
@@ -349,6 +349,24 @@ public void DoesNotThrowOnCrossDomainCallsWhenLeaseExpired()
...
+
+        static async Task TestWithSyncContext(Func<Task> testAction, SynchronizationContext syncContext)
+        {
+            var prevCtx = SynchronizationContext.Current;
+            SynchronizationContext.SetSynchronizationContext(syncContext);
+
+            Task t;
+            try
+            {
+                t = testAction();
```
tsimbalar: Not sure at all about what I'm saying, but ... shouldn't there be an `await` here (i.e. in the `try` instead of **after** the `finally` ?

test/Serilog.Tests/Context/LogContextTests.cs
```diff
@@ -383,4 +401,9 @@ public bool IsCallable()
         }
     }
 #endif
+
+    class ForceNewThreadSyncContext : SynchronizationContext
```
tsimbalar: Mostly a personal opinion, but maybe having a `IDisposable` class would make the test code a bit more readable (less aync lambdas , and less indentation generally) , a bit like https://github.com/serilog/serilog/blob/dev/test/Serilog.Tests/Support/TemporarySelfLog.cs and its usage in https://github.com/serilog/serilog/blob/96bfe61c7e35271450062846374c13cce75b83e5/test/Serilog.Tests/Core/ChildLoggerKnownLimitationsTests.cs#L17 

It would be nice to have test code that looks like : 

```csharp
using(NewThreadSynchronizationContext.Start()){
  // the code that needs to be tested with that SynchronizationContext
}
```

test/Serilog.Tests/Context/LogContextTests.cs
skomis-mm: Thanks for the feedback! The problem with reverting sync context after await is that after await point I'm not owning thread anymore and this thread will lose its original context  (`Xunit.Sdk.AsyncTestSyncContext`)  so I'm not sure that this will be safe for other tests or xunit. You can set/reset safely while you own the thread, for example, for sync implementations. That's why I can't use here  disposable pattern.

test/Serilog.Tests/Context/LogContextTests.cs
skomis-mm: couldn't use await inside disposable block, see explanation above

nblumhardt:
Awesome stuff, I didn't notice we'd taken an implicit dependency on the test runner implementation, here.

PR #1156, opened 2018-05-03
[WIP] Capture any `IDictionary<>` as dictionary instead of sequence

krajek:
Capturing properties as dictionaries is rigid, it works only for `Dictionary` objects. It does not capture dictionaries even if class implements `IDictionary` interface. 

Yes, previously class other than `Dictionary` but implementing `IDictionary<>` would be destructured as sequence, after the PR it will be destructured as dictionary

**Other information**:
Currently, the PR is just a suggestion. In the `Serilog.Exceptions` I had an idea of using custom dictionary implementation to improve performance, but I had to resign because `Serilog` would not capture my custom dictionary as the dictionary.

Please let me know what you think about it. If you will think that such change is worthwhile I will refactor and add more tests.

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -311,8 +311,16 @@ namespace Serilog.Capturing
 
         bool TryGetDictionary(object value, Type valueType, out IDictionary dictionary)
         {
+#if NETSTANDARD1_0 || NETSTANDARD1_3
+        var implementedInterfaces = valueType.GetTypeInfo().ImplementedInterfaces;
+#else
+        var implementedInterfaces = valueType.GetInterfaces();
+#endif
+
+
             if (valueType.IsConstructedGenericType &&
-                valueType.GetGenericTypeDefinition() == typeof(Dictionary<,>) &&
+                implementedInterfaces.Any(
+                    interfaceType => interfaceType.GetTypeInfo().IsGenericType && interfaceType.GetGenericTypeDefinition() == typeof(IDictionary<,>)) &&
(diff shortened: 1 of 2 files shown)


tsimbalar:
Hi Artur and thank you for your PR,

the behaviour about `Dictionary<TKey, TValue>` vs `IDictionary<TKey, TValue>` is documented here : https://github.com/serilog/serilog/wiki/Structured-Data#collections , so unless I am missing something, this is by design.

> `IDictionary<TKey,TValue>` - objects implementing dictionary interfaces are not serialised as dictionaries. Firstly because it is less efficient in .NET to check for generic interface compatibility, and second because a single object may implement more than one generic dictionary interface, creating an ambiguity.

nblumhardt:
Thanks @krajek! Yes, as @tsimbalar points out, this is by design.

Is there an option to plug in a custom `IDestructuringPolicy` for your dictionary type, or a way to just create a `DictionaryValue(...)` directly? Cheers!

krajek:
Thanks guys, I acknowledge that destructuring of `IDictionary` will not be changed, so I will close this PR.

@nblumhardt thanks for mentioning `IDestructuringPolicy`. It looks like this is exactly what I need.

On the other hand, I do not understand what does it mean to create `DictionaryValue` directly. Could you explain, please?

nblumhardt:
@krajek :+1: .. I just mean that if you're able to plug in at the point where you're working in `LogEvent`s, `LogEventPropertyValue`s and so-on (i.e. in an enricher or destructuring policy) you can skip the whole thing and just create the properties directly - `new LogEventProperty("Example", new DictionaryValue(......))`. HTH!

PR #1165, opened 2018-05-13
Fixes a number of issues with `LoggerSinkConfiguration.Wrap()`

nblumhardt:
These were identified via the discussion in #1154.

 - Ensures that `wt.X(); wt.Y()` and `wt.X().WriteTo.Y()` are equivalent
 - Ensures a failing sink doesn't block events from reaching others inside the same wrapper, much as using a secondary logger would do
 - Prevents chained calls from accidentally breaking thread-safety rules by turning `Enrich` inside wrappers into a no-op

**What issue does this PR address?**

**Does this PR introduce a breaking change?**

**Please check if the PR fulfills these requirements**
- [ ] The commit follows our [guidelines](https://github.com/serilog/serilog/blob/dev/CONTRIBUTING.md)
- [ ] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:

PR #1166, opened 2018-05-13
Created Logger should not hold reference to its configuration

skomis-mm:
Logger references to its configuration instance via `Dispose` closure that captures it.

No.

**Other information**:
After checking implications of introducing new state (`.WriteTo`) for `LoggerConfiguration` in the neighboring [pr](https://github.com/serilog/serilog/pull/1165) found that current `CreateLogger()` captures `LoggerConfiguration` instances.

### DIFF
--- a/src/Serilog/LoggerConfiguration.cs
+++ b/src/Serilog/LoggerConfiguration.cs
@@ -136,12 +136,6 @@ namespace Serilog
                 throw new InvalidOperationException("CreateLogger() was previously called and can only be called once.");
             _loggerCreated = true;
 
-            Action dispose = () =>
-            {
-                foreach (var disposable in _logEventSinks.Concat(_auditSinks).OfType<IDisposable>())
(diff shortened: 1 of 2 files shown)


src/Serilog/LoggerConfiguration.cs
```diff
@@ -185,9 +179,18 @@ public Logger CreateLogger()
                 overrideMap = new LevelOverrideMap(_overrides, _minimumLevel, _levelSwitch);
             }
 
+            var disposableSinks = _logEventSinks.Concat(_auditSinks).OfType<IDisposable>().ToList();
```
bartelink: ToArray might be microscopically better?

skomis-mm: There is no reason to not making it microscopically better :) Done.

nblumhardt:
:+1: good catch.

PR #1169, opened 2018-05-18
Update Performance Tests to NetCoreApp2

merbla:
* Updated Perf Tests to `netcoreapp2.0`
* Added output to `results/netcoreapp2.0`

Not sure this is needed for 2.7.1, possibly hold off.


nblumhardt:
Shall we send this one through now, @merbla ?

merbla:
Yep, I think so.

PR #1173, opened 2018-06-07
Decoration is useful pattern when you try to do AOP

bnayae:
Enable to easily wrap Sinks in order to provide cross concern 
functionality like: 
 - send metric like duration of sink execution
 - filter or routing at sink level
 - and more

It use optional parameter in order to enable existing code to compile as is 

*Other information**:
I have context level activation logic which is external to the SeriLog project and I need away to set it per sink.
For example I may want some logs (according to their context into a File Sink and Other into ELK Sink.

I don't like to have separate fork for having this functionality.

### DIFF
--- a/src/Serilog/LoggerConfiguration.cs
+++ b/src/Serilog/LoggerConfiguration.cs
@@ -44,11 +44,17 @@ namespace Serilog
         bool _loggerCreated;
 
         /// <summary>
-        /// Construct a <see cref="LoggerConfiguration"/>.
+        /// Construct a <see cref="LoggerConfiguration" />.
         /// </summary>
-        public LoggerConfiguration()
+        /// <param name="sinkMap">The sink map.</param>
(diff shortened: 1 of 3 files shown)


bnayae:
I was looking at the Wrap method. the problem with this method was that it was hard to leverage the extension methods which come with each provider.
For example, the RollingFile extensions will use PeriodicFlushToDiskSink or RollingFileSink depending on the parameters. 
the use of Wrap will require to rewrite the logic of those extension methods

bnayae:
Sample of the problem:
The Following code should be rewrite and cannot use the provider extension when using Wrap:

var logConfig = new LoggerConfiguration();
logConfig = logConfig
                .WriteTo.**File**(
                    "log.txt", 
                    outputTemplate: "{Timestamp:yyyy-MM} [{Level:u3}] {Message:lj}{NewLine}{Exception}")
                .WriteTo.**Seq**("http://localhost:5341");

bnayae:
Thanks skomis-mm 
The sample code solve the issue.
This pull request is not really needed

PR #1174, opened 2018-06-07
Add partial support for .Destructure.XXX() methods in `KeyValuePairSettings`

### LINKED ISSUE #1172: Add support from `.Destructure.XXX()` methods and extension methods in `ReadFrom.KeyValuePairSettings`
In order to implement serilog/serilog-settings-appsettings#20 it actually needs to be implemented in the core *Serilog* repo.

Similar feature is already implemented in *Serilog.Settings.Configuration* via serilog/serilog-settings-configuration#110 and discussed in serilog/serilog-settings-configuration#106 .

I think that for now, we should keep it simple and just implement support for calling *methods* and *extension methods* on `Serilog.Configuration.LoggerDestructuringConfiguration` but **not support** the *generic* overloads, and only support parameter types that are currently supported by the configuration systems (i.e. no `Func<>` or `Action<>` so far). 

This would also encourage developers of `DestructuringPolicy`s to provide extension methods for a more friendly API and consistency with other configuration parts (*Sinks*, *Enrichers* etc).

Defining if and how we support calls to generic methods should probably be done in a separate issue.

tsimbalar:
Supports Serilog "native" non-generic simple  methods : 
- `.ToMaximumDepth(int maximumDestructuringDepth)`
- `.ToMaximumStringLength(int maximumStringLength)`
- `.ToMaximumCollectionCount(int maximumCollectionCount)`

and also for any *extension method* for `LoggerDestructuringConfiguration`.

Unsupported methods (should they be added ? and how would they be configured ? ) : 
- `.AsScalar(Type scalarType)`
- `.AsScalar<TScalar>()`
- `.With(params IDestructuringPolicy[] destructuringPolicies)`
- `.With<TDestructuringPolicy>()`
- `.ByTransforming<TValue>(Func<TValue, object> transformation)`
- `.ByTransformingWhere<TValue>(Func<Type, bool> predicate, Func<TValue, object> transformation)`       

Most test cases adapted from equivalent PR in serilog/serilog-settings-configuration#110 (thanks @MV10 ! ).

This should fix #1172 and automatically add the functionality for *Serilog.Settings.AppSettings* down the line (see serilog/serilog-settings-appsettings#20 )

If/when merged, we should remember to update the `README`in *Serilog.Settings.AppSettings*.

### DIFF
--- a/src/Serilog/Settings/KeyValuePairs/CallableConfigurationMethodFinder.cs
+++ b/src/Serilog/Settings/KeyValuePairs/CallableConfigurationMethodFinder.cs
@@ -23,37 +23,11 @@ namespace Serilog.Settings.KeyValuePairs
 {
     static class CallableConfigurationMethodFinder
     {
-        internal static LoggerConfiguration FromLogContext(LoggerEnrichmentConfiguration loggerEnrichmentConfiguration)
-        {
-            return loggerEnrichmentConfiguration.FromLogContext();
-        }
-
-        static readonly MethodInfo SurrogateEnrichFromLogContextConfigurationMethod = typeof(CallableConfigurationMethodFinder).GetTypeInfo().DeclaredMethods.Single(m => m.Name == nameof(FromLogContext));
-
-        internal static LoggerConfiguration ToMaximumCollectionCount(LoggerDestructuringConfiguration loggerDestructuringConfiguration,
-            int maximumCollectionCount)
-        {
-            return loggerDestructuringConfiguration.ToMaximumCollectionCount(maximumCollectionCount);
-        }
-
-        static readonly MethodInfo SurrogateDestructureToMaximumCollectionCountConfigurationMethod = typeof(CallableConfigurationMethodFinder).GetTypeInfo().DeclaredMethods.Single(m => m.Name == nameof(ToMaximumCollectionCount));
-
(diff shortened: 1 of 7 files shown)


skomis-mm:
👍

nblumhardt:
```

**Comment by nblumhardt** [comment]
```
.AsScalar(Type scalarType)
```

seems fairly clean to map onto `..:AsScalar.scalarType` with a qualified type name, perhaps?

We might consider creating a surrogate for `With()` that enables `..:destructure:with`, `..:enrich:with`, and `..:filter:with` in an analogous way to `..:using`, where the key might be suffixed with something arbitrary to avoid key collisions, a-la `..:destructure:with:MyPolicy`.

It's slightly murky territory, it'd be nice if everything consistently followed the `Method.parameter` syntax, but failing that, being consistent across these cases is worth something :-)

tsimbalar:
> `.AsScalar(Type scalarType)` -> `"..:AsScalar.scalarType"` 

yep, this one makes a lot of sense I believe . I'll create a new issue for this one (that also means, as a general rule, **support parameters of type `Type` by passing a qualified type name** ... I don't think it is supported right now, but I might be wrong).

> `We might consider creating a surrogate for With()`

I need to wrap my head around this one ... I'll open an issue to keep track of it ;)

PR #1179, opened 2018-06-13
Support for .Destructure.AsScalar(scalarType) in KeyValueSettings

### LINKED ISSUE #1177: Add support for .Destructure.AsScalar(Type scalarType) method in KeyValuePairSettings
based on comments from https://github.com/serilog/serilog/pull/1174#issuecomment-396839756

Ideally, calling 
`new LoggerConfiguration().Destructure.AsScalar(Type scalarType)`
should be supported by `ReadFrom.KeyValuePairs()` with the format : 
`"serilog:destructure:AsScalar.scalarType"` /  `"qualified type name"`.

This means : 
- making the method available even though it is not an *extension method* (needs a *surrogate method* with proper extension-method-ish signature)
- knowing how to handle parameters of type `Type` when given a *qualified type name*.

In addition, it would be good to 
- check if a *Assembly fully qualified type name* (`TopNamespace.SubNameSpace.ContainingClass, MyAssembly, Version=1.3.0.0, Culture=neutral, PublicKeyToken=b17a5c561934e089`) is required, or if we can do with only a *Fully qualified type name* (`TopNamespace.SubNameSpace.ContainingClass`) given that `using` directives already *import* an assembly (see also #1065 ).
- see how that would translate in the *Serilog.Settings.Configuration* side of things ...

tsimbalar:
This should fix #1177 

no

**Other information**:
Adds support for key-value settings like : 

```xml
<add key="serilog:destructure:AsScalar.scalarType" value="MyNamespace.MyType, MyAssembly, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
```

As a side-effect, it also adds global support for calling **configuration methods that have a parameter of type `Type`**, by accepting the type name.

The type lookup relies on [`Type.GetType(string typeName, bool throwOnError)`](https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype?view=netframework-4.7.1#System_Type_GetType_System_String_System_Boolean_) (with `throwOnError` set to `true`) so it accepts a type name in the form : 

> The assembly-qualified name of the type to get. See AssemblyQualifiedName. If the type is in the currently executing assembly or in Mscorlib.dll, it is sufficient to supply the type name qualified by its namespace.

### DIFF
--- a/src/Serilog/Settings/KeyValuePairs/CallableConfigurationMethodFinder.cs
+++ b/src/Serilog/Settings/KeyValuePairs/CallableConfigurationMethodFinder.cs
@@ -25,6 +25,7 @@ namespace Serilog.Settings.KeyValuePairs
     {
         static readonly MethodInfo[] SurrogateMethodCandidates = typeof(SurrogateConfigurationMethods).GetTypeInfo().DeclaredMethods.ToArray();
         static readonly MethodInfo SurrogateEnrichFromLogContextConfigurationMethod = SurrogateMethodCandidates.Single(m => m.Name == nameof(SurrogateConfigurationMethods.FromLogContext));
+        static readonly MethodInfo SurrogateDestructureAsScalarConfigurationMethod = SurrogateMethodCandidates.Single(m => m.Name == nameof(SurrogateConfigurationMethods.AsScalar));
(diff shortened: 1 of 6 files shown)


tsimbalar:
Hi @serilog/reviewers-core & @serilog/reviewers-configuration  

does that look good enough to merge ? :)

nblumhardt:
LGTM! Nice one :+1:

sukchai19933:

PR #1183, opened 2018-06-20
Take into account custom `IFormatProvider` when format is `{Message:lj}`

### LINKED ISSUE #1115: CustomDateFormatter example kills all output
**A few questions before you begin:**

> Is this an issue related to the Serilog core project or one of the [sinks]
Serilog core

- [x] Bug
- [ ] New Feature

2.6.0

- [ ] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [x] 4.6.x
- [ ] 4.5.x

I'm following the CustomDateFormatter example on https://github.com/serilog/serilog/wiki/Formatting-Output to provide custom formatting for DateItem properties in messages.  When I add the CustomDateFormatter to my sinks, it disables all output.

When I debug the GetFormat() method, its always passed an object of type System.ICustomFormatter and not a type like System.Globalization.DateTimeFormatInfo.  

It looks like this is invoked from `ScalarValue.cs`:
```csharp
var custom = (ICustomFormatter)formatProvider.GetFormat(typeof(ICustomFormatter));
```

And I suspect that since the example code is returning an IFormatProvider, that the cast is failing which is why nothing is getting printed.

I used this page (https://msdn.microsoft.com/en-us/library/system.icustomformatter(v=vs.110).aspx) as a guide and reimplemented my class as follows:
```csharp
public class CustomDateFormatter : IFormatProvider, ICustomFormatter
{
	private readonly string _datePattern;

	/// <summary>
	/// Initializes a new instance of the <see cref="DateFormatter"/> class.
	/// </summary>
	/// <param name="datePattern">date format pattern</param>
	/// <param name="basedOn">base format provider</param>
	public CustomDateFormatter(string datePattern)
	{
		_datePattern = datePattern;
	}

	/// <summary>
	/// Gets the Formatter object
	/// </summary>
	/// <param name="formatType">format type</param>
	/// <returns>formatter</returns>
	public object GetFormat(Type formatType)
	{
		if (formatType == typeof(ICustomFormatter))
		{
			return this;
		}

		return null;
	}

	/// <summary>
	/// Formats an object for output
	/// </summary>
	/// <param name="format">optional format string, if not provided the default date pattern is used</param>
	/// <param name="arg">the object to be formatted</param>
	/// <param name="formatProvider">format provider</param>
	/// <returns>formatted string</returns>
	public string Format(string format, object arg, IFormatProvider formatProvider)
	{
		if (arg is DateTime)
		{
			return ((DateTime)arg).ToString(format ?? _datePattern);
		}
		else
		{
			try
			{
				return HandleOtherFormats(format, arg);
			}
			catch (FormatException e)
			{
				throw new FormatException($"The format of '{format}' is invalid", e);
			}
		}
	}

	/// <summary>
	/// Fallback formatter for non DateTime objects
	/// </summary>
	/// <param name="format"></param>
	/// <param name="arg"></param>
	/// <returns></returns>
	private string HandleOtherFormats(string format, object arg)
	{
		if (arg is IFormattable)
		{
			return ((IFormattable)arg).ToString(format, CultureInfo.CurrentCulture);
		}
		else if (arg != null)
		{
			return arg.ToString();
		}
		else
		{
			return string.Empty;
		}
	}
}
```
It seems to be working correctly.  In Console at least.  Does the File sink not use the formatter even if I provide a formatProvider on WriteTo.File() ?

Just need an example for formatting DateTime that reflects current practice.

tsimbalar:
Try to fix #1115 

Based on discussions around the issue about where it should make sense to consider the user-provider `IFormatProvider` : 
- `:j` is a serialization format so it should not be customizable
- `:lj` is targeted towards human beings, so it should be customizable (it is not right now)
- `:l` should be customizable (it is already)
- ` ` should be customizable (it is already)

It makes sense to try and fix `:lj` given that it is the default *Output template* for *Serilog.Sinks.Console* and *Serilog.Sinks.File* (they use 
`"[{Timestamp:HH:mm:ss} {Level:u3}] {Message:lj}{NewLine}{Exception}"` and 
`"{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level:u3}] {Message:lj}{NewLine}{Exception}"` respectively ).

### DIFF
--- a/src/Serilog/Rendering/MessageTemplateRenderer.cs
+++ b/src/Serilog/Rendering/MessageTemplateRenderer.cs
@@ -51,7 +51,7 @@ namespace Serilog.Rendering
                 }
                 else
                 {
-                    var pt = (PropertyToken) token;
+                    var pt = (PropertyToken)token;
                     RenderPropertyToken(pt, properties, output, formatProvider, isLiteral, isJson);
                 }
             }
@@ -92,9 +92,12 @@ namespace Serilog.Rendering
 
         static void RenderValue(LogEventPropertyValue propertyValue, bool literal, bool json, TextWriter output, string format, IFormatProvider formatProvider)
         {
-            if (literal && propertyValue is ScalarValue sv && sv.Value is string str)
+            // _literal_ means scalar values are not wrapped in "quotes"
+            // _json_ means ... json
+            // _literal_ AND _json_ means scalar values without quotes and structured values as json
+            if (literal && propertyValue is ScalarValue sv)
             {
-                output.Write(str);
+                RenderLiteralScalar(sv, output, format, formatProvider);
             }
             else if (json)
             {
@@ -104,6 +107,18 @@ namespace Serilog.Rendering
             {
                 propertyValue.Render(output, format, formatProvider);
             }
+        }
+
+        static void RenderLiteralScalar(ScalarValue scalarValue, TextWriter output, string format, IFormatProvider formatProvider)
+        {
+            if (scalarValue.Value is string str)
+            {
+                // render without double quotes. _format_ makes no sense for strings
+                output.Write(str);
+                return;
+            }
+
+            scalarValue.Render(output, format, formatProvider);
         }
     }
 }
--- a/test/Serilog.Tests/Formatting/Display/MessageTemplateTextFormatterTests.cs
+++ b/test/Serilog.Tests/Formatting/Display/MessageTemplateTextFormatterTests.cs
@@ -309,7 +309,7 @@ namespace Serilog.Tests.Formatting.Display
         [Theory]
         [Trait("Bugfix", "#1115")]
         [InlineData("", true)]
-        [InlineData(":lj", true)]
+        [InlineData(":lj", false)]
         [InlineData(":j", false)]
         [InlineData(":l", true)]
         public void FormatProviderIsUsedForNonJsonFormatWithStructure(string format, bool shouldUseCustomFormatter)
--- a/test/Serilog.Tests/Formatting/Display/MessageTemplateTextFormatterTests.cs
+++ b/test/Serilog.Tests/Formatting/Display/MessageTemplateTextFormatterTests.cs
@@ -277,31 +277,68 @@ namespace Serilog.Tests.Formatting.Display
             Assert.Equal(expected, sw.ToString());
         }
 
-        [Fact]
+        [Theory]
         [Trait("Bugfix", "#1115")]
-        public void CustomDateFormatProviderCanBeUsed()
+        [InlineData("", true)]
+        [InlineData(":lj", true)]
+        [InlineData(":j", false)]
+        [InlineData(":l", true)]
+        public void CustomDateFormatProviderCanBeUsedInNonJsonFormatWithStructure(string format, bool shouldUseCustomFormatter)
         {
             // relying on the example from wiki : https://github.com/serilog/serilog/wiki/Formatting-Output
-            // using 5 digits for year on purpose, so it is visible !
-            var customDateFormatProvider = new CustomDateFormatter("yyyyy-MM-dd", CultureInfo.InvariantCulture);
+            var date = new DateTime(2018, 01, 01);
 
-            var formatter = new MessageTemplateTextFormatter("{Message}", customDateFormatProvider);
-            var evt = DelegatingSink.GetLogEvent(l => l.Information("{@Item}", new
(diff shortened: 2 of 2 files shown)


tsimbalar:
For now, I have only added tests to be sure I understand the expected outcome :)

Of the tests below, all the ones targetting format `:lj` are failing .... the other ones are green .

```csharp
[Theory]
[Trait("Bugfix", "#1115")]
[InlineData("", true)]
[InlineData(":lj", true)]
[InlineData(":j", false)]
[InlineData(":l", true)]
public void CustomDateFormatProviderCanBeUsedInNonJsonFormatWithStructure(string format, bool shouldUseCustomFormatter)
{
    // relying on the example from wiki : https://github.com/serilog/serilog/wiki/Formatting-Output
    var date = new DateTime(2018, 01, 01);

    // using 5 digits for year on purpose, so it is visible !
    var customDateFormatProvider = new CustomShortDateFormatProvider("yyyyy", CultureInfo.InvariantCulture);
    var expectedFormattedDate = shouldUseCustomFormatter ? date.ToString(customDateFormatProvider) : date.ToString("O");

    var formatter = new MessageTemplateTextFormatter("{Message" + format + "}", customDateFormatProvider);
    var evt = DelegatingSink.GetLogEvent(l =>
    {
        l.Information("{@Item}", new
        {
            CreatedOn = date
        });
    });
    var sw = new StringWriter();
    formatter.Format(evt, sw);

    Assert.Contains(expectedFormattedDate, sw.ToString());
}

[Theory]
[Trait("Bugfix", "#1115")]
[InlineData("", true)]
[InlineData(":lj", true)]
[InlineData(":j", false)]
[InlineData(":l", true)]
public void CustomDateFormatProviderCanBeUsedInNonJsonFormatWithScalar(string format, bool shouldUseCustomFormatter)
{
    // relying on the example from wiki : https://github.com/serilog/serilog/wiki/Formatting-Output
    var date = new DateTime(2018, 01, 01);

    // using 5 digits for year on purpose, so it is visible !
    var customDateFormatProvider = new CustomShortDateFormatProvider("yyyyy", CultureInfo.InvariantCulture);
    var expectedFormattedDate = shouldUseCustomFormatter ? date.ToString(customDateFormatProvider) : date.ToString("O");

    var formatter = new MessageTemplateTextFormatter("{Message" + format + "}", customDateFormatProvider);
    var evt = DelegatingSink.GetLogEvent(l =>
    {
        l.Information("{MyDate}", date);
    });
    var sw = new StringWriter();
    formatter.Format(evt, sw);

    Assert.Contains(expectedFormattedDate, sw.ToString());
}
```

@serilog/reviewers-core : can you confirm that those tests should pass ?

adamchester:
Just to be sure I understand correctly, are we now saying `{Message:lj}` is basically "JSON, except  when rendering `Scalar` values like `string` or `DateTime` (which should be rendered as something more human-readable" ?

nblumhardt:
@adamchester that's my take on it :+1:

adamchester:
Sounds good :+1:

In that case, may I suggest the tests don't focus purely on custom `DateTime` formatting?

tsimbalar:
> In that case, may I suggest the tests don't focus purely on custom `DateTime` formatting?

Makes sense :) I started there given that we have an example of a custom `IFormatProvider` for dates in the wiki. I'll add more general tests for all `IFormattable` values.

skomis-mm:
Looks good 👍

tsimbalar:
This is ready for review, but failing because of #1188 .

skomis-mm:
👍

skomis-mm:
@tsimbalar its probably need some commit to trigger new AppVeyor build?

tsimbalar:
I probably need to rebase my branch on top of develop and push it again. I'll do that when I get home ;) 

Thanks for the UWP fix @skomis-mm

tsimbalar:
I don't quite understand the build failure here . HEEEELP 🆘

nblumhardt:
Any thoughts on https://github.com/serilog/serilog-sinks-file/issues/57#issuecomment-424154262 ?

tsimbalar:
@nblumhardt based on the discussion in the other issue, this PR should probably be closed without merging. I'll try and prepare another one with only the tests so we can at least keep those ..

nblumhardt:
Thanks for the loop back, @tsimbalar - I guess the corollary now is that the console behavior is not ideal :-) .. I'll find some time to come back and follow up in that direction; heading over there to raise an issue now.

PR #1192, opened 2018-07-17
Add support for Destructure.With(IDestructuringPolicy policy) in settings

### LINKED ISSUE #1190: Missing support for Destructure.With(IDestructuringPolicy policy) in KeyValue Settings
Currently, it is not possible to express the equivalent of : 

```csharp
new LoggerConfiguration()
  .Destructure.With(new MyCustomPolicy());
  .CreateLogger()
```

through key-value settings (and therefore AppSettings). ( this scenario is already handled properly by *Serilog.Settings.Configuration*. )

Given the current handling of settings, I would expect it to take the form : 

```xml
  <appSettings>
    <add key="serilog:using:TestDummies" value="TestDummies" />
    <add key="serilog:destructure:With.policy" value="TestDummies.Policies.CustomPolicy, TestDummies" />
  </appSettings>
```

The key-value settings already know how to call the default constructor of an implementor/subclass when configuration method arguments are `interface`s or `abstract class`es. What is possibly missing here is :
- `Destructure.With(params IDestructuringPolicy[] destructuringPolicies)` is not an extension method , but a *regular* method
- `Destructure.With(params IDestructuringPolicy[] destructuringPolicies)` has a `params` array argument, which is not supported right now in the setting systems ...

There are several possible approaches to fill the gap : 
1. do nothing :)
2. support arrays / `params` arrays in the settings provider(s)
3. add an overload `Destructure.With(IDestructuringPolicy policy)` (or possibly an extension method)
4. add `Destructure.With(IDestructuringPolicy policy)` as a *surrogate method* for the settings system

I probably like **option 3** best, but I am not quite sure about the impact and possible ambiguity caused by introducing such a new method. 
**Option 4** is the way it is handled in *Serilog.Settings.Configuration*, I think (see https://github.com/serilog/serilog-settings-configuration/blob/930f014ddff8fa12173ba9c39211fc7b85df47fe/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs#L424-L425 ) and seems rather reasonable.

Any thoughts ?

tsimbalar:
fixes #1190

The key-value settings will accept the assembly-qualified name of a type implementing IDestructuringPolicy if it has a default constructor

### DIFF
--- a/src/Serilog/Settings/KeyValuePairs/CallableConfigurationMethodFinder.cs
+++ b/src/Serilog/Settings/KeyValuePairs/CallableConfigurationMethodFinder.cs
@@ -29,6 +29,7 @@ namespace Serilog.Settings.KeyValuePairs
         static readonly MethodInfo SurrogateDestructureToMaximumCollectionCountConfigurationMethod = SurrogateMethodCandidates.Single(m => m.Name == nameof(SurrogateConfigurationMethods.ToMaximumCollectionCount));
         static readonly MethodInfo SurrogateDestructureToMaximumDepthConfigurationMethod = SurrogateMethodCandidates.Single(m => m.Name == nameof(SurrogateConfigurationMethods.ToMaximumDepth));
         static readonly MethodInfo SurrogateDestructureToMaximumStringLengthConfigurationMethod = SurrogateMethodCandidates.Single(m => m.Name == nameof(SurrogateConfigurationMethods.ToMaximumStringLength));
+        static readonly MethodInfo SurrogateDestructureWithConfigurationMethod = SurrogateMethodCandidates.Single(m => m.Name == nameof(SurrogateConfigurationMethods.With));
 
         internal static IList<MethodInfo> FindConfigurationMethods(IEnumerable<Assembly> configurationAssemblies, Type configType)
         {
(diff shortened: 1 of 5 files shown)


nblumhardt:
Love your work on getting these all lined up, @tsimbalar ! :+1:

nblumhardt:
Another thought along the same lines - do we currently support `WriteTo.Sink()` this way? Do we want to? :-)

tsimbalar:
> do we currently support WriteTo.Sink() this way? 

I don't think it is currently supported  .... 

> Do we want to? :-)

that's a strange use case, but for consistency among the configuration options that might make sense .

To cover all cases, that would mean support for : 
- `WriteTo.Sink(ILogEventSink logEventSink, LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum, LoggingLevelSwitch levelSwitch = null)`
- `AuditTo.Sink(ILogEventSink logEventSink, LogEventLevel restrictedToMinimumLevel = LevelAlias.Minimum, LoggingLevelSwitch levelSwitch = null)`
- `Enrich.With(params ILogEventEnricher[] enrichers)` (with surrogate `Enrich.With(ILogEventEnricher enricher)` )
- `Filter.With(params ILogEventFilter[] filters)` (with surrogate `Enrich.With(ILogEventFilter filter)` )

both in *key-value* settings and *Serilog.Settings.Configuration*.

tsimbalar:
I created #1194 to track it.

sukchai19933:

PR #1197, opened 2018-07-28
Allow SilentLogger to be contructed as singleton

### LINKED ISSUE #1195: SilentLogger not a singleton, reference comparision with Logger.None
This relates to the Serilog core, this repository

It is debatable if this is a new feature or an existing bug. It is also framework independent.

Versions from Fixes #1105 up until the current dev as of now https://github.com/serilog/serilog/commit/fcd0c8bb128a853b48869dd7821b3e8661963319

The SilentLogger instance of Log.Logger is not the same reference as Logger.None, thus it is impossible to determine if Log.Logger has been initialized or not.

If Log.Logger has not been initialized, it should be set to the same instance as Logger.None, so Object.ReferenceEquals(Log.Logger, Logger.None) returns a boolean that identifies, whether Log.Logger is uninitialized or closed.

To reproduce, test whether at program start Object.ReferenceEquals(Log.Logger, Logger.None) returns true or false. It should return true until Log.Logger has been assigned a newly created logger. It should also return true when Log.CloseAndFlush() has been called.

My current workaround for this is to initialize Log.Logger = Logger.None at program start and whenever after I used Log.CloseAndFlush()

ie-zero:
Fix #1195: This allows comparision Log.Logger = Logger.None

No, the change should not have any breaking effect to the clients

**Other information**:
I do not believe it is appropriate/there is a need to override Equals() method as it is not a value type but I am happy to take any suggestions.

### DIFF
--- a/src/Serilog/Core/Pipeline/SilentLogger.cs
+++ b/src/Serilog/Core/Pipeline/SilentLogger.cs
@@ -20,7 +20,7 @@ namespace Serilog.Core.Pipeline
 {
     class SilentLogger : ILogger
     {
-        public readonly static ILogger None = new SilentLogger();
+        public static readonly ILogger Instance = new SilentLogger();
 
         public ILogger ForContext(ILogEventEnricher enricher)
         {
--- a/src/Serilog/Core/Pipeline/SilentLogger.cs
+++ b/src/Serilog/Core/Pipeline/SilentLogger.cs
@@ -20,6 +20,8 @@ namespace Serilog.Core.Pipeline
 {
     class SilentLogger : ILogger
     {
+        public readonly static ILogger None = new SilentLogger();
+
         public ILogger ForContext(ILogEventEnricher enricher)
         {
             return this;
--- a/src/Serilog/Core/Pipeline/SilentLogger.cs
+++ b/src/Serilog/Core/Pipeline/SilentLogger.cs
@@ -22,6 +22,10 @@ namespace Serilog.Core.Pipeline
     {
         public static readonly ILogger Instance = new SilentLogger();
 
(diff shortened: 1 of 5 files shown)


src/Serilog/Core/Pipeline/SilentLogger.cs
```diff
@@ -20,6 +20,8 @@ namespace Serilog.Core.Pipeline
 {
     class SilentLogger : ILogger
     {
+        public readonly static ILogger None = new SilentLogger();
```
nblumhardt: Declaration order would normally be `public static readonly` according to the conventions of this codebase.

The property name `Instance` would possibly be more appropriate here?

ie-zero:
I amended the code as proposed in the review. 

I realised however that the unit test added fails intermittently for some reason which needs to invetigated further before commited to master.

nblumhardt:
Ah, I see - the test depends on `Log.Logger` being uninitialized, but it's quite possible other tests have already configured the static logger.

Making `SilentLogger`'s constructor private would catch this one: https://github.com/serilog/serilog/blob/dev/src/Serilog/Log.cs#L64

If the line above is changed to use `SilentLogger.Instance`, then `CloseAndFlush()` can be called in the test before asserting the identity of `Logger`.

ie-zero:
Thank you. It makes sense now. It appeared to be a race condition (what else would have been) but I could not locate the source.

I always thought that each test on xUnit was completely isolated but, I guess, this is not the case with static members.

sukchai19933:

sukchai19933:
ok

nblumhardt:
:+1: thanks!

PR #1199, opened 2018-07-30
Level Output Formatter to support 5 chars in order to try & map Log4Net

warrenbuckley:
This is an attempt in order to format text files as the same as Log4Net that uses this in their templates `%-5level` so that we can migrate more seamlessly from Log4Net to Serilog

Perhaps but I doubt anyone was using {Level:u5} in their templates as this would output exactly the first 5 characters, but I may be wrong


src/Serilog/Formatting/Display/LevelOutputFormat.cs
```diff
@@ -26,30 +26,30 @@ namespace Serilog.Formatting.Display
...
         };
 
         static readonly string[][] _uppercaseLevelMap = {
-            new []{ "V", "VB", "VRB", "VERB" },
-            new []{ "D", "DE", "DBG", "DBUG" },
-            new []{ "I", "IN", "INF", "INFO" },
-            new []{ "W", "WN", "WRN", "WARN" },
-            new []{ "E", "ER", "ERR", "EROR" },
-            new []{ "F", "FA", "FTL", "FATL" }
+            new []{ "V", "VB", "VRB", "VERB", "VERBO" },
```
bartelink: Are you sure this is a valid Log4Net value ? maybe should be trace ?

bartelink:
For NLog, the mapping is as follows:- 

```
    let NLogLevelEnricher () =
        let map = function
            | Serilog.Events.LogEventLevel.Verbose ->   "TRACE"
            | Serilog.Events.LogEventLevel.Debug ->     "DEBUG"
            | Serilog.Events.LogEventLevel.Information -> "INFO"
            | Serilog.Events.LogEventLevel.Warning  ->  "WARN"
            | Serilog.Events.LogEventLevel.Error  ->    "ERROR"
            | Serilog.Events.LogEventLevel.Fatal  ->    "FATAL"
            | x -> invalidArg "level" (sprintf "Invalid value %O" x)
        let enrich (e : Events.LogEvent) = e.AddPropertyIfAbsent(Events.LogEventProperty("NLogLevel", Events.ScalarValue(map e.Level)))
        { new Serilog.Core.ILogEventEnricher with member __.Enrich(evt,_) = enrich evt }
```

Arguably, it would be interesting to see if you could make this PR also accommodate this.

(My gut says this sort of backcompat should not go into the core package though - shimming like this should be in infrastructure code in order to nudge this sort of thing out of existence.)

nblumhardt:
Thanks for the PR @warrenbuckley !

I tend to follow @bartelink 's line of thinking, here - would rather avoid the change if we can.

It's possible to work around this using an `ILogEventEnricher` - just add a property `Log4NetLevel` to each event, and then use `{Log4NetLevel}` in the output template instead of `{Level}` - is that an option for your usage?

Cheers!

warrenbuckley:
Yes I will close this PR.
Thanks for the suggestion for a much simpler solution @nblumhardt & @bartelink

bartelink:
Great to see your willingness to contribute - can I suggest a self-answered Q+A on SO about how to do this migration/shimming ? Post it in the gitter so others can vote/doubel check it (can't recall if there is something there already)

warrenbuckley:
@bartelink Here is the posted StackOverflow
https://stackoverflow.com/questions/51629436/how-do-i-modify-serilog-log-levels-to-match-up-with-log4net#51629437

Issue #1224, opened 2018-10-01
Configure AppVeyor to ignore non-source commits


MV10:
Open for discussion before I start submitting PRs:

In all the repos I've checked, AppVeyor builds and deploys a new package for every commit, regardless of scope or content. It is possible to restrict the configuration to only build when source or other build-related files are changed, such that changes to non-package content like README files or unit tests will not trigger a new package. 

I intend for this to be a tracking issue across many Serilog repos (just as Maxime did for #1219 to update many SLNs at once).

Here is an example (the assets folder is included because it's home to the Serilog strong-name signing key and Build.ps1 is the AppVeyor script): 

```yml
only_commits:
  files:
    - serilog-settings-configuration.sln
    - src/Serilog.Settings.Configuration/
    - Build.ps1
    - assets/
```

Since Maxime already did the hard work of tracking down which repos are in active development, I've followed his lead and will update mostly the same list. I've already forked and branched all of them and by the end of today I should have all the configs updated and ready to PR.

Please chime in if you see something I've overlooked, or if you're active in any of these repos and know of special handling they might require. (I realize not all repos follow exactly the same src/sln naming pattern, I'll handle that correctly on a case-by-case basis.) A case I'm wondering about is whether any repo uses AppVeyor for PR unit test validation. The repos I've been involved with all have a separate TravisCI process that covers this regardless of whether AppVeyor runs.

merbla:
@MV10 could we include tests? I would prefer a build runs on any change to tests.

MV10:
@merbla Are you referring to this repo, or in general? Serilog has Travis and that includes tests without uploading a new NuGet dev package.

It would be just one more line to reference the test directory but I figured even in repos without Travis you're unlikely to only commit tests without corresponding app code changes.

MV10:
@merbla Ok, I'll get on it later today. I kind of wondered about the Travis vs AppVeyor thing, they did seem to cover almost the same ground. I'd be happy to see Travis go away, it's painfully slow and I hate walking away from a PR not having full confirmation that something didn't go sideways. I think out of the list I updated only four repos use it.

And you're right, for exactly the same reasons I voiced (test commits without app changes will be rare) this change would produce minimal churn while retaining test validations.

MV10:
All PRs updated (well, all of them which have unit tests, a few do not).

If anyone wants to pursue getting rid of Travis, these use it out of the list I worked with:

* serilog
* serilog-sinks-console
* serilog-sinks-debug
[...]

PR #1228, opened 2018-10-07
only build packaging-related commits

MV10:
Tracked in https://github.com/serilog/serilog/issues/1224


nblumhardt:
@MV10 hope all's going well. When last there was some discussion around this, I think the idea was to exclude README rather than include everything else. No rush on it; I'll close this earlier PR now though to avoid confusion. Cheers!

Issue #1235, opened 2018-10-16
Logger.ForContext method raise ArgumentNullException for the object created by System.ServiceModel.ChannelFactory


balayoglu:
This issue relates to the Serilog core project Capturing module.

**Does this issue relate to a new *feature* or an existing *bug*?**
- [x] Bug

**What version of Serilog is affected?  Please list the related NuGet package.**
I tested with the below versions, error exists on both of them.
https://www.nuget.org/packages/Serilog/2.7.1
https://www.nuget.org/packages/Serilog/2.6.0

**What is the target framework and operating system? See [target frameworks]
- [x] 4.6.x
- [x] 4.5.x

**Please describe the current behavior?**
I am getting the below error when I pass to the **_Logger.ForContext_** method an instance of WCF  SOAP Client.

> System.ArgumentNullException: Value cannot be null.
> Parameter name: type
>    at System.Reflection.IntrospectionExtensions.GetTypeInfo(Type type)
>    at Serilog.Capturing.GetablePropertyFinder.<GetPropertiesRecursive>d__0.MoveNext()
>    at Serilog.Capturing.PropertyValueConverter.<GetProperties>d__21.MoveNext()
>    at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
>    at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
>    at Serilog.Events.StructureValue..ctor(IEnumerable`1 properties, String typeTag)
>    at Serilog.Capturing.PropertyValueConverter.TryConvertCompilerGeneratedType(Object value, Destructuring destructuring, Type valueType, LogEventPropertyValue& result)
>    at Serilog.Capturing.PropertyValueConverter.CreatePropertyValue(Object value, Destructuring destructuring, Int32 depth)
>    at Serilog.Capturing.PropertyValueConverter.DepthLimiter.CreatePropertyValue(Object value, Destructuring destructuring)
>    at Serilog.Capturing.PropertyValueConverter.<GetProperties>d__21.MoveNext()
>    at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
>    at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
>    at Serilog.Events.StructureValue..ctor(IEnumerable`1 properties, String typeTag)
>    at Serilog.Capturing.PropertyValueConverter.TryConvertCompilerGeneratedType(Object value, Destructuring destructuring, Type valueType, LogEventPropertyValue& result)
[...]

PR #1236, opened 2018-10-17
Fix on GetPropertiesRecursive method not to generate exception on the… 

balayoglu:
Fix for the issue #1235.


nblumhardt:
Many thanks for this 👍

PR #1269, opened 2019-01-10
Add .NET Standard 2.0 target

### LINKED ISSUE #1268: Add a .NET Standard 2.0 target
Serilog 2.7.1 supports all current versions of .NET Core [via the `netstandard1.3` TFM](https://github.com/serilog/serilog/blob/dev/src/Serilog/Serilog.csproj#L7). Its smoother to consume libraries that target .NET Standard 2.0 (`netstandard2.0`) with current tooling - we should add this TFM even though it won't expand the platforms reachable by the package.

nblumhardt:
Fixes #1268

### DIFF
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -1,10 +1,10 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Simple .NET logging with fully-structured events</Description>
-    <VersionPrefix>2.7.2</VersionPrefix>
+    <VersionPrefix>2.8.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>net45;net46;netstandard1.0;netstandard1.3</TargetFrameworks>
+    <TargetFrameworks>net45;net46;netstandard1.0;netstandard1.3;netstandard2.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyName>Serilog</AssemblyName>
     <AssemblyOriginatorKeyFile>../../assets/Serilog.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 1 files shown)


skomis-mm:
According to the [guideline](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting#multi-targeting) (thanks to [comment](https://github.com/serilog/serilog-sinks-periodicbatching/issues/27#issue-397707772) from PBS) would it makes sense to also include `net461` target? The issues mentioned probably refers to `System.Net.Http` dependency where there were some troubles, so should not be an issue for this package..but probably needs to check. Otherwise, LGTM 👍

skomis-mm:
> ...would it make sense to also include net461 target? The issues mentioned...
Seems its irrelevant since `Serilog` already have `net46` target and  .NET Framework applications will prefer that target instead of `netstandard2.0`, nvm.

nblumhardt:
Thanks @skomis-mm 👍

src/Serilog/Serilog.csproj
```diff
@@ -52,4 +56,8 @@
     <PackageReference Include="System.Collections.NonGeneric" Version="4.3.0" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="System.Collections.NonGeneric" Version="4.3.0" />
```
davidfowl: I just started playing with Serilog again and noticed it pulling in a bunch of unneeded packages. This is the root cause. Which collection is being used and what isn't it generic? 😄

nblumhardt: 😄 it's this one: https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/Pipeline/MessageTemplateCache.cs#L32

IIRC, `Hashtable`'s contract supports concurrent/unsynchronized reads in the presence of a single writer, while `Dictionary<K,V>` dropped this.

nblumhardt: > PS: Everything versioned 4.3.0 are packages that existed pre-netstandard 2.0. If they don't have a newer version then it was included in the standard.

PR #1314, opened 2019-05-23
Some very minor cleanup

SimonCropp:


src/Serilog/Capturing/DepthLimiter.cs
```diff
@@ -37,7 +37,7 @@ public DepthLimiter(int maximumDepth, PropertyValueConverter propertyValueConver
                 _propertyValueConverter = propertyValueConverter;
             }
 
-            public void SetCurrentDepth(int depth)
+            public static void SetCurrentDepth(int depth)
```
nblumhardt: Originally, the thinking here was that the use of `ThreadStatic` was an implementation detail subject to change, but, I think chances of that not being the case have become very slim :-)

Issue #1310, opened 2019-05-15
Levelled enrichers/`LoggerEnrichmentConfiguration.Wrap()`


nblumhardt:
> @nblumhardt Is there a way to enable an enricher only when a specific severity level is enabled? It would be nice to be able to use this code only when, say, Verbose is enabled.

_Originally posted by @andre-ss6 in https://github.com/serilog/serilog/issues/1084#issuecomment-492396008_

nblumhardt:
It would be nice to go back in time and build levelling support into enricher configuration methods, much like we've done with sinks, but there's a lot of churn involved in that.

Alternatively, we could make it possible for a levelled enricher wrapper to be written:

```csharp
    .Enrich.AtLevel(LogEventLevel.Verbose, e => e.WithMachineName())
```

We provide `LoggerSinkConfiguration.Wrap()` so that wrapper sinks can modify the behavior of other sinks, and `Enrich.AtLevel()` could be built on a similar `LoggerEnrichmentConfiguration.Wrap()` method, were one available.

Thoughts? Are there any other use cases that could be implemented on top of this? Would a similar method for destructuring also find use?

vlm---:
+1 for leveled enrichment.
Leveled approach would be also helpful for destructuring, omitting certain properties on higher levels but adding them on others. It should support including or excluding configuration with list of levels and higher / lower than level approach.
In one of our projects we used forked [Destructurama.Attributed](https://github.com/destructurama/attributed) with added attribute `LogAtAttribute(LogEventLevel level)` for ignoring properties below defined level, but performance of that implementation was not optimal for widespread use and we ditched it after the project matured and verbose /debug logging was not needed as much.

nblumhardt:
Just looking at the example again, we'd need to use `LoggingLevelSwitch` with the API if we want to avoid having to re-specify the minimum level:

```csharp
    .Enrich.AtLevel(LogEventLevel.Verbose, levelSwitch, e => e.WithMachineName())
```

abrasat:
> In one of our projects we used forked [Destructurama.Attributed](https://github.com/destructurama/attributed) with added attribute `LogAtAttribute(LogEventLevel level)` for ignoring properties below defined level, but performance of that implementation was not optimal for widespread use and we ditched it after the project matured and verbose /debug logging was not needed as much.

Could you give some details about bad performance issues ?

Is it basically a "bad" idea to use serilog enrichers, as these impact on application performance?

vlm---:
[...]

PR #1316, opened 2019-05-29
Wrapped enrichers, conditional/leveled enrichers, conditional sinks

nblumhardt:

This introduces some new configuration options, in part arising out of #1310.

 * **`LoggerEnrichmentConfiguration.Wrap(...)`** - mirrors `LoggerSinkConfiguration.Wrap()`, enabling enrichers to be wrapped; the two cases below illustrate its utility; it's public so that e.g. _Serilog.Filters.Expressions_ could implement an expression-based `Enrich.When()`.
 * **`Enrich.When(condition, configuration)`** - apply an enricher only to events that match `condition`; this will enable the requirement in #1310 to be met.
 * **`Enrich.AtLevel(level|switch, configuration)`** - apply an enricher only to events at or above a specified level (this differs from the semantics of `AtLevel()` discussed in the linked issue). I initially thought `FromLevel()` would be more consistent with `Console()`'s `standardErrorFromLevel:`, but the `From` prefix is confusing alongside `Enrich.FromLogContext()`.
 * **`WriteTo.Conditional(condition, configuration)`** - write events matching `condition` to a sink. We already have the shortcut `restrictedToMinimumLevel` options on sink methods; this method is similar in spirit, and avoids an inefficient and verbose `WriteTo.Logger(Filter...WriteTo)` group. Also a difficult naming problem, I'd initially thought to call this `When()` to match the enricher equivalent, but the `WriteTo.Noun` pattern is distinct from `Enrich.ActionPhrase` pattern and using the same name here read strangely. `Conditional()` is the most noun-y option I could come up with, suggestions appreciated.

Of these, the enrichment changes enable possibilities that can't be achieved with Serilog 2.8.0, and I think they're more obviously a win for that.

`WriteTo.Conditional()` seems closer to borderline as it's essentially sugar; the efficiency gain (not having to allocate copied `LogEvents` along with their `Properties` dictionaries) tips the scales in favor of including it, IMO.

Opinions and suggestions most welcome! :-)

No; minor version bump.

**Other information**:


merbla:
🚢 IT

PR #1322, opened 2019-06-14
Fix #1321: just use the most recent SDK version available.

### LINKED ISSUE #1321: The CI builds take too much time.
AppVeyor builds take a solid quarter of an hour, while all they have is to build and package one project.

This happens because before building, the CI process downloads the entire .NET Core SDK to match the exact version specified in `global.json`. As a sidenote, the version currently used, .NET Core SDK 2.2.300 is [not included in AppVeyor](https://www.appveyor.com/docs/windows-images-software/#net-framework).

A solution would be to entirely remove this mechanism, and simply use the latest .NET Core SDK version that is available. It significantly lessens the burden for new programmers who might not have that particular version installed. Besides, the syntax is mostly the same; at least for the features we require.

I will get a PR ready.

teo-tsirpanis:


nblumhardt:
Thanks for the PR! As mentioned on the issue discussion, we've landed here because of problems relying on the default SDK version. Appreciate the help, but we'll stick with things as-is just for the time being.

PR #1325, opened 2019-06-20
Respect explicit format specifiers even when JSON defaults are selected

### LINKED ISSUE #1323: Numeric format specifier not respected
- [x] Bug
- [ ] New Feature

Serilog 2.8.0

- [ ] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [x] 4.6.x
- [ ] 4.5.x

The hex numeric format specifier does not work unless supplied using standard C# string interpolation. This means I cannot use structured logging if I wish to retain the formatted string in the standard log output.

```
int status = 16;

log.Error($"Status [0x{status:X8}]"); // "Status [0x00000010]"
log.Error("Status [0x{status:X8}]", status); // "Status [0x16]"
log.Error("Status [0x{0:X8}]", status); // "Status [0x16]"
```

```
int status = 16;

log.Error($"Status [0x{status:X8}]"); // "Status [0x00000010]"
log.Error("Status [0x{status:X8}]", status); // "Status [0x00000010]"
log.Error("Status [0x{0:X8}]", status); // "Status [0x00000010]"
```
Apologies in advance if this is not how it's meant to work. I tried looking for documentation addressing this specific case and couldn't (easily) find it, however, https://github.com/serilog/serilog/wiki/Writing-Log-Events states:

> Message templates are a superset of standard .NET format strings, so any format string acceptable to string.Format() will also be correctly processed by Serilog

nblumhardt:
Fixes #1323 

We should check this against the behavior of https://github.com/serilog/serilog-sinks-console/blob/dev/src/Serilog.Sinks.Console/Sinks/SystemConsole/Rendering/ThemedMessageTemplateRenderer.cs before merging.

### DIFF
--- a/src/Serilog/Rendering/MessageTemplateRenderer.cs
+++ b/src/Serilog/Rendering/MessageTemplateRenderer.cs
@@ -95,7 +95,7 @@ namespace Serilog.Rendering
             {
                 output.Write(str);
             }
-            else if (json)
+            else if (json && format == null)
             {
                 JsonValueFormatter.Format(propertyValue, output);
             }
--- a/test/Serilog.Tests/Rendering/MessageTemplateRendererTests.cs
+++ b/test/Serilog.Tests/Rendering/MessageTemplateRendererTests.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using Serilog.Events;
+using Serilog.Parsing;
+using Serilog.Rendering;
+using Xunit;
+
+namespace Serilog.Tests.Rendering
+{
+    public class MessageTemplateRendererTests
+    {
+        readonly MessageTemplateParser _messageTemplateParser = new MessageTemplateParser();
+
+        [Theory]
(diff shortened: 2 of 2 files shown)


nblumhardt:
Okay, so, the mechanism that _Console_ uses here is different; at the top level, it falls back to non-JSON "display" formatting:

https://github.com/serilog/serilog-sinks-console/blob/dev/src/Serilog.Sinks.Console/Sinks/SystemConsole/Formatting/ThemedJsonValueFormatter.cs#L47

Which in turn ends up calling `ScalarValue.Render()`:

https://github.com/serilog/serilog-sinks-console/blob/dev/src/Serilog.Sinks.Console/Sinks/SystemConsole/Formatting/ThemedDisplayValueFormatter.cs#L184

We could try lining these up, but I think rather than try to keep them 100% in sync, it's okay as long as each behaves predictably, at least for now.

A better future design might be to adopt the `ConsoleTheme.None` formatting into the core _Serilog_ package, somehow, and share the code.

For now, I think this is ready to go.

nblumhardt:
Thanks @tsimbalar ! 👍

Issue #1331, opened 2019-06-24
[New Feature] Limit boxing for Logger Write methods


Scooletz:
AFAIK Logger now skips boxing and creating an `object[]` if the log level is not enabled. I tried to follow the whole path and eventually, all the calls to `.Write` end in https://github.com/serilog/serilog/blob/dev/src/Serilog/Core/Logger.cs#L361-L375 which effectively boxes values and allocates the mentioned array. I followed the flow which is 

`Logger.Write` -> 
 `MessageTemplateProcessor.Process` -> 
`PropertyValueConverter.CreatePropertyValue`

which potentially could have a generic overload for `CreatePropertyValue` and then allow to propagate the unboxing to the calls above. 

Would that be something interesting for you? I think I could try to spike it and then we could have a follow up discussion under this issue.

This is somewhat related to [my work on extending the public seam](https://github.com/aspnet/Extensions/pull/1874) of `Microsoft.Extensions.Logging` and finding that eventually Serilog will box and allocate anyway.

nblumhardt:
I think with this one still up in the air, it might be best to close it here, and discuss via the PR when we have something to go on. Cheers!

Scooletz:
Sure @nblumhardt This issue was created in the "issue first, PR second" manner 😄

PR #1332, opened 2019-06-24
Limit boxing for Logger Write methods

Scooletz:
This PR addresses https://github.com/serilog/serilog/issues/1331 by limiting the boxing of logging parameters and passing down the strongly-typed `struct` values till the creation of log values. This should reduce the burden of boxing and allocating arrays.

### Benchmarks

Initial benchmarks, based on `AllocationBenchmark` enhanced with two more primitive values are shows below. There's some regression for the `LogAnonymous` case but the scalar cases for `int` and `Guid` look promising.

```ini
BenchmarkDotNet=v0.11.5, OS=Windows 10.0.17134.829 (1803/April2018Update/Redstone4)
Intel Core i7-6700HQ CPU 2.60GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
Frequency=2531247 Hz, Resolution=395.0622 ns, Timer=TSC
.NET Core SDK=2.2.300
  [Host]     : .NET Core 2.2.5 (CoreCLR 4.6.27617.05, CoreFX 4.6.27618.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.2.5 (CoreCLR 4.6.27617.05, CoreFX 4.6.27618.01), 64bit RyuJIT
```

#### Before

|               Method |        Mean |       Error |      StdDev |  Gen 0 | Allocated |
|--------------------- |------------:|------------:|------------:|-------:|----------:|
|             LogEmpty |    10.60 ns |   0.2118 ns |   0.1878 ns |      - |         - |
| LogEmptyWithEnricher |    75.18 ns |   1.0733 ns |   0.9515 ns | 0.0178 |      56 B |
|            LogScalar |   900.97 ns |  10.3762 ns |   9.1982 ns | 0.1192 |     376 B |
|         LogScalarInt |   942.98 ns |  18.8135 ns |  19.3200 ns | 0.1268 |     400 B |
|        LogScalarGuid |   948.60 ns |  11.2454 ns |  10.5190 ns | 0.1287 |     408 B |
|        LogDictionary | 4,715.25 ns |  51.9326 ns |  48.5778 ns | 0.7095 |    2240 B |
|          LogSequence | 2,149.29 ns |  25.7450 ns |  21.4982 ns | 0.2594 |     824 B |
|         LogAnonymous | 9,221.81 ns | 152.1356 ns | 142.3078 ns | 1.0834 |    3440 B |

#### After

|               Method |        Mean |       Error |      StdDev |      Median |  Gen 0 | Allocated |
|--------------------- |------------:|------------:|------------:|------------:|-------:|----------:|
|             LogEmpty |    10.81 ns |   0.2701 ns |   0.7663 ns |    10.71 ns |      - |         - |
| LogEmptyWithEnricher |    84.76 ns |   2.9564 ns |   8.5770 ns |    82.16 ns | 0.0178 |      56 B |
|            LogScalar |   982.95 ns |  25.4815 ns |  73.5199 ns |   958.22 ns | 0.1087 |     344 B |
|         LogScalarInt |   868.16 ns |  17.6402 ns |  49.7546 ns |   846.40 ns | 0.1116 |     352 B |
|        LogScalarGuid |   865.51 ns |  12.0206 ns |  11.2440 ns |   862.36 ns | 0.1135 |     360 B |
|        LogDictionary | 4,488.64 ns |  87.5998 ns | 100.8801 ns | 4,487.15 ns | 0.7095 |    2240 B |
|          LogSequence | 2,130.47 ns |  23.7650 ns |  19.8449 ns | 2,125.31 ns | 0.2594 |     824 B |
|         LogAnonymous | 8,870.72 ns | 175.2878 ns | 256.9346 ns | 8,834.51 ns | 1.0834 |    3440 B |

I don't know yet, but it's possible. It's WIP.

- [ ] The commit follows our [guidelines](https://github.com/serilog/serilog/blob/dev/CONTRIBUTING.md)
- [ ] Unit Tests for the changes have been added (for bug fixes / features)

**Other information**:

This requires a lot of performance testing and validating design decisions for potential breaking changes.


nblumhardt:
Interesting exploration - thanks for taking the time to dig into this. It will be interesting to see what kind of allocation/throughput impact it has.

Scooletz:
@nblumhardt Currently ~100 tests are broken and it would require some more work. The question is, whether this looks interesting enough to push forward. For scalar/primitive cases it allocates 10% less and is at least 10% faster.

I could try to polish it a little bit, take a look at dictionary case and go through the tests. I want to note that because of introduction of the new non-boxing `ScalarValue<T>` this would be a breaking change (major in semver) even if played with subclassing `ScalarValue` (adding `virtual` is also breaking)

nblumhardt:
I think one of the follow-on challenges of `ScalarValue<T>` will be that, even if we can shoehorn it in, in a nonbreaking way (I think there's an option involving a new virtual property on `ScalarValue` that `Value` would call, and `ScalarValue<T> would override), we'd need to provide a nongeneric `Value` for reflection-based code to consume.

In that case, the underlying value would have to be boxed, and if we called `Value` more than once, we might end up boxing up the same scalar multiple times, and defeat the purpose of saving the allocations in the first place 🤔 ..........which could itself be worked around by caching the value, but we're then increasing the size of `ScalarValue` itself........

Intriguing though! I need a bit more time to mull it over.... 😄

Scooletz:
Yeah, we could `object _cache` it. I was thinking about it.

We can't use `virtual`. The call to the site can be `call` not `callvirt` and therefore deploying a new version of serilog would call the base, not the overriden. Recompilation is needed as far as a I see it.

No rush. Let's keep discussions in here. I'll pause my work as fixing that many tests and deciding about major is not a minor thing 😜

nblumhardt:
Thanks for the reply :+1: - we couldn't make `Value` virtual, but I had something more like this in mind:

```csharp
class ScalarValue : // Etc. - details omitted
{
    object _value;

    // It's nonbreaking to introduce a new virtual method
    protected virtual object GetBoxedValue() => _value;

    // Nonbreaking to change impl to call a virtual method
    public object Value => GetBoxedValue();
}

class ScalarValue<T> : ScalarValue
{
    T _value;

    public ScalarValue(T value)
        : base(null) // <- No box
    {
    }

    // Not even sure there would be any situations where this could actually 
    // be consumed..
    public T UnboxedValue => _value;

    // Boxes here
    protected override object GetBoxedValue() => _value;
}
```

Thus `new ScalarValue(1)` would get us what we have today; `new ScalarValue<int>(1)` would avoid the box and still be usable as a `ScalarValue`.

The challenge would be converting code to find and use methods on `ScalarValue<T>` to do interesting things without having to touch `Value`. Since the code's likely to be reflection-based, it's hard to avoid boxing - the best we could do would be to expose interfaces like `IFormattable`, `IJsonFormattable` and so-on and call through those to perform tasks rather than access the value directly. This wouldn't be particularly open to extension, though.

Scooletz:
I see. Let me play with this and see how much I can push forward without spending a lot of time on fixing tests.

Scooletz:
@nblumhardt I moved pieces a little bit. Now the fast path is only for the first call of the `PropertyValueConverter`. Additionally, fast builders, as I call them, are cached in a custom hopscotch hashmap that ensure a fast lookup. 

Now, there are cases where `ScalarValue` is used for enums, if they are not registered as scalars or `ScalarValue<T>` if they are. This introduces dichotomy but there are some gains for primitive types.

Please take a look and comment on it.

Scooletz:
I added IJsonFormattable for several types and rerun benchmarks but without any `Sink` that would validate the non-boxing nature of IJsonFormattable (and somewhat faster dispatch). 

Looking forward to hearing from maintainers.

Scooletz:
Any input on this @nblumhardt ? 😉

nblumhardt:
Howdy! :-) .. I can't add a lot without seeing how the numbers came out, was the final result similar to the initial benches? Cheers!

nblumhardt:
Hi @Scooletz - how'd this one turn out? Did you end up with numbers compelling enough to warrant taking it further? Sorry about the lack of input, it's tricky to keep up with this kind of intricate work without a lot of time to dedicate :-)

Scooletz:
>  it's tricky to keep up with this kind of intricate work without a lot of time to dedicate

I think I landed in the same area 😄 Let me close it.

PR #1348, opened 2019-07-29
INFLUENCES.md

### LINKED ISSUE #1347: INFLUENCES.md
A recent Twitter thread got me thinking about the poor job we (and software projects in general) have done of crediting our influences. If it wasn't for log4net, SLAB, .NET format strings, and their own inspirations, Serilog would not have come into being.

Copyright leaves its mark in file headers; books and papers cite their sources; as far as I'm aware, there's no similarly-widespread mechanism for documenting the lineage of ideas in a codebase (patents having entirely different goals and application).

Trawling through some Google results led me to https://github.com/rust-lang-nursery/reference/blob/master/src/influences.md - influences contributing to the Rust programming language.

Taking that a step further, and echoing other useful de-facto standards like `README`, `LICENSE`, and `CODE_OF_CONDUCT`, perhaps we could include a top-level `INFLUENCES.md`?

```markdown
# Influences

Serilog draws directly upon ideas found in:

 * [Apache log4net](https://logging.apache.org/log4net/) &mdash; `ILogger` interface, levels, types/namespaces as log sources
 * [Enterprise Library Semantic Logging Application Block](https://github.com/microsoftarchive/semantic-logging) (SLAB) &mdash; logs-as-event-streams
 * [.NET Framework](https://dot.net) &mdash; format string syntax
```

It would be nice to see something like this recorded for more OSS libraries in .NET. I'm keen to do the same for a couple of other projects I'm involved in so that the links to their predecessors aren't lost to time.

I think for this to be useful , though:

 - It would have to be limited to the few primary influences on a project, otherwise it's an unmanageable task
 - It needs to record only subjective, _direct_ influences, i.e. not necessarily the original sources, to avoid inaccuracies and time-wasting/contentious debate (future software philologists can worry about this!)

I'd PR this to the README, but it would be in danger of creating distracting noise for newcomers looking for a quickstart.

Any thoughts/suggestions/additions/criticisms?

nblumhardt:
Acknowledge the major sources of ideas and inspiration.

Resolves #1347

### DIFF
--- a/INFLUENCES.md
+++ b/INFLUENCES.md
@@ -2,6 +2,6 @@
 
 Serilog draws directly on ideas found in:
 
- * [.NET Framework](https://dot.net) &mdash; format string syntax
+ * [.NET Framework](https://dot.net) &mdash; format string syntax adapted for message templates
  * [Apache log4net](https://logging.apache.org/log4net/) &mdash; `ILogger` interface, levels, types/namespaces as log sources
  * [Enterprise Library Semantic Logging Application Block](https://github.com/microsoftarchive/semantic-logging) (SLAB) &mdash; logs-as-event-streams
--- a/INFLUENCES.md
+++ b/INFLUENCES.md
@@ -1,0 +1,7 @@
+# Influences
+
+Serilog draws directly on ideas found in:
+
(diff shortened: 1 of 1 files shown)


nblumhardt:
Great, thanks 👍

Issue #1351, opened 2019-08-02
Optionally write final layer of ToMaximumDepth as ScalarValue instead "Null"


HeartofTheForce:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [x] New Feature

**What version of Serilog is affected?  Please list the related NuGet package.**
2.8.0 

**Please describe the current behavior?**
Currently reaching maximum depth sets all values on that layer to "Null"

**Please describe the expected behavior?**
Have an optional flag to serialize final depth as using ScalarValue types e.g.
`.Destructure.ToMaximumDepth(3, writeFinalDepthAsScalar: true)`

nblumhardt:
That's an excellent idea, thanks - if it had been suggested early on, we'd have made this the default, I think. We still might want to consider changing the default, if we can implement this seamlessly enough (a 3.0 feature?).

HeartofTheForce:
@nblumhardt I've made a PR for this here #1357

nblumhardt:
Hi @HeartofTheForce - I notice you closed your PR; thanks for the help all the same! I'll mark this as up-for-grabs, let me know if you are interested in picking it back up.

Jlalond:
@HeartofTheForce any reason you deleted your PR @nblumhardt I'm looking into this btw.

nblumhardt:
Seems as though this one is stale; still a really interesting idea if anyone decides to look into it, but I'll close this as it's now been quite some time without activity.

PR #1352, opened 2019-08-03
Feature/optionally write final depth as scalar

HeartofTheForce:
#1351

Probably not. Some unit tests were passing when they possibly shouldn't be.

**Other information**:
Unsure about the Destructuring enum being changed to a flag, but found what seems like bugs because of it specifically:
- `ValueTuplesAreRecognizedWhenDestructuring` was acting despite destrucuturing set implicitly to false
- `AllTupleLengthsUpToSevenAreSupportedForCapturing` also acting despite destrucuturing being implicitly false
- `MaximumStringLengthNOTEffectiveForString` ~~passing despite being give a 4 length string and a 3 length limit~~ According to method description this should only apply during destructuring

### DIFF
--- a/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
@@ -154,9 +154,9 @@ namespace Serilog.Configuration
         }
 
         /// <summary>
-        /// When destructuring objects, depth will be limited to 5 property traversals deep to
+        /// When destructuring objects, depth will be limited to 10 property traversals deep to
         /// guard against ballooning space when recursive/cyclic structures are accidentally passed. To
-        /// increase this limit pass a higher value.
+        /// change this limit pass a new maximum depth.
         /// </summary>
         /// <param name="maximumDestructuringDepth">The maximum depth to use.</param>
         /// <param name="writeFinalDepthAsScalar">When set to true write the final depth as scalar instead of null.</param>
--- a/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
@@ -27,15 +27,17 @@ namespace Serilog.Configuration
         readonly LoggerConfiguration _loggerConfiguration;
         readonly Action<Type> _addScalar;
         readonly Action<IDestructuringPolicy> _addPolicy;
-        readonly Action<int> _setMaximumDepth;
+        readonly SetMaximumDepthDeletegate _setMaximumDepth;
         readonly Action<int> _setMaximumStringLength;
         readonly Action<int> _setMaximumCollectionCount;
+
+        internal delegate void SetMaximumDepthDeletegate(int depth, bool writeFinalDepthAsScalar);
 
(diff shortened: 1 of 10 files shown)


src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
```diff
@@ -147,7 +149,23 @@ public LoggerConfiguration ByTransforming<TValue>(Func<TValue, object> transform
         public LoggerConfiguration ToMaximumDepth(int maximumDestructuringDepth)
         {
             if (maximumDestructuringDepth < 0) throw new ArgumentOutOfRangeException(nameof(maximumDestructuringDepth));
-            _setMaximumDepth(maximumDestructuringDepth);
+            _setMaximumDepth(maximumDestructuringDepth, false);
+            return _loggerConfiguration;
+        }
+
+        /// <summary>
+        /// When destructuring objects, depth will be limited to 5 property traversals deep to
```
rob-earwaker: The default depth is actually 10. See updated summary for the [`ToMaximumDepth(int maximumDestructuringDepth)`](https://github.com/serilog/serilog/blob/dev/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs#L139) overload on dev.

HeartofTheForce: I've updated the summary to match the latest version

nblumhardt:
Thanks for the 👍 

This one will take some time to work through; I'll try to carve out some time for it ASAP, but am on vacation for the next two weeks and so only online intermittently.

The tests that failed after the changes were, I think, by design - Serilog always disassembles tuples whether destructuring is enabled or not, for example. AFAIK there shouldn't be any breaking changes required for this design (with some careful overloading of `ToMaximumDepth()`), but if you think you have found some bugs we should raise those separately.

It would also be great if we can avoid changing `Destructuring` - I think the changes there might be an interesting design space to explore at some point, but keeping this PR as narrow and tactical as possible will help to get the changes through.

HTH!

Nick

nblumhardt:
Hi @HeartofTheForce - hope this finds you well! Did you have a chance to consider my comment above, and are you planning to take this one further? Cheers!

HeartofTheForce:
I'm happy to strip out the `Destructuring` changes, only reasons it was added was because I couldn't find a clean way to reuse `PropertyValueConverter.CreatePropertyValue` to only write the final `Scalar` values while avoiding destructuring any types that would result in the maximum depth being exceeded.

Would you prefer having `PropertyValueConverter.CreatePropertyValue`'s method signature changed to force all values to `Scalar` for the final depth?

I can also move the `Destructuring` enum changes to another PR so it can be reviewed/considered separately and then move from there/

nblumhardt:
Thanks for the update! I don't think a separate PR for the destructuring changes would be accepted at this stage, since the role of the type is to publicly represent the syntax of message templates, and making these changes would cause us to drift away from that.

Is there any way we could get the `_writeFinalDepthAsScalar` flag in as a constructor argument to `DepthLimiter`, and implement the behavior there? https://github.com/serilog/serilog/blob/dev/src/Serilog/Capturing/DepthLimiter.cs#L45

PR #1357, opened 2019-08-30
Update summary for default destructuring depth

rob-earwaker:
The summary for the [`LoggerDestructuringConfiguration.ToMaximumDepth`](https://github.com/serilog/serilog/blob/dev/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs#L140) method incorrectly states that the default maximum destructuring depth is 5, when in fact the [default is 10](https://github.com/serilog/serilog/blob/dev/src/Serilog/LoggerConfiguration.cs#L41).

There is no Issue raised for this. Since it does not change any implementation and is such a minor change I assumed it would not be necessary. Let me know if this is not the case and I'll happily add one!

No. This is a documentation change only.

**Other information**:
As well as updating the summary, I've also added a test to confirm that the default destructuring depth of 10 is honoured.

### DIFF
--- a/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
@@ -137,9 +137,9 @@ namespace Serilog.Configuration
         }
 
         /// <summary>
-        /// When destructuring objects, depth will be limited to 5 property traversals deep to
+        /// When destructuring objects, depth will be limited to 10 property traversals deep to
         /// guard against ballooning space when recursive/cyclic structures are accidentally passed. To
-        /// increase this limit pass a higher value.
+        /// change this limit pass a new maximum depth.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
```diff
@@ -137,9 +137,9 @@ public LoggerConfiguration ByTransforming<TValue>(Func<TValue, object> transform
         }
 
         /// <summary>
-        /// When destructuring objects, depth will be limited to 5 property traversals deep to
+        /// When destructuring objects, depth will be limited to 10 property traversals deep to
         /// guard against ballooning space when recursive/cyclic structures are accidentally passed. To
-        /// increase this limit pass a higher value.
+        /// change this limit pass a new maximum depth.
```
nblumhardt: :+1:

PR #1373, opened 2019-10-17
Enrich.ByRemovingProperty(name)

nblumhardt:
Frameworks set up for structured logging often include a range of contextual properties in log events, via `LogContext.PushProperty()`, or `ILogger.BeginScope()`. Not all applications need or want all of these.

`Enrich.ByRemovingProperty()` addresses this:

```csharp
    .Enrich.ByRemovingProperty("Test")
    .CreateLogger();

Log.ForContext("Test", 1).Information("Hello");
// --> no `Test` property appears in the output
```

The enricher will remove `Test` even if it appears in the message template:

```csharp
Log.ForContext("Running {Test}", testName);
```

This is the default behavior because a) it's cheap, and b) it's safer when the property is (questionably) something like `CreditCardNumber` (do not do this 🤨).

To check the message template, and retain the property when it appears in it, an additional flag can be set:

```csharp
    .Enrich.ByRemovingProperty("Test", retainWhenInMessage: true)
```

(The name of this flag could leave room for improvement.)

No

**Other information**:

CC @rocklan

### DIFF
--- a/src/Serilog/Configuration/LoggerEnrichmentConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerEnrichmentConfiguration.cs
@@ -84,6 +84,18 @@ namespace Serilog.Configuration
         }
 
         /// <summary>
+        /// Remove the specified property from events.
+        /// </summary>
+        /// <param name="name">The name of the property to remove.</param>
+        /// <param name="retainWhenInMessage">If <c>true</c>, the property will be removed only when
+        /// it doesn't appear in the event's message template; otherwise, the property will be removed regardless.</param>
+        /// <returns></returns>
+        public LoggerConfiguration ByRemovingProperty(string name, bool retainWhenInMessage = false)
+        {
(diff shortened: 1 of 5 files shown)


nblumhardt:
Any thoughts on this one? I'm on the fence about it - in many cases it's better to just not add unwanted properties - but there are a few situations (mostly when consuming logs from libraries like ASP.NET Core) where property removal would help control event size. Any more :+1: or :-1:?

bddckr:
I have the need to ignore properties, but not actually yet on a root-level so to speak. I'm not yet at the state where I want to control specific messages to decrease the log size, but I can definitely see how this change would become useful in the future.

This feature seems like it can go hand-in-hand with [destructurama/by-ignoring](https://github.com/destructurama/by-ignoring). So together a user has a way to ignore properties as a whole, as well as properties' properties (😅), which seems like all you ever want besides filtering a whole message or at a log level.

nblumhardt:
I'm closing this for now, though we may revisit in the future :-)

PR #1385, opened 2019-11-13
Adds the ability to filter object properties recorded in StructureValue

sungam3r:

New feature for filtering properties.

No.

### DIFF
--- a/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
@@ -14,6 +14,7 @@
 
 using System;
 using System.Collections;
+using System.Reflection;
 using Serilog.Core;
 using Serilog.Policies;
 
@@ -30,6 +31,7 @@ namespace Serilog.Configuration
         readonly Action<int> _setMaximumDepth;
         readonly Action<int> _setMaximumStringLength;
         readonly Action<int> _setMaximumCollectionCount;
+        readonly Action<Func<PropertyInfo, object, bool>> _setDestructuringPropertyFilter;
 
         internal LoggerDestructuringConfiguration(
             LoggerConfiguration loggerConfiguration,
@@ -37,7 +39,8 @@ namespace Serilog.Configuration
             Action<IDestructuringPolicy> addPolicy,
             Action<int> setMaximumDepth,
             Action<int> setMaximumStringLength,
-            Action<int> setMaximumCollectionCount)
+            Action<int> setMaximumCollectionCount,
+            Action<Func<PropertyInfo, object, bool>> setDestructuringPropertyFilter)
         {
             _loggerConfiguration = loggerConfiguration ?? throw new ArgumentNullException(nameof(loggerConfiguration));
             _addScalar = addScalar ?? throw new ArgumentNullException(nameof(addScalar));
@@ -45,6 +48,7 @@ namespace Serilog.Configuration
             _setMaximumDepth = setMaximumDepth ?? throw new ArgumentNullException(nameof(setMaximumDepth));
             _setMaximumStringLength = setMaximumStringLength ?? throw new ArgumentNullException(nameof(setMaximumStringLength));
             _setMaximumCollectionCount = setMaximumCollectionCount ?? throw new ArgumentNullException(nameof(setMaximumCollectionCount));
+            _setDestructuringPropertyFilter = setDestructuringPropertyFilter ?? throw new ArgumentNullException(nameof(setDestructuringPropertyFilter));
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


nblumhardt:
Thanks!

These cases overlap quite a bit with `IDestructuringPolicy` (exclude properties on types that match a condition) and `ILogEventEnricher` (modify the members of structure values attached to events).

Perhaps shipping an extension package (_Serilog.Destructure.ByFiltering_?) to gauge the benefits/usage would be a good intermediate step before we consider baking this in directly?

sungam3r:
I agree that partially the functionality overlaps. If you look from the side, then in the serialog many things can be solved in different ways. Another thing is that to solve it will have to make more or less effort.

> These cases overlap quite a bit with `IDestructuringPolicy` (exclude properties on types that match a condition)

Yes it's very close to `IDestructuringPolicy`. In fact, this is the default destructuring policy setting for `TryConvertEnumerable`/`TryConvertValueTuple`/`TryConvertCompilerGeneratedType` family of methods. At least I consider this setting in such way. For me, the main drawback of using custom `IDestructuringPolicy` to the proposed solution is that by implementing the `IDestructuringPolicy` **I will be forced to process any object of an arbitrary level of complexity**, `PropertyValueConverter` is a private class. Instead, a much cheaper solution is not to repeat the internal machinisms of the serilog, but to complement them with the missing setting.

>  and `ILogEventEnricher`

I already wrote an enricher (anti-enricher) which deals with the removal of properties by some criteria. This is a working solution, but I do not like it. Enricher works with the serilog object model after the serialog has processed the original object. It would be more convenient to work with the source object instead.

> Perhaps shipping an extension package (Serilog.Destructure.ByFiltering?) to gauge the benefits/usage would be a good intermediate step before we consider baking this in directly?

Well, I don’t think that making a separate package for such a small change is rational intermediate step, especially if its code is merged into this project. Honestly, I look at the current serilog configuration api and see that the new setting fits perfectly into it. Don't you think so?

src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
```diff
@@ -180,5 +184,23 @@ public LoggerConfiguration ToMaximumCollectionCount(int maximumCollectionCount)
             _setMaximumCollectionCount(maximumCollectionCount);
             return _loggerConfiguration;
         }
+
+        /// <summary>
+        /// When destructuring objects, allows to filter properties recorded in StructureValue.
+        /// </summary>
+        /// <param name="filter">Delegate used for filtering.</param>
+        /// <returns>Configuration object allowing method chaining.</returns>
+        public LoggerConfiguration AllowOnly(Func<PropertyInfo, object, bool> filter)
```
sungam3r: @nblumhardt I could not find a really good name for these two extension methods.

nblumhardt:
Thanks for your reply!

> Well, I don’t think that making a separate package for such a small change is rational intermediate step, especially if its code is merged into this project. Honestly, I look at the current serilog configuration api and see that the new setting fits perfectly into it. Don't you think so?

Yes and no :-) ... it's a totally reasonable addition, but over time the reasonable additions add up and it gets very hard to keep widely-used libraries from getting bloated. Serilog "core" has reached a point where, unless there are many requests for the same feature over a long period, we tend to have a default "no" to most additions, especially when the existing plug-points enable them to be implemented as extensions (for this one, the destructuring policy approach is viable, and we already support a bunch of similar extensions, e.g. https://github.com/destructurama/attributed/blob/dev/src/Destructurama.Attributed/Attributed/AttributedDestructuringPolicy.cs#L26 )

Putting improvements out as separate packages is a big win for the long-term viability of Serilog, since it distributes the maintenance work and keeps the core codebase manageable - it's one of the reasons Serilog has stayed pleasant to work with.

I still maintain a bunch of extension packages for things I personally like and use - e.g. the "Destructurama" series of packages, _SerilogTimings_, and a few more here and there.

sungam3r:
> For me, the main drawback of using custom IDestructuringPolicy to the proposed solution is that by implementing the IDestructuringPolicy I will be forced to process any object of an arbitrary level of complexity, PropertyValueConverter is a private class. Instead, a much cheaper solution is not to repeat the internal machinisms of the serilog, but to complement them with the missing setting.

What about this? Copy-paste?

nblumhardt:
The `ILogEventPropertyValueFactory` parameter passed to `TryDestructure()` exposes the majority of what `PropertyValueConverter` is supplying; the destructuring policies are plug-ins into `PropertyValueConverter`, so although there's some duplicated responsibility (enumerating public properties of an object and getting values) the complex stuff is not duplicated.

The policy in https://github.com/destructurama/attributed/blob/dev/src/Destructurama.Attributed/Attributed/AttributedDestructuringPolicy.cs#L31 provides attribute-driven destructuring, and uses runtime caching, in ~50 significant lines of code. 

Some duplication/effort in the ecosystem again helps spread the load of maintenance - every line added to core Serilog accumulates, and centralizes the work.

Given the choice between adding one line to Serilog, or 100 lines in an extension, we just about always would choose the extension - at least to gauge utility, anyway. I know it's somewhat limiting and seems to limit how much Serilog can be improved in the short term, but over the long term it's been a useful strategy so far.

sungam3r:
Ok let's go this way. I will be back with the result after a while.

sungam3r:
I looked at the project you were talking about - destructurama. I realized that I can use `ILogEventPropertyValueFactory`, this interface is implemened by `PropertyValueConverter`. And the policy method will be called recursively. Thanks!

I hope this PR destructurama/attributed#35 will be useful.

PR #1402, opened 2020-02-13
Feature/creating rendered json formatter

Everwig:
…/issues/1401

Missing RenderedJsonFormatter 

No

**Other information**:
Nothing clever, I'm not trying to take credit or anything. All the work is based on the RenderedCompactJsonFormatter from the Serilog.Formatting.Compact NuGet

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+2.9.1
+* #1401 - implementing RenderedJsonFormatter
+
 2.7.1
  * #1075 - overload of `LoggerSinkConfiguration.Wrap()` accepting `LoggingLevelSwitch`
  * #1083 - update package metadata, including `https://` project and icon URLs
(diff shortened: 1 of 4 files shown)


bartelink:
Any particular reason you feel this needs to go in the core repo / assembly?

Everwig:
Well, first I thought it should be going under the Serilog.Formatting.Compact, but to me, this is not a compact format so, the obvious choice would be under the Serilog.Formatting.Json.

If you feel that it should be moved there, then I am happy to oblige. :)

bartelink:
I am but a lurker - I agree putting it compact is debatable; maybe the repo needs renaming too, but then there's already a focused package (could the repo make 2 nugets?)

My main point though, is that for me there should be a very high bar to put something into Serilog's core, and I'm not convinced this in particular meets that bar

nblumhardt:
Thanks for sending the PR! I've responded in your original issue thread:

> Thanks for the note. In the linked readme, RenderedJsonFormatter is a benchmark name, not a type name. I've updated the readme to fix this.

> We already support this through `new JsonFormatter(renderMessage: true)`; the format used by `JsonFormatter` is heading towards "legacy" so we will probably not make this change.

Thanks again, though - appreciate you taking the time to send a PR.

Issue #1409, opened 2020-03-16
Nonscalar (structured) properties with format specificer are passed to the sink


snuup:
- [x] New Feature

**current behavior**
all non scalars are destructed. their format specifiers are ignored.

**new behavior**
nonscalars that have a format specifier are not destructed

     logger.Information("{persons:ids}", new[] {11, 22, 33}) 
     // currently the format would be ignored and the array would be destructured. 

this allows late resolving of scalar-collections into meaningful diagnostic text, like resolving object-ids into object-titles. the log message above could be formatted like so:

    "[11]adam, [22]beatrice, [33]claude"

by configuring a Formatprovider on the sink that resolved the person.ids. Such an example is provided in the unit tests in the pull request.

a pull request is prepared: #1410

PR #1410, opened 2020-03-16
Structured types with associated format are not destructured but forw…

snuup:
…arded to the sink.

Structured logging in Serilog allows passing the properties of structured values as separate scalars to sinks, so that log-stores can provide searches for the values of those scalar property values.

### Using Serilog to resolve object-ids at the logging stage
I tried to use the logging pipeline for a different purpose. Many systems process objects by their id, person-ids and so on. Often, reading logs that hold those ids only is cumbersome. Replacing the ids with names can be done with passing down resolver functions like so:

    getuserprofile (int personid, diagnosticgetusername)
        ... log(  ... diagnosticgetusername(personid) ... )

passing around diagnostic resolver functions in that way clutters the apis, adding complexity to the code.

Serilog greatly helps here. First I tried to use custom Destructurers like so:

   ... log("   {@personid} ", new PersonId( 42) ...

but this requires wrapping the scalar value and adding the @ symbol. Finally I realised that a custom formatter is more appropriate. The design I cam up with is the register resolver functions for format-names, so I can do:

     ... log("   {person:id} ", 42 ...  // format "id" relates to a resolver function configured below

CustomFormatter.Formatters["id"] = id => getname(id) // register resolver in the logger configuration

### Respecting a format specifier for non scalars - preventing destructuring
So far that works with the current Serilog version. I also have collections of ids and would like to format them in a similar way:

    ... log("   {persons:ids}  ", persons ) // where persons: int[]

At this point, the current destructure logic of Serilog interferes: The property is destructured and the format is ignored. To get around this I propose this pull request. A unit test describes the scenario of such logger-stage resolver functions and how they work with collections and arbitrary objects. The code changes should speak for themselves.

I tried to make the least possible changes and in the sense of the current code base.

### No API or relevant behavior change
Structured objects without format specifier (which would be ignored in the current version, so its highly unlikely that anyone assigned such) are destructured as before. No API changes.

related issue: #1409

### DIFF
--- a/src/Serilog/Capturing/PropertyBinder.cs
+++ b/src/Serilog/Capturing/PropertyBinder.cs
@@ -123,7 +123,7 @@ namespace Serilog.Capturing
         {
             return new EventProperty(
                         propertyToken.PropertyName,
-                        _valueConverter.CreatePropertyValue(value, propertyToken.Destructuring));
+                        _valueConverter.CreatePropertyValue(value, propertyToken.Destructuring, propertyToken.Format));
         }
     }
 }
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -96,7 +96,7 @@ namespace Serilog.Capturing
             return CreatePropertyValue(value, destructureObjects, 1);
         }
 
-        public LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructuring, string format = "")
+        public LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructuring, string format = null)
         {
             try
             {
@@ -123,7 +123,7 @@ namespace Serilog.Capturing
                 depth);
         }
 
-        LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructuring, int depth, string format = "")
+        LogEventPropertyValue CreatePropertyValue(object value, Destructuring destructuring, int depth, string format = null)
         {
             if (value == null)
                 return new ScalarValue(null);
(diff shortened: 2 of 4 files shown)


nblumhardt:
Thanks for your suggestion and PR.

This will need some careful consideration - for example, scalar values are assumed by Serilog to be safe to send across threads, so the change as currently written would potentially introduce threading issues in code that uses format specifiers incorrectly (probably not a huge number of cases, but very likely nonzero).

snuup:
Thx Nicholas

I used custom Destructors before and find this way to write the formatting code much nicer, shorter and more flexible. I realized that "late resolving" is not destructiring. I had to artificially wrap integers with structs so I can then "destruct" those structs and then resolve names for integers. Late resolving is much more (enhanced) formatting, not destructuring.

I do not see how this modified code and its usage is less thread safe than doing exactly the same with Destructors - the very same things happen in the pipeline. The only difference is how to write the logging code and the configuration. If this is has concurrency aspects then the available custom Destructor mechanism faces exactly the same issues. But maybe I miss something.

Anyway, I am *very* happy with Serilog and this modding and I tried my best to give a little back to this really nice project. Thank you for your work + keep it up.

nblumhardt:
Thanks very much for the follow-up. I think it's a solid proposal, but I don't think we should merge this at present.

The behavior essentially could be explained as "when formatting an array, pass the format string through to each element" - which is logical, but isn't generally how format strings work in .NET.

Although message templates do diverge from format strings in several ways, we try to keep these divergences extremely limited, and the pay-off in this case doesn't necessarily justify the change. It would, for example, introduce discrepancies between how message templates work in Serilog, MEL, LibLog, and NLog.

Overriding `ToString()` on the target class, or creating an array of ids to log instead of the values themselves, would be reasonable workaround (you can use `if (log.IsLevelEnabled())` to avoid unnecessary work in the latter case.

Regarding thread safety, some Serilog sinks are asynchronous; passing `value` through here, rather than the stringy (or destructured) version of it, would pass the value through directly to background threads and potentially cause issues:

```csharp
            if (!string.IsNullOrEmpty(format))
            {
                return new ScalarValue(value);
            }
```

Thanks again for the PR and discussion! Sorry we can't accept it, but it's appreciated nonetheless :+1:

PR #1414, opened 2020-03-27
update Overrides during runtime fixes #1413

### LINKED ISSUE #1413: Feature Request: Add LogLevelOverride during runtime
- [x] New Feature

Im trying to close a feature gap to Microsoft.Extensions.Logging which is able to change loglevel (override) configuration on the fly. 

im not sure if its a feature request or a bug but with the current api im not able to add overrides. updating existing overrides works fine.

```csharp

       [TestMethod]
        public void TestSerilogOverride()
        {
            var nsLogSwitch = new LoggingLevelSwitch(LogEventLevel.Warning);
            var configuration = new LoggerConfiguration()
                .MinimumLevel.Is(LogEventLevel.Warning)
                .MinimumLevel.Override("ns", nsLogSwitch);

            var logger = configuration
                .CreateLogger();
            
            var context = logger.ForContext("SourceContext", "ns.typename");

            Assert.IsFalse(context.IsEnabled(LogEventLevel.Information));
            
            nsLogSwitch.MinimumLevel = LogEventLevel.Information;
            Assert.IsTrue(context.IsEnabled(LogEventLevel.Information));
            nsLogSwitch.MinimumLevel = LogEventLevel.Warning;
            
            
            var typeLogSwitch = new LoggingLevelSwitch(LogEventLevel.Information);
            configuration.MinimumLevel.Override("ns.typename", typeLogSwitch);
            Assert.IsTrue(context.IsEnabled(LogEventLevel.Information));
        }
```

Adding new overrides should behave the same as updating an existing override. 

via:
a. add an api "UpdateOverride" to update all existing loggers after a new override is added to the configuration. 

```diff
+ no need to generate pseudo overrides
- there is currently no tracking of "All" loggers so its hard to find out what to update 
```

b. an internal change to
https://github.com/serilog/serilog/blob/8ae332d983b31044f4da0fa34e3b9cb85ba68bc9/src/Serilog/Core/LevelOverrideMap.cs#L67-L81
so that every logger creates a new override if none is existing 

```diff
- increased memory usage for users that dont use overrides
```

JanEggers:
this pr is an attempt to fix #1413

I hope not. I did not add the method to the ILogger Interface just to the default implementation.
this may be debateable?!

- [x] The commit follows our [guidelines]

**Other information**:

- I had to update one test that was failing. 
   LoggerShouldNotReferenceToItsConfigurationAfterBeingCreated

   I think I did not break it in the first place as it was just the local method that kept a reference to the configuration object.

- I created a new class to be able to retrieve the context for subloggers. I copied over the license header from another file. Im not sure if you should update the year in all the headers?

### DIFF
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -131,7 +131,7 @@ namespace Serilog.Configuration
             configureLogger(lc);
 
             var subLogger = lc.CreateLogger();
-            if (subLogger.HasOverrideMap)
+            if (subLogger.HasAnyOverrides)
             {
                 SelfLog.WriteLine("Minimum level overrides are not supported on sub-loggers " +
                                   "and may be removed completely in a future version.");
@@ -158,7 +158,7 @@ namespace Serilog.Configuration
         {
             if (logger == null) throw new ArgumentNullException(nameof(logger));
 
-            if (logger is Logger concreteLogger && concreteLogger.HasOverrideMap)
+            if (logger is Logger concreteLogger && concreteLogger.HasAnyOverrides)
             {
                 SelfLog.WriteLine("Minimum level overrides are not supported on sub-loggers " +
                                   "and may be removed completely in a future version.");
--- a/src/Serilog/Core/Enrichers/SourceContextEnricher.cs
+++ b/src/Serilog/Core/Enrichers/SourceContextEnricher.cs
@@ -1,0 +1,38 @@
+﻿// Copyright 2013-2015 Serilog Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using Serilog.Events;
+
+namespace Serilog.Core.Enrichers
+{
+    class SourceContextEnricher : ILogEventEnricher
+    {
+        public string Context { get; }
+
+        readonly EventProperty _eventProperty;
+
+        public SourceContextEnricher(string context)
+        {
+            Context = context;
+            _eventProperty = new EventProperty(Constants.SourceContextPropertyName, new ScalarValue(context));
+        }
+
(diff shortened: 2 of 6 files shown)


nblumhardt:
Hi Jan - thanks for the PR!

I think this will take some time to work through; it's a critical part of Serilog and all of the trade-offs here have had a lot of thought put into them - there's a chance this feature can't be introduced without a performance impact on other scenarios, unfortunately.

I think the main performance implications are around lock contention, though it's worth noting that `Logger._overrideMap` is optional because specifying one results in additional memory barriers that are avoided in the simple `_minimumLevel` case, and so requiring an override map will potentially regress this.

There are also memory leak scenarios to explore (child loggers implement `Dispose` only on their implementation types, they're usually not disposed by applications and therefore may not unhook themselves from the override map).

Needs some time, hopefully I can take a closer look soon.

Cheers,
Nick

JanEggers:
Hi Nick,

thx for the feedback.

regarding:
> I think the main performance implications are around lock
> There are also memory leak scenarios to explore (child loggers implement Dispose only on their implementation types
> there's a chance this feature can't be introduced without a performance impact on other scenarios, unfortunately.

I can explore in that direction. If understand that this kind of core functions need to be as pay for play as possible.

that would require: 

creating an interface to be able to swap override map implementations

having a method like EnableOverrideUpdates()
- that creates the override map if not present 
- that swaps the default override map with the new one that supports updates and contains locks 

use weak event pattern instead of dispose to ensure no memory leaks 


please let me know if Im heading the right direction.

JanEggers:
an alternative would be to wrap the logger and have the core loglevel aways verbose. and do the whole IsEnabled check again in the wrapper but I would really like to avoid wrapping all loggers and do the enabled check twice

JanEggers:
another idea I have is removing the seal of Logger and make the IsEnabled Method virtual. 

then create 3 loggers:

1 that uses LogEventLevel directly 
1 that uses LoggingLevelSwitch directly
1 that fetches the loglevel from the override map (that is the only way I see how to get rid of the event)

nblumhardt:
Thanks for the follow-up, sorry about the delay in replying.

There's a lot of detail in this tiny bit of code, since it's a very hot path. Even things like introducing interfaces or virtual methods can have a measurable impact - the benchmarks that are included in the repo are the best place to start digging around - see _Serilog.PerformanceTests_.

We don't have any useful benchmarks that can show the impact of memory barriers/lock contention; the effects of these are non-local and depend on the app, though we could really improve our test suite around this, too. I don't think any added synchronization that wasn't opt-in would be acceptable here, since the no-synchronization constraint has been a big driver of how Serilog's internals work to date.

If there's any way to implement a wrapper system that can be added onto a Serilog logger, that would be interesting and could also be a useful way to try out APIs/concepts in an add-on (gist or nupkg) fashion.

Would love to send more detail but short for time, hope this helps!

JanEggers:
thx for the direction

Issue #1396, opened 2020-01-15
Custom format specifiers get ignored when applied inside the message template


AroglDarthu:
**Does this issue relate to a new *feature* or an existing *bug*?**
Could be either, but I suspect it is a bug.

**What version of Serilog is affected?  Please list the related NuGet package.**
Serilog 2.9.0
Serilog.Sinks.Console 3.1.1

**What is the target framework and operating system?**
netCore 3.1

**Please describe the current behavior?**
I have a class that has it's own format specifiers. These are being ignored when used in the message template.

**Please describe the expected behavior?**
I would expect the same behavior as when using `string.Format`, `ToString(format)` and `string interpolation`.

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**

```
static void Main(string[] args)
{
    Serilog.Debugging.SelfLog.Enable(msg => Debug.WriteLine(msg));
    var logger = new LoggerConfiguration()
        .WriteTo.Console(outputTemplate: "{Timestamp:HH:mm} [{Level:u3}] {Message}{NewLine}")
        .CreateLogger();

    logger.Information("Hello World!");

    var temperature = new Temperature(12);
    logger.Information("MessageTemplate : {Temperature:K}", temperature);
    logger.Information("String.Format   : {Temperature}", string.Format("{0:K}", temperature));
    logger.Information("ToString(format): {Temperature}", temperature.ToString("K"));
    logger.Information("Interpolation   : {Temperature}", $"{temperature:K}");
}
```
Output:
```
// [INF] MessageTemplate : "12,00°C"
// [INF] String.Format   : "285,15K"
// [INF] ToString(format): "285,15K"
// [INF] Interpolation   : "285,15K"
```

Temperature class taken from: https://docs.microsoft.com/en-us/dotnet/standard/base-types/formatting-types#the-iformattable-interface

```
using System;
using System.Globalization;

public class Temperature : IFormattable
{
   private decimal m_Temp;

   public Temperature(decimal temperature)
   {
      this.m_Temp = temperature;
   }
   
   public decimal Celsius
   {
      get { return this.m_Temp; }
   }
   
   public decimal Kelvin
   {
      get { return this.m_Temp + 273.15m; }   
   }
   
   public decimal Fahrenheit
   {
      get { return Math.Round((decimal) this.m_Temp * 9 / 5 + 32, 2); }
   }

   public override string ToString()
   {
      return this.ToString("G", null);
   }
   
[...]

PR #1434, opened 2020-05-18
Use formatting string for custom formattable

SimonTouchtech:
#1396

No.

**Other information:**
This is my first PR to Serilog and I don't have much experience with the more complex parts of the library, so I hope this doesn't break anything.

It looked like the issue appeared in the last line of `PropertyValueConverter.CreatePropertyValue` where `new ScalarValue(value.ToString())` was used as a fallback. By calling `ToString()` and thus converting to a string the custom formatting capabilities are lost. Therefore I added a special case for `IFormattable` just before the `ToString()` call. The reason for putting it last was to not interfere with any destructuring.

It is arguable whether a custom implementation of `IFormattable` could be considered "scalar" as the type can potentially be very complex. The current behavior of `ScalarValue` worked as requested though, so I decided to use it. An alternative approach could be to do something like `FormattableValue` to make the naming a bit more fitting for complex custom `IFormattable` implementations.

**Changes:**
- The `PropertyValueConverter` now returns a `ScalarValue` without modification for `IFormattable` objects

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -172,6 +172,9 @@ namespace Serilog.Capturing
             if (TryConvertCompilerGeneratedType(value, destructuring, valueType, out var compilerGeneratedResult))
                 return compilerGeneratedResult;
 
+            if (value is IFormattable)
+                return new ScalarValue(value);
+
             return new ScalarValue(value.ToString());
         }
 
--- a/test/Serilog.Tests/Core/MessageTemplateTests.cs
+++ b/test/Serilog.Tests/Core/MessageTemplateTests.cs
@@ -114,6 +114,22 @@ namespace Serilog.Tests.Core
             Assert.Equal("Welcome, customer 0012", m);
         }
 
+        class CustomFormattable : IFormattable
+        {
+            public override string ToString() => "No format used";
+
+            public string ToString(string format) => ToString(format, null);
+
+            public string ToString(string format, IFormatProvider provider) => $"Format string {format} used";
+        }
+
+        [Fact]
+        public void FormatStringsArePropagatedForCustomIFormattable()
+        {
+            var m = Render("{Formatted:F}", new CustomFormattable());
+            Assert.Equal("Format string F used", m);
+        }
+
         [Theory]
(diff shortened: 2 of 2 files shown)


nblumhardt:
Thanks! Unfortunately, there's a bit of a subtlety around `ScalarValue`. Whatever it wraps needs to be thread-safe, at least for formatting, and this is hard to guess/guarantee for arbitrary `IFormattable` types, hence the current `ToString()` capturing.

SimonTouchtech:
Thanks for the feedback @nblumhardt! That is indeed unfortunate, as that makes this problem significantly harder than I initially thought.

I might look more into this later in the week, but for now I don't see a clear path to solve it. As you mentioned, thread safety can't be guaranteed for an arbitrary `IFormattable` type, I also don't see a way to create another implementation of `LogEventPropertyValue` (or an extension of `ScalarValue`) that would enforce thread-safety. Therefore I don't see a way to "keep" the formatting capability after the `LogEventPropertyValue` has been created.

My current best idea would be to somehow use the formatting capability before it's "removed" by the `ToString()` capturing. One fairly straight forward way of doing that would be something like

```csharp
if (value is IFormattable formattable)
    return new ScalarValue(formattable.ToString(format));

return new ScalarValue(value.ToString());
```

I see two rather big issues with such a solution though:

1. Formatting/rendering is done in a place where it doesn't belong. It usually happens in the `MessageTemplateRenderer` but now it also happens in the creation of the property value. However, it could be argued that even the existing `ToString()` capturing is a form of rendering, so the only difference would be that the captured value will be more "correct".
2. The `format` is not available in `CreatePropertyValue` so the method signature would have to be changed. As the method is public that would not only be a breaking change, but it would also arguably be a breaking change that doesn't make much sense (see 1).

I guess it would be possible to not change the method signature of `CreatePropertyValue` by abusing the fact that it takes an `object` and we could do something like the following for `IFormattable` implementations.

```csharp
_valueConverter.CreatePropertyValue(new Foo(value, propertyToken.Format), propertyToken.Destructuring));
```

And then have some sort of special case for `Foo`. But IMO that'd be even worse than changing the method signature...

In summary, I don't see any solution that I think would be good. I'll try to have another look later this week, but unless I come up with something better than what I wrote above I'll likely close this PR.

Issue #861, opened 2016-09-14
Default implementation for ILogger


dotnetjunkie:
Serilog's core `ILogger` interface contains 79 members. This makes it extremely cumbersome to create custom implementation, since all 79 members need to be implemented. This design is very similar to log4net that has an `ILog` interface with the same kind of structure.

To work around the problem that the wide interface causes, log4net contains a default implementation of `ILog` called `LogImpl` . `LogImpl` allows overriding individual members and forwards calls to a wrapped logger. Serilog however seems to lack such default implementation, which forces the user to create such implementation itself. This is cumbersome, error prone and far from a good way to spend your time as a developer.

There are a wide range of reasons to create custom implementations, for instance to create a generic implementation of `ILogger` that can be registered in your DI library. Take for instance [this generic log4net `ILog` implementation](https://stackoverflow.com/questions/25112251/log4net-with-di-simple-injector) that can be registered contextually in Simple Injector.

Other reasons to create such wrapper is to decorate instances, or to override individual members of the implementation.

I suggest to create an `LogImpl` equivalent for Serilog. This `LogImpl` could by default forward calls to `Log.Logger`, while allowing a user to override this behavior.

Example:

``` c#
/// <summary>
/// Implementation of logger that forwards the calls to the wrapped <see cref="Logger"/>.
/// </summary>
public class LogImpl : ILogger
{
    /// <summary>Returns the logger to forward to.</summary>
    public virtual ILogger Logger => Log.Logger;

    /// <summary>Create a logger that enriches log events via the provided enrichers.</summary>
    /// <param name="enricher">Enricher that applies in the context.</param>
    /// <returns>A logger that will enrich log events as specified.</returns>
    public virtual ILogger ForContext(ILogEventEnricher enricher)
    {
        return this.Logger.ForContext(enricher);
    }

    // etc.
}
```

This would allow a user to create the following implementation:

``` c#
public sealed class LogImpl<T> : LogImpl
{
    private static readonly ILogger logger = Log.Logger.ForContext<T>();
    public override ILogger Logger { get { return logger; } }
}
```

You might even want to include this `LogImpl<T>` OOTB as well.

nblumhardt:
Thanks for raising the ticket! Definitely in favour of simplifying this use case. (Changed the title to avoid confusing it with a bug.)

[...]

PR #1435, opened 2020-05-19
Default ILogger implementations (netstandard 2.1)

skomis-mm:
 * Added `netstandard 2.1` tfm
 * Added default `ILogger` implemenation, addresses #861 

No.

**Other information:**
 * Default `ILogger` implementation will make user story easier to implement the interface; with ability to further interface evolve. 72 overloads mirrors `Logger` methods: added test that checks the sync. For remaining methods added independent tests.

 * `netstandard2.1` allows adding further improvements utilizing `System.Span<>`, `System.Buffers` etc without introducing new dependencies. 

Additional cleanup, fixes, improvements, such as:
 * Fixed `LoggerShouldNotReferenceToItsConfigurationAfterBeingCreated` test that is failed under Debug;
 * Build / performance scripts actualization;
 * Added / changed tests tfms to LTS targets (should be easier to grab and build for those that have default VS setup);
 * Added .NET Framework reference packages to be able to build under Linux environment (checked in VSCode with WSL)


test/Serilog.Tests/MethodOverloadConventionTests.cs
```diff
@@ -30,6 +35,126 @@ public class MethodOverloadConventionTests
...
+#if FEATURE_DEFAULT_INTERFACE
+        public static IEnumerable<object[]> DefaultInterfaceMethods =>
+            typeof(ILogger).GetMethods()
+                .Where(mi => mi.GetMethodBody() != null)
+                .Where(mi => typeof(Logger).GetInterfaceMap(typeof(ILogger)).InterfaceMethods.Contains(mi))
+                .Select(mi => new object[] { mi });
+
+        [Theory]
+        [MemberData(nameof(DefaultInterfaceMethods))]
+        public void ILoggerDefaultMethodsShouldBeInSyncWithLogger(MethodInfo defaultInterfaceMethod)
```
nblumhardt, line 47: Brilliant! Wouldn't have thought of this 😎

nblumhardt:
Looks awesome 👍 

I'm wondering about the selection of non-default methods:

```csharp
        public bool BindMessageTemplate(string messageTemplate, object[] propertyValues, out MessageTemplate parsedTemplate, out IEnumerable<LogEventProperty> boundProperties) => _inner.BindMessageTemplate(messageTemplate, propertyValues, out parsedTemplate, out boundProperties);
        public bool BindProperty(string propertyName, object value, bool destructureObjects, out LogEventProperty property) => _inner.BindProperty(propertyName, value, destructureObjects, out property);
        public ILogger ForContext(ILogEventEnricher enricher) => _inner.ForContext(enricher);
        public ILogger ForContext(string propertyName, object value, bool destructureObjects = false) => _inner.ForContext(propertyName, value, destructureObjects);
        public bool IsEnabled(LogEventLevel level) => _inner.IsEnabled(level);
        public void Write(LogEvent logEvent) => _inner.Write(logEvent);
        public void Write(LogEventLevel level, Exception exception, string messageTemplate, params object[] propertyValues) => _inner.Write(level, exception, messageTemplate, propertyValues);
```

I think these are the methods that any wrapper or non-trivial `ILogger` should probably implement explicitly :+1:

For simple cases it seems like we could get away with `Write(LogEvent)` being the only requirement, and:

 * `BindMessageTemplate()`, `BindProperty()` - use a default implementation analogous to `new LoggerConfiguration().CreateLogger()` and use it to implement these
 * `ForContext(...)` - return a new `ILogger` implementation that wraps `this` and applies enrichment
 * `IsEnabled()` - return `true`
 * `Write(LogEventLevel, Exception, ...)` - implement in terms of the default `BindMessageTemplate()` etc.

On the one hand, this would make it slightly harder to write a useful delegating logger/wrapper (the user would need to discover the requirement to forward methods other than `Write(LogEvent)`, but, it would increase the chances of a user successfully implementing an `ILogger` that has no Serilog `Logger` on the other side...

I think the choices in the PR are the safer ones, just keen to know what you think about the trade-off.

skomis-mm:
@nblumhardt  totally makes sense 👍 
Initially I provided methods with defaults that are not dependent on the state of the `Logger`s methods.

I just realized that was the wrong reason, since `ILogger` itself is pure abstraction, without implementation rules applied to it, so any reasonable implementation should be correct by default. 

I will adjust the PR with your suggestions and will add custom tests for the remaining methods

Build.ps1
```diff
@@ -21,12 +21,12 @@ $buildSuffix = @{ $true = "$($suffix)-$($commitHash)"; $false = "$($branch)-$($c
 echo "build: Package version suffix is $suffix"
 echo "build: Build version suffix is $buildSuffix" 
 
-foreach ($src in ls src/*) {
+foreach ($src in dir src/*) {
```
adamchester, line 24: Since we're changing this anyway, should we just avoid using the aliases? They give me a lot of headaches.

skomis-mm: Makes sense

src/Serilog/ILogger.cs
```diff
@@ -179,15 +323,43 @@ public interface ILogger
...
+        {
+            if (!IsEnabled(level)) return;
+            if (messageTemplate == null) return;
+
+            // Catch a common pitfall when a single non-object array is cast to object[]
+            if (propertyValues != null &&
+                propertyValues.GetType() != typeof(object[]))
+                propertyValues = new object[] { propertyValues };
+
+            if (DefaultLoggerImpl.BindMessageTemplate(messageTemplate, propertyValues, out var parsedTemplate, out var boundProperties))
```
nblumhardt, line 340: Should this be `this.BindMessageTemplate()`?

nblumhardt:
Looks good! Think it would be interesting to get this out there. Bumped the minor version to 2.10.

I missed the potential to call `this.BindMessageTemplate()` in `Write()` on my first pass - seems like that should probably be the flow, there.

src/Serilog/ILogger.cs
skomis-mm: right, fixed

Issue #1398, opened 2020-02-06
Request: User defined time provider


ChristophAxthelm:
**Does this issue relate to a new *feature* or an existing *bug*?**

- [ ] Bug
- [x] New Feature

**What version of Serilog is affected?  Please list the related NuGet package.**

Serilog 2.9.0

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [x] netCore 2.0
- [x] netCore 1.0
- [x] 4.7
- [x] 4.6.x
- [x] 4.5.x

**Please describe the current behavior?**

Serilog currently uses `DateTimeOffset.Now` to create the timestamp for each new log event. The precision of this time is in the range of 10 milli seconds, which is sufficient for most applications. However for some use cases is required to get a more precise timestamp.

**Please describe the expected behavior?**

For that reason it makes sense to extend Serilog to accept a time, which "somehow" is provided by the user of Serilog.

Options:

1. Pragmatic and easy, but less "beautiful": Add a method for Serilog.ILogger.Write() with a DateTimeOffset parameter: `Write (DateTimeOffset, LogEventLevel, Exception, string, object[])`
2. Allow configuring Serilog to use a user defined time provider. Example:
   ```
   Action<DateTimeOffset> myTimeProvider = ...;
   new LoggerConfiguration()
       .ConfigureTimeProvider(myTimeProvider)
       ....;
   ```

bartelink:
I'd use such a feature to use `DateTimeOffset.UtcNow` so no enricher is required to make a Utc date.

(As a workaround, you can currently add an enricher that adds a property for your custom date)

nblumhardt:
:+1:

The "pragmatic" option is already supported through `Write(LogEvent)`, AFAIK.

You can also implement the second option today, at the expense of a few allocations, using the code from this example:

https://github.com/nblumhardt/serilog-sinks-timewarp/blob/master/Serilog.Sinks.TimeWarp/Program.cs#L42

(Instead of setting every timestamp to a Sunday, the delegate can switch it to UTC, replace it with a more precise one, etc.)

[...]

PR #1440, opened 2020-05-28
Improvements/time provider

ChristophAxthelm:
This PR implements https://github.com/serilog/serilog/issues/1398. Therefor I added a feature for optionally providing a time provider via the fluent api. This allows the user to specify e.g. a high precision time provider of his/hers choice. Example:
```
new LoggerConfiguration()
    .TimeProvider.SetTo(() =>
    {
        return // DateTimeOffset
    })....
```

No

**Other information**
* The test `IfNoTimeProviderIsSpecifiedTheDefaultIsUsed()` checks, that the existing behavior (without specifying an explicit time provider) still works as expected.
* The test `ASpecifiedTimeProviderMustBeUsed` checks, that a specified time provider actually replaces the default time provider.

### DIFF
--- a/src/Serilog/Configuration/LoggerTimeProviderConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerTimeProviderConfiguration.cs
@@ -1,0 +1,31 @@
+﻿using System;
+
+namespace Serilog.Configuration
+{
+    /// <summary>
+    /// Controls sink configuration.
+    /// </summary>
+    public class LoggerTimeProviderConfiguration
+    {
+        readonly LoggerConfiguration _loggerConfiguration;
+        readonly Action<Func<DateTimeOffset>> _setTimeProvider;
+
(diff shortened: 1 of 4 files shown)


nblumhardt:
Thank you for this! :+1:

I think we still need to consider the broader implications; it's a very straightforward approach, but I'm not sure how it will interact with other components in the Serilog ecosystem that use e.g. `ILogger.Write()`.

The advantage of a "middleware"-style approach would be that the timestamps are transformed or replaced as events pass through the pipeline, making it less likely some might skip using the time provider.

Thanks for giving this a nudge; coming back around to look at the next set of changes for Serilog so it's good to have this to consider!

ChristophAxthelm:
@nblumhardt 
Ok. Do you think, it will make it to the 2.10 release? If I can support anyhow (e.g. also for discussions etc.), please don't hesitate to come back to me.

PR #1445, opened 2020-06-10
Updated docs regarding LoggerMinimumLevelConfiguration.Override()

### LINKED ISSUE #1444: Update docs regarding LoggerMinimumLevelConfiguration.Override()
Referring to https://github.com/serilog/serilog/pull/1033...

The XML docs for this method should specify that this is intended behavior (that it is ignored when configuring sub-loggers). A message in a log that is meant for debugging, and is opt-in, is not sufficient to provide discoverability for this behavior.

JakenVeina:
Updated docs for `LoggerMinimumLevelConfigration.Override()` to indicate that this API is not supported for configuration of sub-loggers.

Resolves #1444


nblumhardt:
Thanks! 👍

Issue #1441, opened 2020-05-29
Add overload to LoggerConfiguration.Wrap that allows caller to create the sink to be wrapped


augustoproiete:
Opening this one to start a conversation - happy to send a PR once (and if) we have a design agreed.

I'm building a sink that wraps other sinks, similar to what `Serilog.Sinks.Map` does, using `LoggerConfiguration.Wrap`. One of my requirements is to be able to catch any exceptions thrown by the sink (or sinks) being wrapped.

If my configuration has only one sink, `Wrap` returns an instance of that one sink I configured, which allows me to catch any exceptions when I call `Emit` on the wrapped sink. However, if my configuration has more than one sink, `Wrap` will aggregate them in a `SafeAggregateSink` which means I no longer can catch exceptions given `SafeAggregateSink` will catch them and write messages to `SelfLog`.

[LoggerSinkConfiguration.cs#L253-L255](https://github.com/serilog/serilog/blob/5e2164756efb6023a548b4eb91a71ed7edc3d9e1/src/Serilog/Configuration/LoggerSinkConfiguration.cs#L253-L255):
```
    var enclosed = sinksToWrap.Count == 1 ?
        sinksToWrap.Single() :
        new SafeAggregateSink(sinksToWrap);
```

I'd like to be able to tell `Wrap` which sink to use for aggregating multiple sinks, so that I can use my own implementation of an `AggregateSink` which would give me control over what to do in case of an exception.

With that said, I'd propose an overload that takes a function `Func<List<ILogEventSink>, ILogEventSink>` that replaces the code above and returns the right `ILogEventSink` to be wrapped.

Does that make sense? Is there an alternative that I'm missing?

ps: I did consider just bringing the code of `LoggerSinkConfiguration.Wrap` into my sink and make the changes, but it relies on `private` and `internal` members which my sink can't access.

nblumhardt:
Hi @augustoproiete!

It sounds like what you want is very close to how a `LoggerAuditSinkConfiguration.Wrap()` method would work; I don't think the method exists, yet, but it would make sense for it to be exposed. Would this be a workable option?

augustoproiete:
Hi @nblumhardt!

[...]

PR #1449, opened 2020-06-18
Update `LoggerSinkConfiguration.Wrap()` to use `AggregateSink` ...

augustoproiete:
... to let exceptions bubble up the stack to be consistent with the behavior of wrapping a single sink

Relates to #1441


nblumhardt:
I've merged this into #1466, so when that PR is merged, GitHub should work its magic and mark this one as "merged" :crossed_fingers:

nblumhardt:
Merged! Thanks @augustoproiete 👍

PR #1457, opened 2020-07-06
Add extension methods for lazy evaluated property values

thomashauser:
Add extension methods to use delegates to generate property values only if the corresponding log level is active.


thomashauser:
See issue: https://github.com/serilog/serilog/issues/1458
I don't know how I can link the issue with the PR.

PR #1465, opened 2020-07-21
Fix #1463 - LoggerSinkConfiguration.Wrap() fails to dispose wrapped sinks in some scenarios 

### LINKED ISSUE #1463: LoggerSinkConfiguration.Wrap() fails to dispose wrapped sinks in some scenarios
**Description**
When using async sinks with multiple sinks dispose does not work.
The async sink tries to dispose its childs but the child is a SafeAggregateSink that is not disposable and therefore breaking the cleanup. as SafeAggregateSink is part of the core I opened the bug here.

**Reproduction**

```csharp
var logger = new Configuration().WriteTo.Async(a => a.WriteTo.File("test.log").WriteTo.Console()).CreateLogger();
logger.Dispose();

 logger = new Configuration().WriteTo.Async(a => a.WriteTo.File("test.log")).CreateLogger();  // throws file in use exception
```

**Expected behavior**
the first logger should be disposed correctly

serilog.core

JanEggers:
while creating tests to nail down #1463 I figured out that it was not the fault of the SafeAggregateSink but just the case that the Sink has been wrapped. like for example the Async sink like in the issue I faced.

I repeated the pattern used in LoggerConfiguration.CreateLogger that way even if the wrapper does not implement Dispose all wrapped sinks are disposed.

If this solution is not viable let me know then we can iterate on it.


nblumhardt:
Looks like a good analysis 👍 - thanks for digging in!

I'm wondering whether this will break wrappers that themselves have non-trivial dispose logic/ordering, though, like:

https://github.com/serilog/serilog-sinks-async/blob/dev/src/Serilog.Sinks.Async/Sinks/Async/BackgroundWorkerSink.cs#L59

Even though this is the cleaner fix, perhaps a safer first step is to switch from using `SafeAggregateSink`, here, to a `DisposingSafeAggregateSink`, so that we keep roughly the current semantics?

Following on from that fix, we could (if we want) analyze all of the wrappers that are out there, and their dispose logic, to see whether the more elegant solution might cover it? (We'd also need to benchmark in case the extra indirection in `DisposeWrappingSink.Emit()` has any performance impact, in that case.)

JanEggers:
> Even though this is the cleaner fix, perhaps a safer first step is to switch from using SafeAggregateSink, here, to a DisposingSafeAggregateSink, so that we keep roughly the current semantics?

I could bring back the check and only add the new wra

JanEggers:
oopsie didnot mean to close

nblumhardt:
Thanks! I'll need to loop back to it, seems like it's very close 👍

nblumhardt:
:wave: howdy! Hope you're having a good weekend.

I think the fix here is good; I've got a few ideas around refactoring the various helper sinks (they're taking on a few more roles now), which I'll tinker with down the line, but let's merge and work from here - thanks for sorting this out :+1:

PR #1466, opened 2020-07-26
Refactor wrapper sinks

nblumhardt:
* Breaks the inheritance relationship between `SafeAggregateSink` and `DisposingSafeAggregateSink` - generally we're aiming to avoid inheritance wherever we can
 * Merges #1449 from @augustoproiete - `DisposingSafeAggregateSink` should be `DisposingAggregateSink`, so that wrapper sinks can observe exceptions in more cases
 * Renames `DisposeWrappingSink` to `DisposeDelegatingSink` and removes a level of indirection, since the introduction of the disposing aggregate sink means that the enclosed/wrapped sink will always be `IDisposable` in the case that disposal is required
 * Some tidy-up based on Rider's latest suggestions...


nblumhardt:
While I have some time to work through the queue, I'm going to merge a few fairly straightforward things; 👀 always appreciated after-the-fact, this won't be out on stable until we cut a 2.10.0 release :-)

Issue #1453, opened 2020-06-29
MinimumLevel:Override does not work when MinimumLevel.Default has higher restriction level.


dfrunet:
**Description**
MinimumLevel.Override works only when increasing the restriction level comparing to MinimumLevel.Default and does not work when Override value is less than the Default.

**Reproduction**
Setting `"MinimumLevel": { "Default": "Warning", "Override": {"Foo": "Error", "Foo.Bar": "Information"}}`, output level is `Error` for the `Foo` namespace and `Warning` for all the rest, including `Foo.Bar` (override does not have effect).

**Expected behavior**
Setting `"MinimumLevel": { "Default": "Warning", "Override": {"Foo.Bar": "Information"}}`, output level is `Information` for the `Foo.Bar` namespace only and `Warning` for all the rest.

**Relevant package, tooling and runtime versions**
Serilog (2.8.0)
Serilog.Extensions.Logging (3.0.1)
Serilog.Settings.Configuration (3.1.0)

Finally [realized](#issuecomment-653955010) that the main configuration is not supposed to be fully applied to subloggers, registered like:
```cs
.WriteTo.Logger(lc =>
                    {
                        lc
                          // the next line is the reason to use the sublogger, 
                          // but it's not relevant for the bug, thus commented out
                          //.Filter.ByIncludingOnly(evt => evt.Level < LogEventLevel.Warning)
                           .WriteTo.Console(theme: AnsiConsoleTheme.Code,
                                outputTemplate: "[{Timestamp:HH:mm:ss.fff}][{Level}]: {Message}{NewLine}");
                    });     
``` 
and that's by design.

skomis-mm:
Hi @dfrunet ,

Could not reproduce this..can you make a minimal repro project? Thanks!

dfrunet:
@skomis-mm thank you for the answer, will create the repo in a day or two.

dfrunet:
Now it seems that I have to say sorry and close the item. I started with a minimal repo and was unable to reproduce the issue too, and then when I went back to my work project and started playing with the settings, the problem disappeared there too. I'll play there a bit more, but for now I see no problem at all.

dfrunet:
@skomis-mm Finally I found the way to reproduce the issue. When started to play with different configurations I used json output for simplicity and everything worked fine. But in real project I have to use flat formatting with several format switches. It looks like:

```
LoggerConfiguration loggerConfiguration = new LoggerConfiguration().ReadFrom
[...]

Issue #804, opened 2016-07-08
Multiple sinks not working?


styxke:
Hi,

I'm trying to use multiple sinks for events of a different loglevel, but I can't seem to get it to work.
The configuration:

```csharp
return new LoggerConfiguration()
                  .MinimumLevel.Verbose()
                  .WriteTo.Logger(config => config
                    .MinimumLevel.Verbose()
                    .Filter.ByIncludingOnly(e => e.Level == LogEventLevel.Debug)
                    .WriteTo.File("c:\\...\\log.txt"))
                  .WriteTo.Logger(config => config
                    .MinimumLevel.Verbose()
                    .Filter.ByIncludingOnly(e => e.Level == LogEventLevel.Information)
                    .WriteTo.Seq("http://...:5341/"))
                  .CreateLogger();
```

I get only information level events in Seq and an empty textfile.
Maybe I'm interpreting the docs incorrectly?

nblumhardt:
Thanks for the note.

Just to make sure we're on the same page, I've created a .NET 4.5 console app, installed the `Serilog.Sinks.Console` package, and slightly modified your example to:

```csharp
var log = new LoggerConfiguration()
    .MinimumLevel.Verbose()
    .WriteTo.Logger(config => config
        .MinimumLevel.Verbose()
        .Filter.ByIncludingOnly(e => e.Level == LogEventLevel.Debug)
        .WriteTo.Console())
    .WriteTo.Logger(config => config
        .MinimumLevel.Verbose()
        .Filter.ByIncludingOnly(e => e.Level == LogEventLevel.Information)
        .WriteTo.Console())
    .CreateLogger();

log.Debug("This is debug");
log.Information("This is info");
```

Both events are written to the console. Changing it to:

```csharp
var log = new LoggerConfiguration()
    .MinimumLevel.Debug()
    .WriteTo.Logger(config => config
        .Filter.ByIncludingOnly(e => e.Level == LogEventLevel.Debug)
        .WriteTo.Console())
    .WriteTo.Logger(config => config
        .Filter.ByIncludingOnly(e => e.Level == LogEventLevel.Information)
        .WriteTo.Console())
    .CreateLogger();
```

Also works as I expect. I think the problem is likely to be specific to the file - does the process writing the file have adequate permissions? You can verify this be enabling [SelfLog](https://github.com/serilog/serilog/wiki/Debugging-and-Diagnostics).

In any case, am I correct to assume that the config you really want is this one?

```csharp
[...]

PR #1472, opened 2020-07-31
Don't inherit parent Minimum Level for inline sublogger configuration

### LINKED ISSUE #1346: Sublogger Ignoring Events from Parent Logger
- [X] Bug
- [ ] New Feature

Serilog.AspNetCore 2.1.1

- [X] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

I set a Default MinimumLevel in my Serilog configuration of Debug, and override the MinumumLevel of one of my SourceContexts to "Verbose".
If I log directly to a Sink (File Sink, for example), it works as expected, with everything except the overridden source being filtered to Debug or higher, but Verbose events making it through for the overridden source.
If I instead put a sub-logger between the main logger and the file sink (to add a filter or something), the Verbose events from the overridden source don't make it through.

Here is a snippet from my appsettings.json configuration file:

```
  "Serilog": {
    "MinimumLevel": {
      "Default": "Debug",
      "Override": {
        "System": "Information",
        "Microsoft": "Information",
        "Namespace.BenchmarkTimer": "Verbose"
      }
    },
    "WriteTo": [
      {
        "Name": "Logger",
        "Args": {
          "configureLogger": {
            "Filter": [
              {
                "Name": "ByIncludingOnly",
                "Args": {
                  "expression": "StartsWith(SourceContext, 'Namespace.BenchmarkTimer')"
                }
              }
            ],
            "WriteTo": [
              {
                "Name": "File",
                "Args": {
                  "path": "d:\\path-to-log-folder\\Benchmark Timer.txt"
                }
              }
            ]
          }
        }
      }
    ]
  },
```

All of the events from the parent logger make it through to the sinks configured in the sub-logger.

Configure an application as described above. Write verbose logs from SourceContext "Namespace.BenchmarkTimer". The verbose events will not make it to the sub-logger's configured File Sink.

**Note**: I did try setting restrictedToMinimumLevel to "Verbose" on the sub-logger, which didn't make a difference.

skomis-mm:
Fixes #1346 , #1453 

Behavior change for subloggers configured inline when no explicit default level specified for them.

**Other information**
Inheriting minimum level was done before overrides where introduced, so it doesn't take them into account now for subloggers which results in counterintuitive behavior.

### DIFF
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -1,4 +1,4 @@
-// Copyright 2013-2015 Serilog Contributors
+// Copyright 2013-2020 Serilog Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -30,13 +30,11 @@ namespace Serilog.Configuration
     {
         readonly LoggerConfiguration _loggerConfiguration;
         readonly Action<ILogEventSink> _addSink;
-        readonly Action<LoggerConfiguration> _applyInheritedConfiguration;
 
-        internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action<ILogEventSink> addSink, Action<LoggerConfiguration> applyInheritedConfiguration)
+        internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action<ILogEventSink> addSink)
         {
             _loggerConfiguration = loggerConfiguration ?? throw new ArgumentNullException(nameof(loggerConfiguration));
             _addSink = addSink ?? throw new ArgumentNullException(nameof(addSink));
-            _applyInheritedConfiguration = applyInheritedConfiguration ?? throw new ArgumentNullException(nameof(applyInheritedConfiguration));
         }
 
         /// <summary>
@@ -126,13 +124,7 @@ namespace Serilog.Configuration
         {
             if (configureLogger == null) throw new ArgumentNullException(nameof(configureLogger));
 
-            var lc = new LoggerConfiguration();
-
-            // apply inherited configuration except minimum level
-            // since parent logger can have overrides
-            _applyInheritedConfiguration(lc);
-            lc.MinimumLevel.Is(LevelAlias.Minimum);
-
+            var lc = new LoggerConfiguration().MinimumLevel.Is(LevelAlias.Minimum);
             configureLogger(lc);
 
             var subLogger = lc.CreateLogger();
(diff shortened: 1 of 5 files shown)


src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -128,7 +128,11 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
 
             var lc = new LoggerConfiguration();
 
+            // apply inherited configuration except minimum level
+            // since parent logger can have overrides
             _applyInheritedConfiguration(lc);
+            lc.MinimumLevel.Is(LevelAlias.Minimum);
```
nblumhardt, line 134: Since `_applyInheritedConfiguration` ends up calling:

```csharp
        void ApplyInheritedConfiguration(LoggerConfiguration child)
        {
            if (_levelSwitch != null)
                child.MinimumLevel.ControlledBy(_levelSwitch);
            else
                child.MinimumLevel.Is(_minimumLevel);
        }
```

it seems like removing that call here might be a cleaner fix?

I had a quick attempt at figuring out the original purpose of `_applyInheritedConfiguration` and can't track it down; I suspect it may have been part of an earlier version that took advantage of `Logger` directly implementing `ILogEventSink.Emit()` (without any level check), but can't immediately tell from the source.

Removing it entirely seems not to break any tests. Am I missing anything?

nblumhardt:
Sorry I missed your mention/comments on 1453 - I've kept meaning to loop back around to this :sweat:

Making some time to catch up, now.

src/Serilog/Configuration/LoggerSinkConfiguration.cs
skomis-mm: I've tracked it down to this [pr](https://github.com/serilog/serilog/pull/654). Part of it was the fix for [804](https://github.com/serilog/serilog/issues/804#issuecomment-231665651) issue.

The issue was about case with default minimum level of the root less than `Information` which results in explicit setting minimum for subloggers by users. 

The open question - why it was choosen the same as for the parent instead of `Verbose`..

Hint about `Logger` as `ILogEventSink` led me to such examples:

Behavior before this pr:
```csharp
new LoggerConfiguration()
    .MinimumLevel.Verbose()
    .WriteTo.Sink(
        new LoggerConfiguration()
            .MinimumLevel.Fatal()
            .WriteTo.Console()
            .CreateLogger())
    .CreateLogger()
    .Verbose("oops"); // surpasse the Fatal filter
```
but
```csharp
new LoggerConfiguration()
    .MinimumLevel.Verbose()
    .WriteTo.Sink(
        new LoggerConfiguration()
            .MinimumLevel.Fatal()
            .WriteTo.Logger(lc => lc.WriteTo.Console())
            .CreateLogger())
    .CreateLogger()
    .Verbose("oops"); // suppressed by the sublogger's Fatal filter
```

After this pr both examples will result in surpassing the Fatal filter. So it actually make the behavior consistent :)

So it makes sense to remove those lines as you pointed.

skomis-mm:
Removed `ApplyInheritedConfiguration` since it was used just in one place

nblumhardt:
LGTM :+1:

PR #1475, opened 2020-08-06
Add LoggerConfiguration.CreateLogger(bool) overload

sungam3r:
Hi. I know that by design Serilog catches all exceptions from sinks. Nevertheless, we have cases when it is necessary to receive exception. I have added `LoggerConfiguration.CreateLogger(bool)` overload. No breaking changes. Just one bool switch to decide whether to wrap sinks into `SafeAggregateSink`. @nblumhardt What do you think?


tsimbalar:
Is `Audit` what you want then? for stronger guarantees about log deliveries, and exceptions bubbling up ?

see https://github.com/serilog/serilog/pull/826

sungam3r:
🤦‍♂️  Exactly! Thanks.

Issue #1512, opened 2020-12-10
'Destructure.ByTransforming<T>' should have an option to consider hierarchies


julealgon:
**Is your feature request related to a problem? Please describe.**
I'm using an `Encoding` property in one of my models, and I'm serializing that into the logger. I noticed that an exception was being thrown because of the specific `Encoding` instance during serialization.

I then proceeded to create a destructuring policy for that type to avoid the problem:

```csharp
.Destructure.ByTransforming<Encoding>(e => e.ToString())
``` 
This did not work: I was still getting the exact same exception as before.

After thinking for a bit, I realized my instance was not exactly an `Encoding` instance, but a derived, internal type returned by `Encoding.UTF8`.

The `ByTransforming` call is never hit because the type check done internally uses the instance's exact type:

https://github.com/serilog/serilog/blob/e82213e98ece04bdfcb13d066f91b987ca4921c6/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs#L118

**Describe the solution you'd like**
I'd like at least an option to also take any derived type into consideration when using `ByTransforming`, even if that requires a specific overload. I believe the current API is misleading.

I propose the following (which is something FluentAssertions does for exceptions):
- `ByTransforming<T>` : match any type that "is a T"
- `ByTransformingExactly<T>` : match only exactly type T

I believe this would result in a more intuitive API.

**Describe alternatives you've considered**
I had to create a custom `IDestructuringPolicy` and perform the logic inside.

I believe I could also use `ByTransformingWhere` but I felt the call would be quite misleading to readers:
`ByTransformingWhere<Encoding>(t => t.IsAssignableTo(typeof(Encoding)), e => e.ToString())`

**Additional context**
FluentAssertions examples:

- `Throw<TException>`:
https://github.com/fluentassertions/fluentassertions/blob/a0262e39961decf9fdf1399a5d775050fed93640/Src/FluentAssertions/Specialized/DelegateAssertions.cs#L28-L39

- `ThrowExactly<TException>`:
https://github.com/fluentassertions/fluentassertions/blob/a0262e39961decf9fdf1399a5d775050fed93640/Src/FluentAssertions/Specialized/DelegateAssertions.cs#L96-L114

julealgon:
I just realized the exact same problem (and suggestion) applies to `AsScalar(Type)` and `AsScalar<T>`.

If I use `AsScalar<Encoding>()`, I still get an exception.

[...]

PR #1513, opened 2020-12-10
Fix miswording on 'AsScalar' destructuring methods

julealgon:
This fixes a small mistake in the `AsScalar` set of destructuring methods' comments.
- "...event when..." to "...even when..."


julealgon:
Found this while making a few tests related to #1512

nblumhardt:
Thanks! 👍

PR #1531, opened 2021-02-02
ProjectedDestructuringPolicy made public

StanEgo:
I think it would be better to have a public modifier for ProjectedDestructuringPolicy. Let me explain a reason.

Imagine that we're designing a complex system with a lot of layers/packages those are conditionally/dynamically bind. Some of them may have their own classes that can be destructured so simple calls like config.ByTransforming(...) is not a good option and violates DIP (high level modules should not depend on low level modules). Sometimes we even don't know which transformation will be provided by a third-party plugin. Good design is to have something like this registered in composition root (talking in terms of DIP principle, IoC design patterns, etc).

```csharp
public class SerilogConfigurator
{
    public SerilogConfigurator(
        ...
        // Serilog enrichers and destructures are collected and injected by IoC. 
        // We don't care here where they come from
        IEnumerable<ILogEventEnricher> enrichers,
        IEnumerable<IDestructuringPolicy> destructures
    )
    {
        ...
        // Very simple dispatching of all collected policies to Serilog LoggerConfiguration
        config.Enrich.With(enrichers.ToArray());
        config.Destructure.With(destructures.ToArray());
        ...
    }
}
```

So when we want to create some projection in some specific package it could be as simple as (with extension methods it can be even shorter):
```csharp
// Autofac syntax used in this example
var projection = new ProjectedDestructuringPolicy(...instantiation options...)
builder.RegisterInstance(projection).As<IDestructuringPolicy>();

builder.RegisterType<PostgresConnectionEnricher>().As<ILogEventEnricher>().SingleInstance();
```

But this is impossible because ProjectedDestructuringPolicy is private so to have a good SOLID design this ~~wheel should be reinvented~~ class should be created again. I think you see a good implementation of OCP/DIP principles here. High level modules do not rely on low level. Configuration can be extended without any changes in the module responsible for configuration, etc.

I'm not sure how big is my investment into the codebase, but may be you'd find some ideas useful for documentation.

I made this change only for ProjectedDestructuringPolicy just because it looks like to the the only policy explicitly called from LoggerDestructuringConfiguration.

### DIFF
--- a/src/Serilog/Policies/ProjectedDestructuringPolicy.cs
+++ b/src/Serilog/Policies/ProjectedDestructuringPolicy.cs
@@ -1,4 +1,5 @@
-﻿// Copyright 2013-2015 Serilog Contributors
+
+// Copyright 2013-2015 Serilog Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -18,17 +19,33 @@ using Serilog.Events;
 
 namespace Serilog.Policies
 {
-    class ProjectedDestructuringPolicy : IDestructuringPolicy
+    /// <summary>
+    /// When destructuring objects, projects instances of the specified type with
+    /// the provided function.
+    /// </summary>
+    public class ProjectedDestructuringPolicy : IDestructuringPolicy
     {
         readonly Func<Type, bool> _canApply;
         readonly Func<object, object> _projection;
 
+        /// <summary>
+        /// Construct policy
+        /// </summary>
+        /// <param name="canApply">
+        /// Predicate for object type to check if policy can be applied.
+        /// </param>
+        /// <param name="projection">
+        /// Projection function transforming incoming type into the projected one.
+        /// </param>
         public ProjectedDestructuringPolicy(Func<Type, bool> canApply, Func<object, object> projection)
         {
             _canApply = canApply ?? throw new ArgumentNullException(nameof(canApply));
             _projection = projection ?? throw new ArgumentNullException(nameof(projection));
         }
 
+        /// <summary>
+        /// Implements <see cref="IDestructuringPolicy.TryDestructure(object, ILogEventPropertyValueFactory, out LogEventPropertyValue)"/>.
+        /// </summary>
         public bool TryDestructure(object value, ILogEventPropertyValueFactory propertyValueFactory, out LogEventPropertyValue result)
         {
             if (value == null) throw new ArgumentNullException(nameof(value));


nblumhardt:
:-) thanks for the follow-up. I did spend quite a bit of time thinking about this, apologies I only had time to write a brief reply.

Public API surface really costs a lot in these kinds of projects; gotta trust me that I love object-oriented programming, reuse, and dependency injection, and I use them all extensively in application-level projects, but for something long-lived, at the bottom of the dependency graph, and used as widely as Serilog, keeping the surface area small is a major goal.

For this project it's been the key to surviving use on a massive scale - even though it might sometimes mean adding a little sprinkling of repetitive code here and there, it's kept Serilog's core relatively legacy-free, and free of breaking changes across a large number of releases.

Regarding `ProjectedDestructuringPolicy`, if we were to make it public we'd need to think through its API more closely. For example, why keep the test/apply style construction? What happens if the first delegate returns `true` for a `Type`, but then the second delegate is called with an object of a different type? Would a better API be `ProjectedDestructuringPolicy<T>` with a constructor accepting `Funct<T, object>`? Etc.

Generally it's only possible to dig into an API at that level of detail when there are a lot of requests for it, or if it's part of a bigger-picture plan. Definitely not against considering this in the future - at present it doesn't seem to meet that bar, but open to more feedback if others come to the same conclusion.

Thanks again!

StanEgo:
@nblumhardt , you're welcome :) I know that such suggestions to change a lot of things for some ephemeral purpose may be annoying. But widely used projects are also responsible for education. At least I think so. And if there are not so many requests for such changes just tells us that there is a big room for users improvement.

If you like to build applications with all these good design stuff you mentioned just imagine the situation like I described. You have a big, scalable platform with a lot of business "plugins" which can provide there own destructuring policies, enrichers and so on. And most of such plugins are developed by third-parties. How would you go with Serilog?

And sorry, I'm not sure that understand explanation "good design is not good for a widely used projects". I contribute to many of them. E.g. NUnit accordingly to GitHub stats has 6-7x more usages than Serilog. And constrains in it also has all this nifty shortcuts based on static methods like Is.Null, Has.Property, etc. But all this based on a system of more than 70 constrains those can have a legion of different combinations and almost 100% of are public and I can reuse them to construct my own. The same for attributes. I easily inherited [Test] attribute of NUnit in my [Specification] or [Feature] ones to have an extended behavior.

I'm just trying to understand the principle of making a choice for modifier. Massive scale is not a good one, sorry. In massive scale bad reusing is a drawback. For example why PropertyEnricher is public and ConditionalEnricher is not? Is it to dangerous to inherit the last one and provide custom predicate? So it will be better to create a custom, potentially fragile version? How then NUint survive having similar PropertyConstraint and PredicateConstraint public ;)

Having so many question regarding ProjectedDestructuringPolicy implementation just means that you know about its drawbacks and simply hide this bomb by removing "public" modifier ;) And of course thinking towards having generic implementation is valuable. Better type control, boxing, etc. As a hassle-free, temporary solution ProjectedDestructuringPolicy can be left as is and public ProjectedDestructuringPolicy<T> inherits it, provides default CanApply son only one Func<T, object> delegate will be used for construction. But probably better to change it completely. Anyway nobody else used it before :)

Hope I wasn't too rude :) If you want to make design better, but just have a lack of resources and this is the reason you closed the PR, I can try to convert it into a more sophisticated solution. The current atomic change was used as a probe to understand values/goals/communications/etc.

nblumhardt:
Thanks for your reply. At the moment, no,  thanks - it's great to have another perspective, but Serilog's a different project from NUnit, with different priorities. Will definitely keep this one in mind, though.

PR #1533, opened 2021-02-08
more precise logevent timestamp

Avisss:
Processing of message template can take a bit of time and it offsets the real timestamp of log event.
I catch max additional offset about 12 milliseconds with small testing


augustoproiete (approved):
LGTM

nblumhardt:
👍

PR #1545, opened 2021-03-09
Made MinimumLevel virtual to allow overriding

mc0re:
Made LoggingLevelSwitch.MinimumLevel virtual to allow overriding.


nblumhardt:
Thanks for the PR; please see: https://github.com/serilog/serilog/issues/1468

mc0re:
#1468 is the other way around. I want to change the `MinimumLevel` based on the current HTTP context (User), not to listen to it.

PR #1554, opened 2021-03-16
Add CodeQL

### LINKED ISSUE #1553: Add continuous security analysis using GitHub CodeQL
- [CodeQL](https://github.com/github/codeql) is an open source code scanning tool for discovering vulnerabilities across a codebase

Built with ❤ by [Pipeline Foundation](https://pipeline.foundation).

aleks-ivanov:
- add `codeql-analysis.yml` which automatically enables CodeQL code security and quality scanner. It executes on every push commit, PR, manually and every day at 8:00AM UTC. A scan check can be viewed on commits and PRs. All alerts and fix suggestion can be viewed at **Security** tab -> **Code scanning alerts** -> **CodeQL**

Resolves #1553

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,36 @@
+name: CodeQL Analysis
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 8 * * *'
+
+jobs:
+  analyze:
+    name: CodeQL Analysis
+    runs-on: ubuntu-latest
+    env:
+      SOLUTION_PATH: Serilog.sln
+    steps:
+      - name: Checkout repository
+        id: checkout_repo
+        uses: actions/checkout@v2
+
+      - name: Initialize CodeQL
+        id: init_codeql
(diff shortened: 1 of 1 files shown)


nblumhardt:
Hi!

Have you run the tool to determine what the current state of analysis is? I'm not keen to merge if it immediately creates downstream work suppressing false positives.

Thanks!

aleks-ivanov:
Yes to this as well 🙂 

Unfortunately, I won't be able to show you the result of the scan like with Dependabot, since it is under the Security panel, but I guess a screenshot won't be too useless in this case 🙂 

Screenshot_1

Although for sure there would be some false positives, I believe long-term both Dependabot and CodeQL will be of substantial value.

nblumhardt:
Thanks! I just checked two of these (the highlighted one, and the LoggerTests:47 one a few up from it), both are straightforward code that does not behave in the manner the analysis is suggesting. The codebase has had a lot of eyes on it over a long period, I would bet that there are still some edge cases lurking, but 150 results seems like it will be a long way off.

I would be inclined to pass on CodeQL right now, we might check it out again in the future though!

Issue #1560, opened 2021-04-20
Add support for collections/`string[]` in key-value settings


stochmal:
using Serilog.Settings.AppSettings and Serilog.Sinks.Datadog.Logs

entry in App.config causes exception:

		<add key="serilog:write-to:DatadogLogs.tags" value="name:ZA2-FUNDSTI1;client:IRESS;owner:Trading;key:trading-production-za;application:FundsTradingInterface" />

I can make pull request to fix it but can't push for some reason

stochmal:
added this code to fix this exception
![image-2021-04-20-14-05-00-088](https://user-images.githubusercontent.com/12511085/115393828-6854c980-a1e2-11eb-9f34-ce4f1fe33da7.png)

stochmal:
#1561

PR #1561, opened 2021-04-20
Add support for collections/`string[]` to key-value settings

stochmal:
https://github.com/serilog/serilog/issues/1560


nblumhardt:
Thanks Tomasz!

Initial thoughts -

 * We'd need to add some tests for this
 * Should we consider `StringSplitOptions.RemoveEmptyEntries` or similar?
 * Is `;` the best separator, vs. e.g. `,` or `|`, etc.?
   - ';' can be hard to grapple with in shells (e.g. where it might be a command separator)
   - using `,` might lead us towards supporting CSV as the value format, down the track, which has reasonably well-known escaping mechanisms
   - is there any precedent in related projects we should consider?
 * The codebase doesn't use explicit `private` or `internal`
 * Should we be supporting any `T[]` type here by recursively converting the values?
 * Or, should this be one of the `ExtendedTypeConversions`?

stochmal:
I need this fix for company project I am working on. Lets see what I can change.

nblumhardt:
Hi @stochmal - still interested in pushing this forward?

stochmal:
yes need this fix as it's broken in .NET Framework (works fine in .NET Core)

tomasz-stochmal:
> Any thoughts on:
> 
> > * Should we be supporting any `T[]` type here by recursively converting the values?
> 
> Would seem to make this a more well-rounded addition.

tags are set in config file so it's always string to string[] conversion so don't see any use case for other array type

stochmal:
> > Any thoughts on:
> > > * Should we be supporting any `T[]` type here by recursively converting the values?
> > 
> > 
> > Would seem to make this a more well-rounded addition.
> 
> tags are set in config file so it's always string to string[] conversion so don't see any use case for other array type

this entry in App.config is crashing Serilog

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <appSettings>
    <add key="serilog:write-to:DatadogLogs.tags" value="name:ZA2-FUNDSTI1,client:IRESS,owner:Trading,key:trading-production-za,application:FundsTradingInterface" />
  </appSettings>
</configuration>
```

nblumhardt:
> tags are set in config file so it's always string to string[] conversion so don't see any use case for other array type

Configuration method args can be any type; if a sink takes `int batchSize` then `value="2"` will be converted today. If a method takes `int[]` then a value like `"2,3,4"` should probably work, if we're aiming for consistency.

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
```diff
@@ -114,12 +114,30 @@ public static object ConvertToType(string value, Type toType)
...
+                {
+                    if(int.TryParse(item, out int result))
+                    {
+                        list.Add(result);
+                    }
+
+                }
+                return list.ToArray();
+            }
+
```
nblumhardt, line 140: Thanks for the update!

This approach might get a bit onerous once we have to support `bool[]`, `long[]`, etc.

Rather than these two blocks, we can support any collection type with:

```csharp
if (toType.IsArray && toType.GetArrayRank() == 1)
{
    var elementType = toType.GetElementType();
    var items = value?.Split(',');
    var length = items?.Length ?? 0;
    var result = Array.CreateInstance(elementType, length);
    if (items != null)
    {
        for (var i = 0; i < length; ++i)
        {
            var item = items[i];
            var element = ConvertToType(item, elementType);
            result.SetValue(element, i);
        }
    }
    return result;
}
```

(Should keep the existing and added tests passing 🤞 )

stochmal: Implemented your code and fixed failing tests on empty arrays. I think it's good to be merged into dev?

nblumhardt:
Awesome, thanks @stochmal ! Looks good to me, too - should have this in today 👍

stochmal:
Nice, Thanks for your help. When will there be official release that contains that fix?

nblumhardt:
I've just opened https://github.com/serilog/serilog/pull/1614 to track the next stable release; I'll add more details there as we figure out the schedule 👍

PR #1584, opened 2021-06-14
Fix scenario where LogContext throws a SecurityException when making cross-AppDomain calls inside an AppDomain with limited permissions

### LINKED ISSUE #1486: LogContext throws SecurityException when making cross AppDomain calls in AppDomain with limited permissions
**Description**
On .NET Framework 4.5.x, the `LogContext` is stored in the `CallContext` because there is not an AsyncLocal primitive. As a result of storing the object in the `CallContext` and invoking the .NET Remoting infrastructure, issue #987 began occurring with a fix being issued in #992 that was published in Serilog 2.6.0. That original problem was resolved, but it created a new problem: the solution to keeping Serilog objects in the `CallContext` alive so they are not disconnected by the "server" introduced new logic into the object that was transmitted to the server so that it would keep itself alive. However, in the case of a cross AppDomain call where the new AppDomain does not have permission to use the remoting infrastructure (`SecurityPermissionFlag.RemotingConfiguration`), this lifetime management fails when the permission is demanded, with the following error message:

```
Unhandled Exception: System.Security.SecurityException: Request for the permission of type 'System.Security.Permissions.SecurityPermission, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' failed.
```

**Reproduction**
Please provide code samples showing how you're configuring and calling Serilog to produce the behavior.
```csharp
        public static void Main(string[] args)
        {
            // Set up an AppDomain with only code execution permissions
            // Notably, it lacks the SecurityPermissionFlag.RemotingConfiguration flag
            var permSet = new PermissionSet(PermissionState.None);
            permSet.AddPermission(new SecurityPermission(SecurityPermissionFlag.Execution));
            var remote = AppDomain.CreateDomain("Remote", null, AppDomain.CurrentDomain.SetupInformation, permSet);

            try
            {
                using (LogContext.PushProperty("Prop1", 42))
                {
                    remote.DoCallBack(HelloWorld);
                }
            }
            finally
            {
                AppDomain.Unload(remote);
            }
        }

        public static void HelloWorld()
        {
            Console.WriteLine("Hello World");
        }
```

**Expected behavior**
The program should execute without any issues, and the cross-AppDomain call should write "Hello World" to the console.

- Serilog versions 2.6.0-2.10.0
- .NET Framework 4.5.x (which use the `CallContext` to store ambient `LogContext` information)

**Additional context**
This problem was encountered in the DataDog/dd-trace-dotnet repo in https://github.com/DataDog/dd-trace-dotnet/pull/925, so I will follow up with a proposed solution and see if it is appropriate for this library. The solution we implemented there was to modify the `DisposableObjectHandle` that is stored in the `CallContext` so that its override of `InitializeLifetimeService` returns `null` instead of invoking any remoting infrastructure. This avoids the remoting permission issue at the cost of keeping the object alive indefinitely. However, this is easily handled by disconnecting the object ourselves when `Dispose()` is called.

zacharycmontoya:
Fixes #1486. See issue for full reproduction and analysis details.

Fix: In the DisposableObjectHandle class, no longer have the proxy object keep itself alive on the server by invoking base.InitializeLifetimeService(), which can cause .NET code access permission errors. Instead, return null to keep it alive indefinitely and forcefully remove the remote object when the local object is disposed.

This PR also adds the restricted AppDomain settings to the existing test scenarios in Serilog.Tests.Context.LogContextTests

### DIFF
--- a/src/Serilog/Context/LogContext.cs
+++ b/src/Serilog/Context/LogContext.cs
@@ -244,26 +244,50 @@ namespace Serilog.Context
 
         sealed class DisposableObjectHandle : ObjectHandle, IDisposable
         {
-            static readonly ISponsor LifeTimeSponsor = new ClientSponsor();
+            private bool _disposed;
 
             public DisposableObjectHandle(object o)
                 : base(o)
             {
             }
 
-            public override object InitializeLifetimeService()
+            public override sealed object InitializeLifetimeService()
             {
-                var lease = base.InitializeLifetimeService() as ILease;
-                lease?.Register(LifeTimeSponsor);
-                return lease;
+                // For net45 where AsyncLocal is not a built-in Framework class, we must propgate the ambient context using LogicalCallContext.
+                // One consequence of this is that we invoke the System.Runtime.Remoting infrastructure, and if a remote method call or cross-AppDomain
+                // method call is made, objects stored inside the LogicalCallContext are transmitted to the other side (aka the "server").
+                //
+                // For distributed garbage collection, the "server" uses leases / sponsors to keep track of what objects are needed by "clients"
+                // (our original AppDomain) and should not be garbage collected. If we do not refresh the lease on objects stored inside the LogicalCallContext,
+                // the "server" will garbage collect them and disconnect them, meaning any accesses of the object inside the LogicalCallContext will throw a
+                // RemotingException.
+                //
+                // This solves the problem by returning null so that the lease on the object never expires. To make sure the object doesn't live forever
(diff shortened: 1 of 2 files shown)


nblumhardt:
Thanks Zach! This looks good to me, but, it's been a long time since I tangled with the CLR's remoting infrastructure :-)

@serilog/reviewers-core anyone able to give this a second look over?

khellang:
Yeah, I'm in the same boat as @nblumhardt. Can't really comment on the remoting code 😅

PR #1585, opened 2021-06-15
make TestDummies use nullable

SimonCropp:


test/TestDummies/Console/DummyConsoleSink.cs
```diff
@@ -7,13 +7,13 @@ namespace TestDummies.Console
...
     {
-        public DummyConsoleSink(ConsoleTheme theme = null)
+        public DummyConsoleSink(ConsoleTheme? theme = null)
         {
             Theme = theme ?? ConsoleTheme.None;
         }
 
         [ThreadStatic]
-        public static ConsoleTheme Theme;
+        public static ConsoleTheme Theme = null!;
```
nblumhardt, line 16: I think a `null` value can be observed, here, if no sink has been constructed on the thread.

SimonCropp: fixed

nblumhardt:
Nice 👍

PR #1588, opened 2021-06-15
add nullable for ilogger and implementations

SimonCropp:


nblumhardt:
👋 

Although the majority of methods on `ILogger` won't throw when `null` is supplied, most of them also won't act in any useful way.

They accept `null` so as to not break an application at runtime due to programmer error, but with null checking enabled, a warning or error seems appropriate when called incorrectly.

This _could_ mean build breakages down the line - but, I think since it's early days for `nullable`, and consumers are opting-in, that would be welcome (would a minor version communicate this adequately, or should we bump major?).

What do you think?

SimonCropp:
> They accept null so as to not break an application at runtime due to programmer error, but with null checking enabled, a warning or error seems appropriate when called incorrectly.

Does that apply to all the things? 

My assumption would the following should not be null: `logEvent`, `source`, `enrichers`, `messageTemplate`

But that the following may be null: `propertyValues`, `Exception`, `out LogEventProperty`,  `out MessageTemplate? parsedTemplate`


> not break an application at runtime due to programmer error

i would prefer that it did break at runtime if invalid values are passed in. eg if i incorrectly pass in a null message template, i want to know, not have the log entry ignored. but i understand your reasoning

> This could mean build breakages down the line - but, I think since it's early days for nullable, and consumers are opting-in, that would be welcome (would a minor version communicate this adequately, or should we bump major?).

I would lean towards a minor. they can always suppress to ignore the build error/warning until they get time to fix it

nblumhardt:
> My assumption would the following should not be null: logEvent, source, enrichers, messageTemplate
> But that the following may be null: propertyValues, Exception, out LogEventProperty, out MessageTemplate? parsedTemplate

I think that's correct 👍 

> i would prefer that it did break at runtime if invalid values are passed in. eg if i incorrectly pass in a null message template, i want to know, not have the log entry ignored. but i understand your reasoning

That was my original inclination, too - the no-throw guarantee was inherited from log4net. I've come around to it over the years 😄 ... avoiding exceptions when trying to log a failure is the winning argument, for me: throwing while failing often masks the original root cause, the code paths involved are rarely properly exercised in tests.

> I would lean towards a minor. they can always suppress to ignore the build error/warning until they get time to fix it

Sounds good to me 👍

SimonCropp:
how does it look now?

SimonCropp:
gentle bump. so i can move onto the next part

nblumhardt:
Sorry, will be back to this shortly! Need to think through it carefully :-)

SimonCropp:
no worries. let me know if u want to jump on a call

src/Serilog/Core/Pipeline/SilentLogger.cs
```diff
@@ -26,11 +28,11 @@ class SilentLogger : ILogger
         {
         }
 
-        public ILogger ForContext(ILogEventEnricher enricher) => this;
+        public ILogger ForContext(ILogEventEnricher? enricher) => this;
```
nblumhardt, line 31: I don't think `enricher` should be nullable, here (implementations receiving null just bail out instead of failing hard)

nblumhardt (commented):
I think this is a good starting point; I'm not _100%_ sure that the `params object[]` args should themselves be nullable, but those methods do work when passed `null`, so I'm leaning towards the same conclusion that you drew, here 👍 

Did spot one `ForContext()` arg that should probably be non-nullable - what do you think?

src/Serilog/Core/Pipeline/SilentLogger.cs
SimonCropp: done

nblumhardt:
Awesome, thanks 👍

PR #1591, opened 2021-06-21
make tests nullable

SimonCropp:


test/Serilog.Tests/Capturing/PropertyValueConverterTests.cs
```diff
@@ -113,13 +113,13 @@ public void ByDefaultADestructuredNonNullNullableIsItsValue()
 
         class A
         {
-            public B B { get; set; }
+            public B B { get; set; } = null!;
```
nblumhardt, line 116: Any reason not to just make this (and the similar ones below) into nullable properties? I.e.:

```csharp
public B? B { get; set; }
```

This would match the current semantics of the code.

Using `null!` catches my eye and causes me to dig around more, trying to figure out how the assertion "this property begins life as null but that value can never be observed" could be upheld by the code.

test/Serilog.Tests/Context/LogContextTests.cs
```diff
@@ -38,7 +38,7 @@ public LogContextTests()
         [Fact]
         public void PushedPropertiesAreAvailableToLoggers()
         {
-            LogEvent lastEvent = null;
+            LogEvent lastEvent = null!;
```
nblumhardt, line 41: Since the null value is observable, would:

```csharp
LogEvent? lastEvent = null;
```

be more appropriate here, and in the similar cases below?

test/Serilog.Tests/Core/LoggerTests.cs
```diff
@@ -200,7 +200,7 @@ public void DelegatingLoggerShouldDelegateCallsToInnerLogger()
             Assert.Equal(42, collectingSink.SingleEvent.Properties["number"].LiteralValue());
             Assert.Equal(
                 expected: new SequenceValue(new[] { new ScalarValue(1), new ScalarValue(2), new ScalarValue(3) }),
-                actual: (collectingSink.SingleEvent.Properties["values"] as SequenceValue),
+                actual: (SequenceValue)collectingSink.SingleEvent.Properties["values"],
```
nblumhardt, line 203: :+1: 👍

nblumhardt (commented):
Looks good; would prefer to avoid `null!` wherever we can - what do you think?

test/Serilog.Tests/Capturing/PropertyValueConverterTests.cs
SimonCropp: good point. done

test/Serilog.Tests/Context/LogContextTests.cs
SimonCropp: since u dont assert.notnull i assumed `null!;`. should i add a assert.notnull?

nblumhardt: `Assert.NotNull(logEvent)` would be a pleasant addition :+1: .. I think there's an implicit `Assert.NotNull()` in many of these cases - dotting through the reference to assert on `Properties` will fail if the event's null (not very elegantly, of course :-))

SimonCropp: done

test/Serilog.Tests/Core/ChildLoggerTests.cs
```diff
@@ -195,17 +195,17 @@ static object[] T(string rs, int? rl, string cs, int? cl, bool r, LogEventLevel
...
+            string? childOverrideSource,
             int childOverrideLevelIncrement,
             bool eventShouldGetToChild)
         {
             var incomingEventLevel = Information;
             var rootOverrideLevel = incomingEventLevel + rootOverrideLevelIncrement;
             var childOverrideLevel = incomingEventLevel + childOverrideLevelIncrement;
 
-            LogEvent evt = null;
+            LogEvent evt = null!;
```
nblumhardt, line 208: Another here, and on line 258?

test/Serilog.Tests/LoggerConfigurationTests.cs
```diff
@@ -465,7 +466,7 @@ public void MaximumCollectionCountNotEffectiveForDictionaryWithAsManyKeysAsLimit
 
         static string LogAndGetAsString(object x, Func<LoggerConfiguration, LoggerConfiguration> conf, string destructuringSymbol = "")
         {
-            LogEvent evt = null;
+            LogEvent evt = null!;
```
nblumhardt, line 469: Looks like one last one here, too :)

test/Serilog.Tests/Support/DelegatingSink.cs
```diff
@@ -20,7 +20,7 @@ public void Emit(LogEvent logEvent)
 
         public static LogEvent GetLogEvent(Action<ILogger> writeAction)
         {
-            LogEvent result = null;
+            LogEvent result = null!;
```
nblumhardt, line 23: I think I missed this one

SimonCropp: lol. why are they all named different ;)

PR #1592, opened 2021-06-21
Write to SelfLog to warn about settings with missing assembly references

### LINKED ISSUE #1539: KeyValuePairSettings to use SelfLog
**Is your feature request related to a problem? Please describe.**
I have been using [serilog-settings-xml](https://github.com/rsabirov/serilog-settings-xml) for configuration and couldn't get it to create a simple File sink.
The other settings were OK but not the File sink - it just didn't get created.

I was using the Configuration sample (on the README.md) for guidance but I don't think it was correct
Eventually, I tried adding `<add name="Serilog.Sinks.File" />` and it worked.

**Describe the solution you'd like**
Use SelfLog in KeyValuePairSettings to identify any KeyValue pairs that will effectively be ignored because of a missing Include or any other issue.

**Describe alternatives you've considered**
None

**Additional context**
I am a newbie to serilog.

octos4murai:
## Summary

- Writes a warning message in `SelfLog` when a setting is found to have no implementation in any of the loaded assemblies
- Helpful when a user misses a `using` reference

## Context
Fixes #1539

### DIFF
--- a/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
+++ b/src/Serilog/Settings/KeyValuePairs/KeyValuePairSettings.cs
@@ -19,6 +19,7 @@ using System.Reflection;
 using System.Text.RegularExpressions;
 using Serilog.Configuration;
 using Serilog.Core;
+using Serilog.Debugging;
 using Serilog.Events;
 
 namespace Serilog.Settings.KeyValuePairs
@@ -230,7 +231,12 @@ namespace Serilog.Settings.KeyValuePairs
             {
                 var target = SelectConfigurationMethod(configurationMethods, directiveInfo.Key, directiveInfo);
 
-                if (target != null)
+                if (target == null)
+                {
+                    SelfLog.WriteLine("Setting \"{0}\" could not be matched to an implementation in any of the loaded assemblies. " +
+                        "To use settings from additional assemblies, specify them with the \"serilog:using\" key.", directiveInfo.Key);
+                }
+                else
                 {
                     var call = (from p in target.GetParameters().Skip(1)
                                 let directive = directiveInfo.FirstOrDefault(s => s.ArgumentName == p.Name)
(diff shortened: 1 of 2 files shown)


test/Serilog.Tests/Settings/KeyValuePairSettingsTests.cs
```diff
@@ -124,6 +125,37 @@ public void SinksAreConfigured()
             Assert.Empty(DummyRollingFileAuditSink.Emitted);
         }
 
+        [Fact]
+        public void WritingToUndeclaredSinkWritesToSelfLog()
+        {
+            List<string> messages = new();
```
nblumhardt, line 131: Unfortunately, because the tests run in parallel, this will collide with others. Right now, I think the best workaround would be to move this one test over to `SelfLogTests`, add:

```csharp
            Messages = new();
            SelfLog.Enable(m =>
            {
                Messages = Messages ?? new();
                Messages.Add(m);
            });
```

instead of this block, and then remove the `SelfLog.Disable()` calls from this and the other test in that class.

Longer term, we should be able to move this to a test helper 🤔

nblumhardt (commented):
Looks great! Thought of a couple of really minor points, but otherwise think it's good to go :-)

test/Serilog.Tests/Settings/KeyValuePairSettingsTests.cs
octos4murai: Thanks for catching that. The new test does seem to fit in better under `SelfLogTests`. I don't think I can remove `SelfLog.Disable()` in the existing test though. It relies on that method call to do an additional test -- that nothing is written to `SelfLog` when it is disabled. Should I proceed to remove `SelfLog.Disable()` in the new test but keep the existing one as is?

nblumhardt: Looks like adding xUnit.net's `[CollectionDefinition("SelfLog", DisableTestParallelization = true)]` to the class would allow us to keep the `Disable()` calls in there.

octos4murai: It looks like parallelization is already disabled on the entire assembly: https://github.com/serilog/serilog/blob/dev/test/Serilog.Tests/Properties/AssemblyInfo.cs. When I run `./build.sh`, I get all tests passing.

nblumhardt:
Great! Thanks 👍

PR #1595, opened 2021-06-22
Make Nullable a private dependency; squeeze in a net5.0 target

nblumhardt:
Shouldn't need to be conditional this way - the `net5.0` version of Nullable ships nothing; the `netstandard2.1` version still supplies a couple of attributes (IIRC ;-)).


augustoproiete (approved):
👍🏻

nblumhardt:
Thanks @Numpsy, @augustoproiete 👍

PR #1597, opened 2021-06-24
Moar nullables

SimonCropp:
needs a squash

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -12,9 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#nullable enable
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Reflection;
 using Serilog.Core;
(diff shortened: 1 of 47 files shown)


src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -281,7 +283,7 @@ bool TryConvertValueTuple(object value, Destructuring destructuring, Type valueT
             return false;
         }
 
-        bool TryConvertCompilerGeneratedType(object value, Destructuring destructuring, Type valueType, out LogEventPropertyValue result)
+        bool TryConvertCompilerGeneratedType(object value, Destructuring destructuring, Type valueType, [NotNullWhen(true)] out LogEventPropertyValue? result)
```
nblumhardt, line 286: These are nicer to consume when they're written as:

```csharp
[MaybeNullWhen(false)] out LogEventPropertyValue result
```

I.e. the out var is not nullable, and the compiler checks that the return value has been used. (This is how `Dictionary<K,V>` works when `V` is a non-nullable reference type.)

SimonCropp: but LogEventPropertyValue is a reference type, so it is not comparable to the Dictionary scenario?

nblumhardt: That's a good question; I have a couple of things to check - will tinker and report back :-) 👍

SimonCropp: any outcome from the tinkering?

nblumhardt: Sorry, hectic couple of weeks again! Starting to catch up but slow going :)

SimonCropp: no rush on my side, just making sure it hadnt slipped off your radar

nblumhardt: Thanks :-) :+1:

nblumhardt: Thanks for hanging in there, mate - I think your approach is the right one, squashing now!

PR #1614, opened 2021-08-20
2.11.0 Release

nblumhardt:
WIP; think we should get full/final nullable reference type support into this one.


nblumhardt:
If you're able to try out `2.11.0-dev-01387`, this should be pretty close to the final build (just #1669 to go).

SimonCropp:
works on my machine

src/Serilog/ILogger.cs
```diff
@@ -1452,16 +1327,14 @@ void Fatal(Exception? exception, string messageTemplate, params object?[]? prope
         /// render the <paramref name="boundProperties"/> as text.</param>
         /// <param name="boundProperties">Captured properties from the template and <paramref name="propertyValues"/>.</param>
         /// <example>
-        /// <code>
         /// MessageTemplate template;
-        /// IEnumerable&lt;LogEventProperty&gt; properties;
+        /// IEnumerable&lt;LogEventProperty&gt; properties>;
```
bartelink, line 1331: this line does not look right (have not checked rest in detail)

nblumhardt: Nice spotting; pushed the fix to https://github.com/serilog/serilog/pull/1669 (this reimplements #1664, which had to be reverted)

PR #1635, opened 2021-11-15
(ReadOnly)Memory<byte> scalar policy; NotSupportedExeption handling for PropertyValueConverter

### LINKED ISSUE #1632: Log of arg throws NotSupportedException
This throws an `NotSupportedException: Specified method is not supported.` exception:
```c#
var serilogLogger = new LoggerConfiguration().CreateLogger();
var microsoftLogger = new SerilogLoggerFactory(serilogLogger).CreateLogger<ExceptionTests>();
microsoftLogger.LogInformation("{@asd}", new ReadOnlyMemory<byte>());
```
Repo: https://github.com/matthiaslischka/SerilogMethodNotSupportedException

Its caused by the property `public unsafe ReadOnlySpan<T> Span` on `GetValue()`.
I did not yet debug any further to see why exactly this is an issue but I think this special case should be handled by serilog.

BR Matthias

skomis-mm:
fixes #1632

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -1,4 +1,4 @@
-// Copyright 2013-2017 Serilog Contributors
+// Copyright 2013-2021 Serilog Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


nblumhardt:
Looks like the best we can do for now :+1:

nblumhardt (approved):

PR #1672, opened 2022-05-04
Finish adding nullables 

SimonCropp:


src/Serilog/Context/ImmutableStack.cs
```diff
@@ -23,9 +23,10 @@
 namespace Serilog.Context
 {
     class ImmutableStack<T> : IEnumerable<T>
+        where T : class
     {
-        readonly ImmutableStack<T> _under;
-        readonly T _top;
+        readonly ImmutableStack<T> _under = null!;
```
nblumhardt, line 28: Seems like `_under` is expected to be null on regular code paths - should perhaps be `ImmutableStack<T>?` here?

src/Serilog/Context/LogContext.cs
```diff
@@ -233,6 +233,11 @@ static ImmutableStack<ILogEventEnricher> Enrichers
             }
             set
             {
+                if (value == null)
+                {
+                    throw new ArgumentException();
```
nblumhardt, line 238: Should this check be here? (If the value is nullable, why not accept null?)

If the flow is correct - this should be `ArgumentNullException` I think.

src/Serilog/Settings/KeyValuePairs/SettingValueConversions.cs
```diff
@@ -31,11 +32,16 @@ class SettingValueConversions
...
+                { typeof(Type), s => Type.GetType(s, throwOnError: true)! },
             };
 
-        public static object ConvertToType(string value, Type toType)
+        public static object? ConvertToType(string? value, Type toType)
         {
+            if (value == null)
+            {
+                return null;
+            }
```
nblumhardt, line 43: I think this will break; if I ask for `ConvertToType(null, DateTimeKind)`, `null` is not a valid return value (it'll just trigger a failure further up the stack, when we try to use `null` as a `DateTimeKind`).

nblumhardt (commented):
Awesome! 😎 Had a couple of minor comments, LGTM though!

PR #1673, opened 2022-05-04
target and test on net6

SimonCropp:


nblumhardt:
Looks like we need to use the `Visual Studio 2022` image, via appveyor.yml.

SimonCropp:
done

nblumhardt:
Hmm not sure what's going on for the UWP native toolchain build; may need some other deps on the build image? I'm not sure how much value that step in the build is currently providing, it'd be useful to dig through it and update/disable as needed (unsure who has the most knowledge/context to contribute, there...)

skomis-mm:
Change should be made to this [this](https://github.com/serilog/serilog/blob/dev/BuildNativeUWP.ps1#L8) line. 
I beleive this step can be safely removed. It was added for rare cases like #993 related to compiler bug, but since then I don't remember any regressions.

nblumhardt:
👍 let's scrap it entirely; thanks @skomis-mm

SimonCropp:
so delete `BuildNativeUWP.ps1` ?

nblumhardt:
@SimonCropp yes, I think so :+1:

SimonCropp:
done

Issue #1697, opened 2022-07-14
PushProperty overload to take Func<T> delegate


mattheys:
**Is your feature request related to a problem? Please describe.**
Not a problem.

**Describe the solution you'd like**
I want to be able to asign a function delegate to PushProperty so that each time a long entry is written the result of the delgate is used for that property.

**Describe alternatives you've considered**
Manually logging that property each time on every log call.

**Additional context**
I was hoping I could pass a Func<T> like this example and have it get the result for each log entry, saving me from having to log it separetly each time and ensuring consistency.  I understand this might add an overhead on something that logs a lot or a function that takes a long time to return.  Maybe a separate enrichment package so the cavets are clearer?

As you can see `elapsed` just logs the Func<T> ToString() rather than executing it.

```C#
using Serilog;
using Serilog.Context;

Log.Logger = new LoggerConfiguration()
                    .Enrich.FromLogContext()
                    .WriteTo.Console(formatter: new Serilog.Formatting.Compact.CompactJsonFormatter())
                    .CreateLogger();

var stopWatch = System.Diagnostics.Stopwatch.StartNew();
Func<long> GetMillis = () => stopWatch.ElapsedMilliseconds;

using (LogContext.PushProperty("elapsed", GetMillis))
{
    for (int i = 0; i < 10; i++)
    {
        Log.Information("Attempt {i} in {ms}", i, GetMillis());
        Thread.Sleep(1000);
    }
}

Console.ReadKey();
```

```
{"@t":"2022-07-14T15:55:37.1183411Z","@mt":"Attempt {i} in {ms}","i":0,"ms":1,"elapsed":"System.Func`1[System.Int64]"}
{"@t":"2022-07-14T15:55:38.1433272Z","@mt":"Attempt {i} in {ms}","i":1,"ms":1036,"elapsed":"System.Func`1[System.Int64]"}
{"@t":"2022-07-14T15:55:39.1641471Z","@mt":"Attempt {i} in {ms}","i":2,"ms":2057,"elapsed":"System.Func`1[System.Int64]"}
{"@t":"2022-07-14T15:55:40.1801729Z","@mt":"Attempt {i} in {ms}","i":3,"ms":3073,"elapsed":"System.Func`1[System.Int64]"}
{"@t":"2022-07-14T15:55:41.1921403Z","@mt":"Attempt {i} in {ms}","i":4,"ms":4085,"elapsed":"System.Func`1[System.Int64]"}
[...]

PR #1698, opened 2022-07-14
Add an overload to LogContext.PushProperty to accept a Func<T> delegate.

mattheys:
I logged an issue #1697  but didn't realise how easy it would be to add the functionaility and do a pull request.

I'm not sure how to best test this, especially to make sure it's thread safe.  I've tried the below which both work.

```C#
using Serilog;
using Serilog.Context;

                    .Enrich.FromLogContext()
                    .WriteTo.Console(formatter: new Serilog.Formatting.Compact.CompactJsonFormatter())
                    .CreateLogger();

var stopWatch = System.Diagnostics.Stopwatch.StartNew();
Func<long> GetMillis = () => stopWatch.ElapsedMilliseconds;

using (LogContext.PushProperty("elapsed", GetMillis))
{
    for (int i = 0; i < 10; i++)
    {
        var t = new System.Threading.Thread(() => {
            Log.Information("Thread attempt {i}", i);
            Thread.Sleep(1000);
        });

        t.Start();
        t.Join();
    }

    for (int i = 0; i < 10; i++)
    {
        await Task.Run(async () =>
        {
            Log.Information("Task attempt {i}", i);
            await Task.Delay(1000);
        });
    }

}

Console.ReadKey();




```

### DIFF
--- a/src/Serilog/Core/Enrichers/FunctionPropertyEnricher.cs
+++ b/src/Serilog/Core/Enrichers/FunctionPropertyEnricher.cs
@@ -1,0 +1,67 @@
+// Copyright 2013-2015 Serilog Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#nullable enable
+using System;
+using Serilog.Events;
+
+namespace Serilog.Core.Enrichers
+{
+    /// <summary>
+    /// Adds a new property enricher to the log event.
+    /// </summary>
+    public class FunctionPropertyEnricher<T> : ILogEventEnricher
+    {
+        readonly string _name;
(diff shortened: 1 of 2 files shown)


src/Serilog/Core/Enrichers/FunctionPropertyEnricher.cs
```diff
@@ -0,0 +1,67 @@
...
+        readonly string _name;
+        readonly Func<T> _value;
+        readonly bool _destructureObjects;
+
+        /// <summary>
+        /// Create a new property enricher.
+        /// </summary>
+        /// <param name="name">The name of the property.</param>
+        /// <param name="value">The delegate to the property.</param>
+        /// <returns>A handle to later remove the property from the context.</returns>
```
sungam3r, line 35:
```suggestion
```

nblumhardt:
Thanks for this @mattheys! As you've discovered, `LogContext.Push()` enables this use case without needing the additional overload on `ILogger` etc.

As it's a bit of a niche scenario, and already well-supported without additional code in the core, I think we should leave things as-is.

src/Serilog/Core/Enrichers/FunctionPropertyEnricher.cs
mattheys: Interesting you flagged this to remove as I took the PropertyEnricher.cs as the template to do this and that line is also in there.  However this isn't getting merged anyway but you might want to double check this line to make sure it's right.   https://github.com/serilog/serilog/blob/81e9dee4aaa86538f9e754fa9887a8603800274a/src/Serilog/Core/Enrichers/PropertyEnricher.cs#L35

PR #1699, opened 2022-07-15
Remove wrong tag

sungam3r:


src/Serilog/Core/Enrichers/PropertyEnricher.cs
```diff
@@ -32,7 +32,6 @@ public class PropertyEnricher : ILogEventEnricher
         /// </summary>
         /// <param name="name">The name of the property.</param>
         /// <param name="value">The value of the property.</param>
-        /// <returns>A handle to later remove the property from the context.</returns>
         /// <param name="destructureObjects">If true, and the value is a non-primitive, non-array type,
         /// then the value will be converted to a structure; otherwise, unknown types will
         /// be converted to scalars, which are generally stored as strings.</param>
```
bartelink, line 37: is L39 superfluous too?

sungam3r: EnsureValidName throws this exception

bartelink: Apologies; L38 in your version, i.e. wondering (without having it in an IDE) whether `/// <returns></returns>` is necessary before the exception list (there isn't one before the method after, but perhaps it is necessary for a ctor)

sungam3r: ooops, fixed

nblumhardt:
Thanks @sungam3r, looks good!

PR #1701, opened 2022-07-25
add extra constructors for LogEventProperty

SimonCropp:
i have code like this in many places

```
var properties = new List<LogEventProperty>
{
    new("type", new ScalarValue(type)),
    new("endDate", new ScalarValue(endDate)),
    new("deleteCount", new ScalarValue(briefs.Count)),
    new("futureCount", new ScalarValue(futureCount)),
    new("briefs", ToDictionary(briefs)),
};
```

it would be nice if i could do the following for known scalar types
```
var properties = new List<LogEventProperty>
{
    new("type", type),
    new("endDate", endDate),
    new("deleteCount", briefs.Count),
    new("futureCount", futureCount),
    new("briefs", ToDictionary(briefs)),
};
``

### CODE REVIEW

**Comment by nblumhardt** [comment]
I'm in two minds about this one - the syntactic brevity is nice, but since it's not universal (e.g. I'd still need to write `new LogEventProperty(..., new ScalarValue(myScalarType))` I don't know if it's worth the extra baggage.

_In the olden days_ we'd have added `LogEventPropety.CreateScalar(name, value)`, but target-typed `new` makes that even more verbose in modern .NET 🤔

**Comment by SimonCropp** [comment]
> for types not in the included set) I don't know if it's worth the extra baggage.

two points

 * the majority of my usage of `LogEventProperty` is for known types
 * while it looks like a lot of baggage in the diff, it is little effort to maintain and delivers non trivial value

but fully respect whatever decision you go with

**Comment by SimonCropp** [comment]
do u care about cls compliance?

File: src/Serilog/Events/LogEventProperty.cs [hunk 1]
```diff
@@ -39,6 +39,113 @@ public LogEventProperty(string name, LogEventPropertyValue value)
             Value = value;
         }
 
+#if NET6_0_OR_GREATER
+        /// <inheritdoc />
+        public LogEventProperty(string name, TimeOnly value) : this(name, new ScalarValue(value))
```
nblumhardt, line 44: Perhaps naming the `value` argument in these overloads `scalarValue` would make the difference between them and the original constructor clearer?

nblumhardt:
I'm not sure whether we do, it's been a long time since I checked in on the relevance of CLS compliance in the modern .NET world - needs a quick bit of investigation 🤔

SimonCropp:
yeah i dropped CLS compliance for all my libs several years ago. so far no one has complained

SimonCropp:
@nblumhardt do u want to make a call on this one?

PR #1702, opened 2022-07-25
DateOnly and TimeOnly support

SimonCropp:


src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -42,7 +42,10 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
             typeof(long), typeof(ulong), typeof(float), typeof(double), typeof(decimal),
             typeof(string),
             typeof(DateTime), typeof(DateTimeOffset), typeof(TimeSpan),
-            typeof(Guid), typeof(Uri)
+            typeof(Guid), typeof(Uri),
+            #if NET6_OR_GREATER
```
nblumhardt, line 46: Rather than use TFM-based conditionals, can we use a flag like `FEATURE_DATE_AND_TIME_ONLY` here?

We've stuck to this scheme for a while since it makes it a bit easier to tease apart what is enabled for each target.

nblumhardt (commented):
Looks great, just had one minor suggestion

src/Serilog/Capturing/PropertyValueConverter.cs
SimonCropp: yep. will do it now

SimonCropp: done

nblumhardt, line 46: Missed this one?

nblumhardt:
😎

SimonCropp:
@nblumhardt thanks

SimonCropp:
will seq support properties that are dateonly and time only? like: https://docs.datalust.co/docs/working-with-dates-and-times

nblumhardt:
@SimonCropp it'll need some thought 🤔  - thanks for the nudge

PR #1703, opened 2022-07-25
add explicit net6 target

SimonCropp:


sungam3r:
This is not the first time I see such PRs. It’s just interesting - what gives an addition of net6 TFM if the compiled code is no different from that for net5 (no ifdefs for net6)?

nblumhardt:
@sungam3r I think this is the foundation for https://github.com/serilog/serilog/pull/1702

nblumhardt:
LGTM @SimonCropp - thanks!

PR #1711, opened 2022-07-27
add missing VALUETUPLE check

SimonCropp:
`VALUETUPLE` was being used but not defined in any constants

```
#if VALUETUPLE
            if (definition == typeof(ValueTuple<>) || definition == typeof(ValueTuple<,>) ||
                definition == typeof(ValueTuple<,,>) || definition == typeof(ValueTuple<,,,>) ||
                definition == typeof(ValueTuple<,,,,>) || definition == typeof(ValueTuple<,,,,,>) ||
                definition == typeof(ValueTuple<,,,,,,>))
#else
            // ReSharper disable once PossibleNullReferenceException
            if (definition.FullName is
                "System.ValueTuple`1" or
                "System.ValueTuple`2" or
                "System.ValueTuple`3" or
                "System.ValueTuple`4" or
                "System.ValueTuple`5" or
                "System.ValueTuple`6" or
                "System.ValueTuple`7")
#endif
```

### DIFF
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -3,7 +3,7 @@
     <Description>Simple .NET logging with fully-structured events</Description>
     <VersionPrefix>2.12.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.3;netstandard1.0;net45;net46;net5.0;net6.0</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


SimonCropp:
side note is there any real value in doing type check here? wouldn't `definition.FullName.StartsWith("System.ValueTuple``"` suffice

skomis-mm:
Looks good 👍 

> wouldn't definition.FullName.StartsWith("System.ValueTuple``" suffice

I think since we constrain length of the tuples to 8 for destructuring type name checks against interned strings should perform better than `StartsWith(..)`

PR #1722, opened 2022-08-02
nullable in MessageTemplateParser

SimonCropp:


src/Serilog/Parsing/MessageTemplateParser.cs
```diff
@@ -161,47 +163,44 @@ static bool TrySplitTagContent(string tagContent, out string propertyNameAndDest
                 propertyNameAndDestructuring = tagContent;
                 format = null;
                 alignment = null;
+                return true;
```
SimonCropp, line 166: i did some early returns to simplify the branching logic here. let me know if u prefer it back the previous way

nblumhardt: LGTM :+1:

PR #1725, opened 2022-08-08
Introduce file-scoped namespaces

sungam3r:
File-scoped namespace setting applied for two aux projects so there should not be any conflict with other PRs. No code changes.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,3 +16,6 @@ end_of_line = lf
 
 [*.{cmd, bat}]
 end_of_line = crlf
+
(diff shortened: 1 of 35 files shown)


SimonCropp:
or we could delete all the namespaces in the test project since they add no value 😄

nblumhardt (approved):

nblumhardt:
Looks great, thanks @sungam3r :+1:

sungam3r:
Note that for existing code switching to file-scoped namespaces may be completely destructive for pending PRs.

PR #1729, opened 2022-08-23
Finish nullable. now enabled globally

SimonCropp:


src/Serilog/Core/Logger.cs
```diff
@@ -1328,7 +1327,7 @@ public void Fatal(Exception? exception, string messageTemplate, params object?[]
         /// }
         /// </example>
         [MessageTemplateFormatMethod("messageTemplate")]
-        public bool BindMessageTemplate(string messageTemplate, object?[]? propertyValues,
+        public bool BindMessageTemplate(string? messageTemplate, object?[]? propertyValues,
```
nblumhardt, line 1330: I don't think the intention is for `messageTemplate` to be nullable here; IIRC the dynamic check against null is just to preserve the "no throw" guarantee (as for the same method on `ILogger`).

src/Serilog/Events/LogEventProperty.cs
```diff
@@ -67,7 +68,7 @@ internal LogEventProperty(EventProperty property)
         /// </summary>
         /// <param name="name">The name to check.</param>
         /// <returns>True if the name is valid; otherwise, false.</returns>
-        public static bool IsValidName(string name) => !string.IsNullOrWhiteSpace(name);
+        public static bool IsValidName([NotNullWhen(true)] string? name) => !string.IsNullOrWhiteSpace(name);
```
nblumhardt, line 71: Not sure the attribute makes sense here?

nblumhardt (commented):
This is awesome! Couple of minor queries :-)

src/Serilog/Core/Logger.cs
SimonCropp: fixed

src/Serilog/Events/LogEventProperty.cs
SimonCropp: without this we get build errors, since the compiler cant derive the nullability

```
  Logger.cs(1369, 65): [CS8604] Possible null reference argument for parameter 'name' in 'LogEventProperty MessageTemplateProcessor.CreateProperty(string name, object? value, bool destructureObjects = false)'.
```

nblumhardt: Wow! TIL, thanks :+1:

PR #1742, opened 2022-09-01
guard against null

SimonCropp:


nblumhardt:
Definitely an improvement; it's a shame we can't use `ArgumentNullException.ThrowIfNull()` and that there's no generic argument-returning version of that method :-(

Perhaps since our eyes are all going to get trained to glance over `ArgumentNullException.ThrowIfNull()` we could try aligning the naming somehow? `ArgumentNullExceptionEx.ThrowIfNull()` is truly horrific, but ... could maybe live with it? `ANE.ThrowIfNull()`? :thinking: 

I might need a chance to ponder a bit; agreed we should push forward in this direction somehow though :+1:

SimonCropp:
note that `ArgumentNullException.ThrowIfNull` is not available in older frameworks. so even if it wasnt generic, we should still need to polyfill 

happy to name the type+method whatever u want. it could even be an extension method. so instead of `Guard.AgainstNull(scalarType);` it could be be `scalarType.NotNull();`

SimonCropp:
i must admit i much prefer `Guard.AgainstNull` over `ArgumentNullExceptionEx.ThrowIfNull`. less characters, and reads better

nblumhardt:
:laughing: yeah!

I'm not so much worried about the naming as drifting from what other codebases use and what the tooling supports.

E.g. in Rider I frequently add argument null checks using the

 * "Check argument for null" and
 * "Combine null check with assignment" actions,

both of which would generate inconsistent code once we switch to `Guard`.  We'll also have to watch out for incoming changes in PRs, or end up with an inconsistent codebase :thinking:

I'd swim against the current happily if the net positive was big enough, but with some more thought, I'm not sure it adds up in this case. What do you think?

SimonCropp:
i see your concern. two points

 * this is (mostly) for public APIs. which dont change too much in a codebase
 * i can write a unit test that looks for `ArgumentNullException` and fails with "please use `Guard.AgainstNull`"

SimonCropp:
> i can write a unit test that looks for ArgumentNullException and fails with "please use Guard.AgainstNull"

done

nblumhardt:
Nice! I'll loop back to this shortly. I'm wondering if we could use the same approach to enforce that we only use `Guard` on public APIs? :thinking:

SimonCropp:
>  I'm wondering if we could use the same approach to enforce that we only use Guard on public APIs?

it would be possible with cecil. but IMO it is reaching the point of diminishing returns

PR #1745, opened 2022-09-03
remove redundant LogEventLevel

SimonCropp:


nblumhardt:
Think this one is going to quietly cause my brain to switch into Rust mode (it's the common way to work with enum members there).. Looks very strange to me in C# but I think I'll adjust - nice for the test code, anyway! :-)

SimonCropp:
LOL. its not your granddads C# anymore

Issue #1749, opened 2022-09-08
Support `IAsyncDisposable` on `Logger` and sinks


nblumhardt:
Buffered/networked Serilog sinks often need to flush via asynchronous APIs on shutdown, and mechanisms for doing this in `IDisposable.Dispose()` are plagued with issues. (The current `PeriodicBatchingSink.CloseAndFlush()` implementation is the best we've come up with, but this still apparently [fails on AWS Lambda](https://github.com/datalust/serilog-sinks-seq/issues/167) in at least some circumstances.)

Supporting `IAsyncDisposable` thorughout Serilog, in a non-breaking way, is the right thing to do to fix this. We should aim to get this into 2.12.

nblumhardt:
Looking at this now :-)

PR #1750, opened 2022-09-09
Initial support for `IAsyncDisposable` sinks

nblumhardt:
Resolves https://github.com/serilog/serilog/issues/1749

This is admittedly very heavy on `#if`. We may be able to do better by polyfilling `IAsyncDisposable` and `ValueTask` for older platforms - keen to explore this, too.

Gets some tests and feature flags in place, though, which should be a good start. Eyes appreciated :-)

### DIFF
--- a/src/Serilog/Core/Sinks/DisposingAggregateSink.cs
+++ b/src/Serilog/Core/Sinks/DisposingAggregateSink.cs
@@ -14,7 +14,7 @@
 
 namespace Serilog.Core.Sinks;
 
-class DisposingAggregateSink : ILogEventSink, IDisposable
+sealed class DisposingAggregateSink : ILogEventSink, IDisposable
 #if FEATURE_ASYNCDISPOSABLE
     , IAsyncDisposable
 #endif
--- a/src/Serilog/Core/Sinks/DisposingAggregateSink.cs
+++ b/src/Serilog/Core/Sinks/DisposingAggregateSink.cs
@@ -74,7 +74,7 @@ sealed class DisposingAggregateSink : ILogEventSink, IDisposable
             {
                 try
                 {
-                    await asyncDisposable.DisposeAsync();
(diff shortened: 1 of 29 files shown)


src/Serilog/Core/Sinks/DisposingAggregateSink.cs
```diff
@@ -57,8 +60,44 @@ public void Dispose()
...
+#if FEATURE_ASYNCDISPOSABLE
+    public async ValueTask DisposeAsync()
+    {
+        foreach (var sink in _sinks)
+        {
+            if (sink is IAsyncDisposable asyncDisposable)
+            {
+                try
+                {
+                    await asyncDisposable.DisposeAsync();
```
skomis-mm, line 77: I think we should add:
```csharp
.ConfigureAwait(false);
```

src/Serilog/LoggerConfiguration.cs
```diff
@@ -172,17 +172,49 @@ public Logger CreateLogger()
...
-            new(processor, _minimumLevel, sink, enricher, Dispose, overrideMap) :
-            new(processor, _levelSwitch, sink, enricher, Dispose, overrideMap);
+#if FEATURE_ASYNCDISPOSABLE
+        async ValueTask DisposeAsync()
+        {
+            foreach (var disposable in disposableSinks)
+            {
+                if (disposable is IAsyncDisposable asyncDisposable)
+                {
+                    await asyncDisposable.DisposeAsync();
```
skomis-mm, line 199:
```csharp
.ConfigureAwait(false);
```

skomis-mm (commented):
Looks good overall 👍 
We should also implement `IAsyncDisposable` on the following types:
- ConditionalSink
- RestrictedSink
- SecondaryLoggerSink

nblumhardt:
Thank you @skomis-mm :+1: 

Fixed, and some more test coverage added :-)

SimonCropp (approved):

PR #1756, opened 2022-10-04
add Half support

SimonCropp:


src/Serilog/Formatting/Json/JsonValueFormatter.cs
```diff
@@ -172,7 +172,11 @@ protected virtual void FormatLiteralValue(object? value, TextWriter output)
 
         if (value is ValueType)
         {
-            if (value is int or uint or long or ulong or decimal or byte or sbyte or short or ushort)
+            if (value is int or uint or long or ulong or decimal or byte or sbyte or short or ushort
+#if FEATURE_HALF
+                or Half
```
nblumhardt, line 177: Since `Half` can be infinite or NaN (AFAIK), we should probably handle it here in the same way as `double` and `float` are handled below

SimonCropp:
@nblumhardt are there any tests i should replicate to verify this change?

src/Serilog/Formatting/Json/JsonFormatter.cs
```diff
@@ -85,6 +85,9 @@ public class JsonFormatter : ITextFormatter
             { typeof(uint), WriteToString },
             { typeof(long), WriteToString },
             { typeof(ulong), WriteToString },
+#if FEATURE_HALF
+            { typeof(Half), WriteToString },
```
nblumhardt, line 89: Same issue here, IIRC - needs a `WriteHalf`

nblumhardt:
I'm not sure about existing tests, it's pretty likely there will be some place to generally hook them in for both cases, but the single/double behavior might not be covered :thinking:

nblumhardt:
:wave: did you mean to close all of these?

SimonCropp:
shit. no. will re-submit

PR #1764, opened 2022-11-01
Enables updating the configured min-level overrides at runtime

bart-vmware:
In response to https://github.com/serilog/serilog-settings-configuration/issues/284#issuecomment-1289664499, this PR shows my attempt to add the ability to replace the set of minimum log level overrides at runtime. The goal is to make it possible to refresh when appsettings.json changes on disk.

My approach attempts to avoid compromising performance or multicore contention at:
https://github.com/serilog/serilog/blob/4d13be50c03e14b6072043799dc7e5dbe4139a19/src/Serilog/Core/LevelOverrideMap.cs#L35-L41
by atomically updating the `LevelOverrideMap` reference with a new snapshot, instead of switching to a concurrent collection.

### DIFF
--- a/src/Serilog/Configuration/LoggerMinimumLevelConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerMinimumLevelConfiguration.cs
@@ -22,15 +22,15 @@ public class LoggerMinimumLevelConfiguration
     readonly LoggerConfiguration _loggerConfiguration;
     readonly Action<LogEventLevel> _setMinimum;
     readonly Action<LoggingLevelSwitch> _setLevelSwitch;
-    readonly Action<string, LoggingLevelSwitch> _addOverride;
+    readonly LevelOverrideMapWrapper _wrapper;
 
     internal LoggerMinimumLevelConfiguration(LoggerConfiguration loggerConfiguration, Action<LogEventLevel> setMinimum,
-        Action<LoggingLevelSwitch> setLevelSwitch, Action<string, LoggingLevelSwitch> addOverride)
+        Action<LoggingLevelSwitch> setLevelSwitch, LevelOverrideMapWrapper wrapper)
     {
         _loggerConfiguration = Guard.AgainstNull(loggerConfiguration);
         _setMinimum = Guard.AgainstNull(setMinimum);
         _setLevelSwitch = setLevelSwitch;
-        _addOverride = Guard.AgainstNull(addOverride);
+        _wrapper = Guard.AgainstNull(wrapper);
     }
 
     /// <summary>
@@ -113,15 +113,12 @@ public class LoggerMinimumLevelConfiguration
     /// <exception cref="ArgumentNullException">When <paramref name="levelSwitch"/> is <code>null</code></exception>
     public LoggerConfiguration Override(string source, LoggingLevelSwitch levelSwitch)
     {
-        Guard.AgainstNull(source);
-        Guard.AgainstNull(levelSwitch);
-
-        var trimmed = source.Trim();
-        if (trimmed.Length == 0) throw new ArgumentException($"A source {nameof(source)} must be provided.", nameof(source));
-
-        _addOverride(trimmed, levelSwitch);
+        Overrides.Add(source, levelSwitch);
         return _loggerConfiguration;
     }
+
+    /// <summary/>
+    public LoggerMinimumLevelOverridesConfiguration Overrides  => new(_loggerConfiguration, _wrapper);
 
     /// <summary>
     /// Override the minimum level for events from a specific namespace or type name.
--- a/src/Serilog/Configuration/LoggerMinimumLevelOverridesConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerMinimumLevelOverridesConfiguration.cs
@@ -1,0 +1,48 @@
+namespace Serilog.Configuration;
+
+/// <summary/>
+public class LoggerMinimumLevelOverridesConfiguration
+{
+    readonly LoggerConfiguration _loggerConfiguration;
+    readonly LevelOverrideMapWrapper _wrapper;
+
+    internal LoggerMinimumLevelOverridesConfiguration(LoggerConfiguration loggerConfiguration, LevelOverrideMapWrapper wrapper)
+    {
+        _loggerConfiguration = Guard.AgainstNull(loggerConfiguration);
+        _wrapper = Guard.AgainstNull(wrapper);
+    }
+
+    /// <summary/>
+    public LoggerConfiguration Add(string source, LoggingLevelSwitch levelSwitch)
+    {
+        Guard.AgainstNull(source);
+        Guard.AgainstNull(levelSwitch);
+
+        var trimmed = source.Trim();
+        if (trimmed.Length == 0) throw new ArgumentException($"A {nameof(source)} must be provided.", nameof(source));
+
+        _wrapper.Add(trimmed, levelSwitch);
+
+        return _loggerConfiguration;
+    }
+
+    /// <summary/>
+    public LoggerConfiguration Set(IDictionary<string, LoggingLevelSwitch> overrides)
+    {
+        Guard.AgainstNull(overrides);
+
+        Dictionary<string, LoggingLevelSwitch> trimmedOverrides = new();
+
+        foreach (var levelOverride in overrides)
+        {
+            var trimmed = levelOverride.Key.Trim();
+            if (trimmed.Length == 0) throw new ArgumentException("A source must be provided.", nameof(overrides));
+
+            trimmedOverrides.Add(trimmed, levelOverride.Value);
+        }
+
+        _wrapper.Replace(trimmedOverrides);
+
+        return _loggerConfiguration;
+    }
+}
--- a/src/Serilog/Core/LevelOverrideMap.cs
+++ b/src/Serilog/Core/LevelOverrideMap.cs
@@ -57,6 +57,11 @@ class LevelOverrideMap
             .ToArray();
     }
 
+    internal LevelOverrideMap WithOverrides(IDictionary<string, LoggingLevelSwitch> overrides)
+    {
+        return new LevelOverrideMap(overrides, _defaultMinimumLevel, _defaultLevelSwitch);
+    }
+
     public void GetEffectiveLevel(
 #if FEATURE_SPAN
         ReadOnlySpan<char> context,
--- a/src/Serilog/Core/LevelOverrideMapWrapper.cs
+++ b/src/Serilog/Core/LevelOverrideMapWrapper.cs
@@ -7,6 +7,8 @@ class LevelOverrideMapWrapper
 
     public LevelOverrideMap? Map => _map;
 
+    public event EventHandler? OverridesChanged;
+
     public void Add(string source, LoggingLevelSwitch levelSwitch)
     {
         _overrides[source] = levelSwitch;
@@ -22,6 +24,7 @@ class LevelOverrideMapWrapper
     public void RefreshMap()
     {
         _map = _map?.WithOverrides(_overrides);
+        OverridesChanged?.Invoke(this, EventArgs.Empty);
     }
 
     public void EnsureMap(LogEventLevel defaultMinimumLevel, LoggingLevelSwitch? defaultLevelSwitch)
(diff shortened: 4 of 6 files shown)


bart-vmware:
To test this, I changed the sample at https://github.com/serilog/serilog-settings-configuration/blob/dev/sample/Sample/Program.cs in the following way:

Replace:

```c#
                .ReadFrom.Configuration(configuration)
                .CreateLogger();

```

with:

```c#
            var loggerConfiguration = new LoggerConfiguration()
                .ReadFrom.Configuration(configuration);

            var overridesConfiguration = loggerConfiguration.MinimumLevel.Overrides;

            var logger = loggerConfiguration.CreateLogger();
```

And replace the logger lines inside the `do` loop with:

```c#
                logger.ForContext(Constants.SourceContextPropertyName, "MyApp.Something.Tricky").Verbose("Hello, world!");
                logger.ForContext(Constants.SourceContextPropertyName, "MyApp.Something.Else").Verbose("Hello, world!");

                Task.Run(() =>
                {
                    var newOverrides = new Dictionary<string, LoggingLevelSwitch>()
                    {
                        ["MyApp.Something.Else"] = new(LogEventLevel.Verbose)
                    };
                    overridesConfiguration.Set(newOverrides);
                }).Wait();
```

When running, it prints the following output:

```
[11:01:54  [Information] Args: []
[11:01:54 MyApp.Something.Tricky [Verbose] Hello, world!

Press "q" to quit, or any other key to run again.

[11:01:55 MyApp.Something.Else [Verbose] Hello, world!

Press "q" to quit, or any other key to run again.
```

shvez:
commenting to be aware of changes here

nblumhardt:
Thanks @bart-vmware. If you create all of your loggers outside of the loop (i.e. don't make fresh `ForContext()` calls after reconfiguring the override map) then I tdon't think the existing loggers will update to reflect the new levels. (This is the kind of predictability issue that might be hard to solve without regressing performance, IIRC.)

bart-vmware:
Thanks, I see what you mean. I've pushed an additional commit to address that and updated the sample accordingly to make it work:
```c#
            var trickyLogger = logger.ForContext(Constants.SourceContextPropertyName, "MyApp.Something.Tricky");
            var elseLogger = logger.ForContext(Constants.SourceContextPropertyName, "MyApp.Something.Else");

            do
            {
                trickyLogger.Verbose("Hello, world!");
                elseLogger.Verbose("Hello, world!");

                // ...
```

nblumhardt:
Thanks for taking a look! Unfortunately, the `IDisposable` implementation on `Logger` is only exposed on the root logging pipeline; child loggers like those produced by `ForContext` are exposed only as the non-disposable `ILogger`, and the API doesn't require or allow disposing these, so the changes would leak all child loggers via their event subscriptions.

The event handling subscribe/unsubscribe/synchronization costs are also a significant factor since in Serilog's design, contextual loggers are cheap/throwaway - it's common to see them created at pretty high frequency.

Digging through this again is reminding me why my conclusions last time were that it's a feature that's somewhat at odds with the way Serilog is designed. Still happy digging in, here, but thinking chances of finding a successful approach might be modest :-)

Just thinking back to your original ticket for a second - have you investigated the reconfigurable "reloadable logger" approach that we currently bundle up with Serilog.Extensions.Hosting?:

https://github.com/serilog/serilog-extensions-hosting/blob/dev/src/Serilog.Extensions.Hosting/Extensions/Hosting/ReloadableLogger.cs

It's fully reconfigurable; it's designed to be eventually "frozen" into a permanent, non-reconfigurable form, but you may have some luck incorporating it/tweaking it to reduce the costs of keeping it in a permanently mutable state 🤔 ... let me know if this might be a lower-effort way to get the behavior you're after.

bart-vmware:
As far as I'm aware, logger instances are cached and reused when created by LoggerFactory [here](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging/src/LoggerFactory.cs#L135). As a compromise, we could do the costly event subscription only when the call originates from `SerilogLoggerProvider.CreateLogger`, but not from child loggers created directly using the public `Logger.ForContext`. That would enable only long-lived loggers to respond to configuration changes.

`ReloadableLogger` doesn't fit our use case. It uses extensive locking as-is, but more importantly, our library is not in control of the Serilog initialization. Consuming applications contain the actual Serilog setup code. While it's okay for us to document that consumers need to add something to their configuration or call a built-in Serilog method, we prefer not to interfere much with the Serilog implementation by replacing parts of it, because that's hard to maintain over time when inner workings change.

nblumhardt:
Thanks for the follow-up. In the typical `UseSerilog()` configuration, the MEL `LoggerFactory` isn't invoked - it's replaced by Serilog's. I think your approach might be applicable to how `AddSerilog()` works, but we generally discourage using it because of other problems around leveling and filtering. HTH!

sungam3r:
@bart-vmware If you rebase your PR to fresh `dev` then proposed changes to public API will become observable.

shvez:
@sungam3r does it mean that we will be able to add new loggers and switch levels on the fly?

sungam3r:
No. I just said to rebase to incorporate changes from #1783.

bart-vmware:
@nblumhardt
> In the typical UseSerilog() configuration, the MEL LoggerFactory isn't invoked - it's replaced by Serilog's. I think your approach might be applicable to how AddSerilog() works, but we generally discourage using it because of other problems around leveling and filtering.

Yes, I was assuming the use of `AddSerilog()`. I wasn't aware this is discouraged. Is that documented somewhere?
Either way, if using Serilog through the ASP.NET `ILogger` abstractions is bad practice, I don't see the point of pursuing this any further.

@sungam3r 
Thanks for the heads-up. I'll incorporate your changes at a later time, if at all. Right now, it is more likely that this PR is going to be abandoned unmerged.

sungam3r:
> if using Serilog through the ASP.NET ILogger abstractions is bad practice

I think about that as sync over async or async over sync transitions. If you can use Serilog APIs directly from your code then just use it instead of jumping into MEL layer only to get control back into Serilog some time later.

bart-vmware:
@sungam3r In an application landscape with various third-party libraries and complex dependencies, having a vendor-neutral logging abstraction in the platform is essential. If every NuGet package that logs took a dependency on Serilog, it would only work if all of them use the exact same Serilog version at any time, worldwide. That's not so practical. That's basically the primary reason why System.Text.Json was created, as well as integrating OpenTelemetry into the platform.

I'm closing this, we no longer have interest in these changes.

sungam3r:
> In an application landscape with various third-party libraries and complex dependencies, having a vendor-neutral logging abstraction in the platform is essential. 

I know and because I wrote my own vendor-neutral logging abstraction a long time ago for all projects that I work on 😄. However, it really uses/maps only Serilog. We dropped NLog support a long time ago. Over time, this layer of abstraction is overgrown with its own functionality in addition to the simple adaptation of various logging frmeworks. So we decided to continue using it.

>  If every NuGet package that logs took a dependency on Serilog, it would only work if all of them use the exact same Serilog version at any time, worldwide.

This can be said for absolutely any package that is very widespread. These are general problems of backward compatibility,

deathchurch:
Where is it documented that using MSEL combined with Serilog is not best practice? Not sure why you would want to go all in with Serilog when you can use MSEL in code and have Serilog do the heavy lifting in the background.

sungam3r:
> Where is it documented that using MSEL combined with Serilog is not best practice?

I didn't say that. You _may_ combine them, eventually all logging will fall into serilog factory if you use `UseSerilog` at the app top level. If you can’t do something  in your own code via MEL so that it is translated into a Serilog, then just do not do it - use Serilog API directly.

nblumhardt:
This one's going to be a matter of personal/team tastes. Can't help adding my $0.02 :-)

Reasons I generally use Serilog's APIs directly -

 * Static `Log` class avoids the need to pass loggers around or make everything a container-created component just to consume logging
 * `LogContext.PushProperty("Name", value)` and `ILogger.ForContext("Name", value)` are marginally more concise than `ILogger.BeginScope()`
 * Fits a wider range of apps/frameworks (doesn't require setting up DI, config, etc.)
 * Shorter logging method names :-)

Other than those, I don't think there's much difference at the API level. MEL is a nice enough structured logging abstraction that I can live with either one.

deathchurch:
Thanks both for the clarification, for a minute I thought I had missed some best practice advice and the MSEL/Serilog combination was not the right thing to be doing.

nblumhardt:
Hi @beppemarazzi - IIRC this is mostly around synchronization. I'd guess that there's little-to-no observable difference for most apps but YMMV; checking the `ReloadableLogger` source and benchmarking would be the only way to say definitively.

beppemarazzi:
@nblumhardt  many thanks for your kind reply!!!

following your suggestion, I'm digging into the `ReloadableLogger` source code... 
Probably given my use case, the "no compromise" option is to create a specific `ReloadableNotFreezableLogger` without all the `_frozen` logic and minimize all the involved overhead ...  BUT ... i see also a "best effort" option keeping the current implementation AS-IS or with a very little modifications.

In detail I'm looking at this function:

https://github.com/serilog/serilog-extensions-hosting/blob/4ec0cc22077a09b977f59f4283d74e43d6aa40b8/src/Serilog.Extensions.Hosting/Extensions/Hosting/CachingReloadableLogger.cs#L189-L202

```
    void Update(ILogger newRoot, ILogger newCached, bool frozen)
    {
        _root = newRoot;
        _cached = newCached;
            
        // https://github.com/dotnet/runtime/issues/20500#issuecomment-284774431
        // Publish `_cached` and then `_frozen`. This is useful here because it means that once the logger is frozen - which
        // we always expect - reads don't require any synchronization/interlocked instructions.
        Interlocked.MemoryBarrierProcessWide();

        _frozen = frozen;

        Interlocked.MemoryBarrierProcessWide();
    }
```

I'm not so expert in lock-free programming, so i cannot fully understand the purpose of `Interlocked.MemoryBarrierProcessWide`, but i know it's quite expensive.
I'm not scared about the case when the configuration actually changes because it's very rare; but about the fact that this function is called also every time a new child logger is created and first used.
So I'm wondering if we could bypass this part in case we are not freezing. Something like this:

```
    void Update(ILogger newRoot, ILogger newCached, bool frozen)
    {
        _root = newRoot;
        _cached = newCached;
        
        if(!frozen)
          return; //we are not changing the _frozen state!!!!!!

   
        // https://github.com/dotnet/runtime/issues/20500#issuecomment-284774431
        // Publish `_cached` and then `_frozen`. This is useful here because it means that once the logger is frozen - which
        // we always expect - reads don't require any synchronization/interlocked instructions.
        Interlocked.MemoryBarrierProcessWide();

        _frozen = frozen;

        Interlocked.MemoryBarrierProcessWide();
    }
```

nblumhardt:
Hi @beppemarazzi; I suspect that change will cause `_cached` to not be visible across all threads when changed, though it takes me a while to get back up to speed with this code every time I revisit it, too.

beppemarazzi:
> Hi @beppemarazzi; I suspect that change will cause `_cached` to not be visible across all threads when changed, though it takes me a while to get back up to speed with this code every time I revisit it, too.

Understood. 
Again: i don't really know the impact of lock vs memorybarrier and how it actually impact on the whole application, so probably I'm wasting time trying to "optimize" it (also because I've no any benchmarck supporting this!!!!)
BUT I've tried to do some refactoring so that when the `Update` function is called it's under the lock in unfrozen case... this may ensure data consistency for all threads without the needing of memorybarriers... (actually my plan is to replace the simple lock with a `ReaderWriterLockSlim` to minimize contention)

If you want to take a look and you are interested in this feature (RuntimeConfigurationReloading) i could make a PR for this...

Thanks again for your support and your wonderful projects!!!!!!

beppemarazzi:
Hi @nblumhardt 

I've just published my fork of `ReloadableLogger`.
In few words it's my attempt to remove the memory-barrier in the "unfrozen" case and to replace the lock with a `ReaderWriterLockSlim` to reduce the lock contention and improve the concurrency in the normal case (someone is logging and there is not any reloading in flight)
If you are interested take a look here: https://github.com/butrockets/serilog-unofficial-hotreloading

Few points:

1. licensing: i keep your  Apache-2.0 license adding my name after "serilog contributors" in the heading of modified files. Is it ok for you?
2. doing this work i realized that it's totally unrelated to `Microsoft.Extension.Hosting` so i removed the relative dependency. IMHO it could be used also without framework abstractions...
3. your code review may be appreciated a lot!!!
3. if you want i could do a PR (against `serilog-extensions-hosting`? against `serilog-what`? )

Many thanks again!

nblumhardt:
Very nice! Thanks for posting your results, all looks good, I'll try to run through and review when I have a moment 👍

PR #1766, opened 2022-11-11
Update to .NET 7.0

eslamisepehr:
Hello 👋

Update Projects to .NET 7.0.

### DIFF
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -1,9 +1,9 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Description>Simple .NET logging with fully-structured events</Description>
     <VersionPrefix>2.12.1</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.3;netstandard1.0;net45;net46;net47;net5.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.3;netstandard1.0;net45;net46;net47;net5.0;net6.0;net7.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>serilog;logging;semantic;structured</PackageTags>
     <PackageIcon>icon.png</PackageIcon>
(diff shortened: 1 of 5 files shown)


nblumhardt:
Thanks! The `install` block from [this appveyor.yml file I just set up](https://github.com/serilog/serilog-sinks-browserconsole/blob/dev/appveyor.yml#L5) will sort out the build side.

src/Serilog/Serilog.csproj
```diff
@@ -1,9 +1,9 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Description>Simple .NET logging with fully-structured events</Description>
     <VersionPrefix>2.12.1</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.3;netstandard1.0;net45;net46;net47;net5.0;net6.0</TargetFrameworks>
```
nblumhardt, line 6: May not be much point yet adding this, since our dependencies and features will match net6.0 exactly, and it's the LTS.

test/Serilog.PerformanceTests/Serilog.PerformanceTests.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
```
nblumhardt, line 3: Dropping 6.0 and perf testing on 7 only seems fine.

test/TestDummies/TestDummies.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;net452</TargetFrameworks>
```
nblumhardt, line 4: May not be much to gain in this project.

src/Serilog/Serilog.csproj
WeihanLi: Should we consider removing `net5.0` since it's out of support

eslamisepehr:
According to the comments, there is no point to update except for tests.
What should I do?

nblumhardt:
Hi @eslamisepehr - updating the tests and SDK version would still be a worthwhile change, if you're interested in giving that a shot?

src/Serilog/Serilog.csproj
sungam3r: It gives a very little benefit. Dropping `netstandard1.0;net45;etc` will give more noticeable benefits. I would not drop so.

SimonCropp:
done in another PR

augustoproiete:
> done in another PR

https://github.com/serilog/serilog/pull/1806

Issue #1769, opened 2022-11-29
Optimize memory allocations


igor84:
**Is your feature request related to a problem? Please describe.**
I am considering replacing our in house logging solution for Unity3D with Serilog but it allocates 3x more memory for logging a simple string.

**Describe the solution you'd like**
Using a few quick experiments  I found that we could bring the allocations to the same minimal level with these 3 changes:
1. Use foreach over _outputTemplate.TokenArray instead of Tokens IEnumerator in MessageTemplateTextFormatter.Format method so no IEnumerator is allocated.
2. Pool and reuse internally created StringWriters using writer.GetStringBuilder.Clear().
3. Pool and reuse LogEvent instances.

**Additional context**
If you agree this would be useful I could make a pull request. It would be nice if I could use ConcurrentBag<T> for the pools but we would have to sack .netstandard1.0 since it is not supported there.

sungam3r:
👍  I would recommend to split into separate PRs to make changes measurable and use assets from Serilog.PerformanceTests to approve improvements.

sungam3r:
These files are not tests actually, these are benchmarks. Docs here https://github.com/dotnet/BenchmarkDotNet
I see some wrapper around benchmarks - https://github.com/serilog/serilog/blob/dev/test/Serilog.PerformanceTests/Harness.cs#L18

igor84:
I found that I was getting the error that Serilog.snk file could not be found. I had to copy the assets folder to test/Serilog.PerformanceTests/bin/Release/ for benchmarks to start working.

Anyway, for pools implementation should I use `lock` with standard collections or should I remove netstandard1.0 from csproj and use `System.Collections.Concurrent`?

nblumhardt:
Dropping `netstandard1.0` would need to wait for a 3.0 release, but it's possible to conditionally include one implementation or another depending on the target platform.

Dictionaries with locks are often faster than ConcurrentDictionary if the lock isn't contended IIRC, which is generally the case at the formatting/output end of Serilog. The logger/input end is often more highly-contended. Would be interesting to check this out via some benches.

sungam3r:
https://github.com/serilog/serilog/pull/1772#issuecomment-1332819952

sungam3r:
To summarize initial post:

- Use foreach over _outputTemplate.TokenArray instead of Tokens IEnumerator in MessageTemplateTextFormatter.Format method so no IEnumerator is allocated. #1770 
[...]

PR #1771, opened 2022-11-30
Added ReusableStringWriter

igor84:


nblumhardt:
Thanks for taking a shot at this! Since it adds some additional code/risk, comparative benchmarks are the most important step forward from here.

src/Serilog/ReusableStringWriter.cs
```diff
@@ -0,0 +1,47 @@
...
+public class ReusableStringWriter: StringWriter
+{
+    static readonly object _poolLock = new();
+    static readonly Dictionary<IFormatProvider, Stack<ReusableStringWriter>> _writerPools = new();
+
+    /// <summary>
+    /// Gets already created StringWriter if there is one available or creates a new one.
+    /// </summary>
+    /// <param name="formatProvider"></param>
+    public static StringWriter GetOrCreate(IFormatProvider? formatProvider = null)
```
nblumhardt, line 15: There's a small risk here that user code could trigger a space leak by supplying unique format providers - would need to consider mitigating this.

src/Serilog/ReusableStringWriter.cs
```diff
@@ -0,0 +1,47 @@
+﻿namespace Serilog;
+
+/// <summary>
+/// Class that provides reusable StringWriters to reduce memory allocations
+/// </summary>
+public class ReusableStringWriter: StringWriter
+{
+    static readonly object _poolLock = new();
```
nblumhardt, line 8: "Pooling" only a single instance in a `[ThreadStatic]` field might get the same speedup in most cases while avoiding synchronization and the possibility of leaks/annoying pool growth.

Something like a last-in-wins replacement of the cached instance, and then a format provider equality check on retrieval, might deal with the varied-format-provider case reasonably well.

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
```diff
@@ -118,7 +118,10 @@ public void Format(LogEvent logEvent, TextWriter output)
                 }
 
                 if (pt.Alignment.HasValue)
+                {
                     Padding.Apply(output, ((StringWriter)writer).ToString(), pt.Alignment);
+                    writer.Dispose();
```
sungam3r, line 123: Why did you place `Dispose` here?

src/Serilog/ReusableStringWriter.cs
sungam3r: The first think I would do - do not use pooling when formatProvider is not null.

src/Serilog/ReusableStringWriter.cs
sungam3r: 1. https://www.nuget.org/packages/Microsoft.Extensions.ObjectPool/
or
2. non-blcoking last-in-wins in-place implementation like [this one](https://github.com/graphql-dotnet/graphql-dotnet/blob/831efff7380f417132f8cdb95accaca4a08994ab/src/GraphQL/Validation/DocumentValidator.cs#L58)

sungam3r: Locks+dictionaries+stacks will kill performance.

igor84: I went with last-in-wins approach with ThreadStatic.

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
igor84: So that writer would get back into a pool and be reused.

src/Serilog/Rendering/MessageTemplateRenderer.cs
```diff
@@ -68,7 +68,7 @@ public static void RenderPropertyToken(PropertyToken pt, IReadOnlyDictionary<str
             return;
         }
 
-        var valueOutput = new StringWriter();
+        using var valueOutput = ReusableStringWriter.GetOrCreate();
         RenderValue(propertyValue, isLiteral, isJson, valueOutput, pt.Format, formatProvider);
         var value = valueOutput.ToString();
```
sungam3r, line 73: Note for subsequent PR - avoid such `ToString()` calls by using `StringBuilder` from `valueOutput` + add `Padding.Apply` overload that accepts `StringBuilder`.

src/Serilog/ReusableStringWriter.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace Serilog;
+
+/// <summary>
+/// Class that provides reusable StringWriters to reduce memory allocations
+/// </summary>
+public class ReusableStringWriter: StringWriter
```
sungam3r, line 6:
```suggestion
internal class ReusableStringWriter: StringWriter
```

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
sungam3r: I would move `Dispose` call from inside here and place after all conditions:
```cs
if (writer is ReusableStringWriter)
    writer.Dispose();
```

sungam3r: OK

nblumhardt:
Looks good; any chance of some comparative benchmarks?

igor84:
When I run Serilog.PerformanceTests.Harness.Allocations I get the same results with and without these changes. Could it be that Benchmark.NET is isolating runs in such a way that each one has an empty pool in that ThreadStatic that it then initialized and doesn't get to reuse it?

sungam3r:
@igor84 Would you continue working on that PR?

igor84:
> @igor84 Would you continue working on that PR? 

I can rebase it on the latest master later today. Is there anything else that needs to be done?

sungam3r:
I would add benchmarks results like here https://github.com/serilog/serilog/pull/1779#issuecomment-1403656707.

sungam3r:
`EmitLogEvent` - just an example. I don't know if changed code is executed on that path. First make sure that you call changed code. There are others tests or you can write your own special test to show improvements most fully.

igor84:
I added a new test to AllocationsBenchmark. It shows these results:

BEFORE:
|        Method |      Mean |     Error |    StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|-------------- |----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
|      LogEmpty |  5.545 ns | 0.0348 ns | 0.0272 ns |  1.00 |    0.00 |      - |         - |          NA |
| FormatMessage | 76.211 ns | 1.1442 ns | 1.0143 ns | 13.73 |    0.24 | 0.0229 |     192 B |          NA |

AFTER:
|        Method |      Mean |     Error |    StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|-------------- |----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
|      LogEmpty |  5.666 ns | 0.0847 ns | 0.0792 ns |  1.00 |    0.00 |      - |         - |          NA |
| FormatMessage | 73.142 ns | 0.1692 ns | 0.1413 ns | 12.92 |    0.18 | 0.0038 |      32 B |          NA |

src/Serilog/ReusableStringWriter.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// </summary>
+    /// <param name="formatProvider"></param>
+    public static StringWriter GetOrCreate(IFormatProvider? formatProvider = null)
+    {
+        var fmtProvider = formatProvider ?? CultureInfo.CurrentCulture;
+        var writer = _pooledWriter;
+        _pooledWriter = null;
+        if (writer == null || !Equals(writer.FormatProvider, fmtProvider))
+        {
+            writer = formatProvider == null ? new ReusableStringWriter() : new ReusableStringWriter(formatProvider);
```
sungam3r, line 22:
```suggestion
            writer = new ReusableStringWriter(fmtProvider);
```

src/Serilog/ReusableStringWriter.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        return writer;
+    }
+
+    ReusableStringWriter()
+    {
+    }
```
sungam3r, line 30:
```suggestion
```

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
sungam3r: Well... I thought about it again. Look - it's very important to call `Dispose`. Otherwise you will lose `ReusableStringWriter._pooledWriter` **forever**. So what I would really like to see here - a guaranteed `Dispose` call.

sungam3r: 🤦‍♂️ I was wrong. In case of exception you will lose `ReusableStringWriter._pooledWriter` until the next call to `ReusableStringWriter.GetOrCreate`.

sungam3r (approved):

nblumhardt:
Looks like it came out nicely 👍

Saibamen:
Related: https://github.com/serilog/serilog/issues/1769

PR #1774, opened 2022-12-02
Introduce ScalarValue.Null

sungam3r:
Saves some allocations.

### DIFF
--- a/src/Serilog/Events/ScalarValue.cs
+++ b/src/Serilog/Events/ScalarValue.cs
@@ -20,6 +20,11 @@ namespace Serilog.Events;
 public class ScalarValue : LogEventPropertyValue
 {
     /// <summary>
+    /// Scalar value representing <see langword="null"/>.
+    /// </summary>
+    public static readonly ScalarValue Null = new(null);
+
+    /// <summary>
     /// Construct a <see cref="ScalarValue"/> with the specified
     /// value.
     /// </summary>
--- a/src/Serilog/Events/ScalarValue.cs
+++ b/src/Serilog/Events/ScalarValue.cs
@@ -22,7 +22,7 @@ public class ScalarValue : LogEventPropertyValue
     /// <summary>
     /// Scalar value representing <see langword="null"/>.
     /// </summary>
-    public static ScalarValue Null { get; }= new(null);
+    public static ScalarValue Null { get; } = new(null);
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/Serilog/Events/ScalarValue.cs
```diff
@@ -19,6 +19,11 @@ namespace Serilog.Events;
 /// </summary>
 public class ScalarValue : LogEventPropertyValue
 {
+    /// <summary>
+    /// Scalar value representing <see langword="null"/>.
+    /// </summary>
+    public static readonly ScalarValue Null = new(null);
```
sungam3r, line 25: or 
```suggestion
    public static ScalarValue Null { get; }= new(null);
```

nblumhardt (approved):
Nice! Seems like a low-risk win 👍 

My only thought was whether there'd be any advantage to keeping the static property entirely internal, but seems reasonable to expose it and encourage similar usage in other places.

`true` and `false` might also be good candidates for this treatment.

sungam3r:
I can add cached common values in another pr.

src/Serilog/Events/ScalarValue.cs
```diff
@@ -19,6 +19,11 @@ namespace Serilog.Events;
 /// </summary>
 public class ScalarValue : LogEventPropertyValue
 {
+    /// <summary>
+    /// Scalar value representing <see langword="null"/>.
+    /// </summary>
+    public static ScalarValue Null { get; }= new(null);
```
augustoproiete, line 25:
```suggestion
    public static ScalarValue Null { get; } = new(null);
```

🛃 nitpick - missing space

augustoproiete (approved):
LGTM

test/Serilog.ApprovalTests/Serilog.approved.txt
```diff
@@ -409,6 +410,7 @@ namespace Serilog.Events
     {
         public SequenceValue(System.Collections.Generic.IEnumerable<Serilog.Events.LogEventPropertyValue> elements) { }
         public System.Collections.Generic.IReadOnlyList<Serilog.Events.LogEventPropertyValue> Elements { get; }
+        public static Serilog.Events.SequenceValue Empty { get; }
```
sungam3r, line 413: @nblumhardt I pulled changes from `dev` to fix approval file after merging #1785.

SimonCropp (approved):

PR #1778, opened 2022-12-03
Tokens -> TokenArray

sungam3r:
What do you think about exposing this property in public API? The same problem as in #1769 for Serilog.Sinks.Console: `foreach (var token in template.Tokens) ` usages in `OutputTemplateRenderer`, `PropertiesTokenRenderer` and `ThemedMessageTemplateRenderer`.


SimonCropp (approved):

nblumhardt:
Thanks @sungam3r! RE exposing the array publicly, arrays are mutable so this would break some assumptions that I think can currently be made about the type (once a `MessageTemplate` has been constructed it won't change).

PR #1780, opened 2022-12-04
Use ArrayPool for EventProperty arrays

sungam3r:
No changes in public API.


src/Serilog/Capturing/PropertyBinder.cs
```diff
@@ -83,11 +92,26 @@ EventProperty[] ConstructPositionalProperties(MessageTemplate template, object?[
...
     }
 
-    EventProperty[] ConstructNamedProperties(MessageTemplate template, object[] messageTemplateParameters)
+#if NETSTANDARD2_1 || NET5_0_OR_GREATER
+    private readonly static System.Buffers.ArrayPool<EventProperty> _eventPropertyPool = System.Buffers.ArrayPool<EventProperty>.Shared;
+#endif
+
+    internal static void Return(EventProperty[] array)
+    {
+#if NETSTANDARD2_1 || NET5_0_OR_GREATER
```
SimonCropp, line 101: i know @nblumhardt refers named constants to represent features. so perhaps add a 

`FEATURE_ARRAYPOOL`

SimonCropp (approved):

nblumhardt:
I think we should reconsider this one, unless the benchmark differences are stark. Adam Sitnik's investigation suggests that `new` + GC will be faster for small arrays, as I'd expect most of these will be.

https://adamsitnik.com/Array-Pool/#the-results

We'd also need to carefully consider where exceptions may be thrown between the rent and return phases, since we'll risk leaking space in the pool otherwise. Adding `finally` blocks would mitigate this but introduces more noise and complexity that the benchmark improvements would need to counterbalance.

sungam3r:
Conflicts resolved.

`dotnet test -c Release -f net7.0 --filter "FullyQualifiedName=Serilog.PerformanceTests.Harness.Pipeline"`

``` ini

BenchmarkDotNet=v0.13.2, OS=Windows 11 (10.0.22621.1105)
.NET SDK=7.0.102
  [Host]     : .NET 7.0.2 (7.0.222.60605), X64 RyuJIT AVX2
  DefaultJob : .NET 7.0.2 (7.0.222.60605), X64 RyuJIT AVX2


```
before

| EmitLogEvent | 160.5 ns | 3.03 ns | 5.99 ns | 0.0448 |     376 B |

after

| EmitLogEvent | 185.0 ns | 3.62 ns | 3.71 ns | 0.0401 |     336 B |

sungam3r:
Log line changed to ` _log.Information(_exception, "Hello, {Name} {Name2} {Name3} {Name4}!", "World", "World2", "World3", "World4");`:

before

| EmitLogEvent | 262.5 ns | 5.20 ns | 5.34 ns | 0.0753 |     632 B |

after 

| EmitLogEvent | 258.5 ns | 2.47 ns | 2.31 ns | 0.0648 |     544 B |

sungam3r:
1. 15% slower, 10% less memory
2. 1,5% slower, 14% less memory

Of course, experience can vary.

sungam3r:
Honestly, I would prefer to refactor the way Serilog works with `EventProperty` and other internal/public stuff but it's not just an optimization, it also requires changes in public APIs. The effect can be an avalanche-like. 

In this regard Serilog is somewhat similar to Newtonsoft.Json (compared to STJ). It provides rich possibilities and object model at the cost of significant memory consumption for the objects created.

nblumhardt:
A couple of years ago I spent some time working on a prototype for a version of Serilog based on struct log events and a matching "middleware"-ish style of pipeline using ref parameters. To avoid forking the ecosystem, the idea was to support both pipeline styles, but only fall back to allocating `LogEvent` objects for the old pipeline when previous-version components (sinks and so on) were present.

On the plus side, I think this could work, but on the minus, it involves quite a lot of machinery and corresponding development effort. Could still be something to explore, though! :-)

On this PR, I think we're better off leaving things as-is - the exception-safety problems around array pooling outweigh the benefits of lower allocation.

nblumhardt:
Looping back - I was wrong about exception safety being a worry, here - it's fine to leak rented arrays. I'll zip back over, fix the conflict, and merge this when I have a chance 👍

sungam3r:
👍 Glad to hear it but... now I doubt whether it really needs to be done 😄. As I said previously
> Honestly, I would prefer to refactor the way Serilog works with EventProperty and other internal/public stuff but it's not just an optimization, it also requires changes in public APIs. 

If you are sure with suggested changes here then OK, I'm fine merging it and can resolve conflict on my own. Ping me back for your decision.

sungam3r:
Conflicts resolved.

nblumhardt:
Thanks for the reply; sorry about the very slow turnaround at the moment, currently very deep in another project so limited cycles to spare for a week or two more.

I might dig a bit more into performance as soon as I have a chance 👍

nblumhardt:
I think this could be one to skip, given that small arrays will make up a disproportionate amount of events in typical apps.

@sungam3r I'll close this since I have a day to run over and triage the open PRs - trying to get back to a nice clean tracker. Feel free to ping/reopen if you think we should still be exploring this one. Thanks for all the effort on it, though!

sungam3r:
I'm fine closing this one.

PR #1782, opened 2022-12-04
Avoid StringWriter.ToString() calls

sungam3r:


src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
```diff
@@ -118,7 +118,13 @@ public void Format(LogEvent logEvent, TextWriter output)
                 }
 
                 if (pt.Alignment.HasValue)
+                {
+#if NETCOREAPP3_0_OR_GREATER
```
nblumhardt, line 122: Rather than `NETCOREAPP3_0_OR_GREATER` can we define an explanatory constant like `FEATURE_WRITE_STRINGBUILDER`? Existing ones are: https://github.com/serilog/serilog/blob/dev/src/Serilog/Serilog.csproj#L44

sungam3r: done

src/Serilog/Formatting/Display/MessageTemplateTextFormatter.cs
```diff
@@ -118,7 +118,13 @@ public void Format(LogEvent logEvent, TextWriter output)
                 }
 
                 if (pt.Alignment.HasValue)
+                {
+#if FEATURE_WRITE_STRINGBUILDER
+                    Padding.Apply(output, ((StringWriter)writer).GetStringBuilder(), pt.Alignment);
```
SimonCropp, line 123: nit: but i prefer to minimise whats inside the `#if`. so i would assign `(StringWriter)writer` to a variable outside the `#if`

SimonCropp (approved):

SimonCropp:
@nblumhardt LGTM

PR #1799, opened 2023-01-24
update sdk to 7.0.100

SimonCropp:


SimonCropp:
think this will need a different approach

PR #1805, opened 2023-01-24
remove OutputProperties.GetOutputProperties

SimonCropp:


src/Serilog/Events/MessageTemplate.cs
```diff
@@ -21,11 +21,6 @@ namespace Serilog.Events;
 /// </summary>
 public class MessageTemplate
 {
-    /// <summary>
-    /// Represents the empty message template.
-    /// </summary>
-    public static MessageTemplate Empty { get; } = new(Enumerable.Empty<MessageTemplateToken>());
```
nblumhardt, line 27: This one's a public API; I don't think we have a good reason to remove it right now.

SimonCropp: @nblumhardt just to play devils advocate... what is the use case?

PR #1784, opened 2022-12-04
NET 7 SDK

sungam3r:
Follow up #1766

PR #1806, opened 2023-01-24
Attempt 4 at SDK 7

SimonCropp:
* installs SDK in appveyor.yml
 * targets net7 in serilog.csproj
 * tests net7
 * remove sdk list in build scrips, since we are installing it explicitly


SimonCropp:
@sungam3r @nblumhardt this one is good to go, and replaces https://github.com/serilog/serilog/pull/1784

src/Serilog/Serilog.csproj
```diff
@@ -3,7 +3,7 @@
     <Description>Simple .NET logging with fully-structured events</Description>
     <VersionPrefix>3.0.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.3;netstandard1.0;net45;net46;net47;net5.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.3;netstandard1.0;net45;net46;net47;net5.0;net6.0;net7.0</TargetFrameworks>
```
sungam3r, line 6: There are no NET7 ifdefs (yet) so targeting net7 is generally useless (except some compiler optimizations that do not change things much IMO).

nblumhardt: 👍  I'd leave `net7.0` everywhere else, but for core Serilog we might be better off just targeting LTS frameworks unless we can take advantage of new features. What do you think @SimonCropp ?

SimonCropp: @sungam3r you are correct. there is no value ATM targeting net7.0. i have gotten into the habit, of always targeting the latest, just in case any nullable attributes are added/changed.

@nblumhardt the only down side if a *very* small increase in build time and package size

SimonCropp:
conflict resolved

src/Serilog/Serilog.csproj
nblumhardt: Fair enough, I'm not greatly concerned either way, let's roll with 7 here now and see how it pans out.

PR #1813, opened 2023-01-25
improve GetPropertiesRecursive

SimonCropp:
* removes a nested loop
 * removes a lambda allocation
 * remove the use of typeinfo. we no longer need it. means we can use bindingflags (which DeclaredProperties uses internally anyway) but can specify flags we want and remove some extra checks
 * use FlattenHierarchy to avoid the basetype hierarchy walk


sungam3r:
You can swith return type to `HashSet<PropertyInfo>` from `IEnumerable<PropertyInfo>` to avoid iterator allocation as well.

PR #1825, opened 2023-01-26
use is Enum for enum check in EnumScalarConversionPolicy

SimonCropp:
IMO more readable, and makes it consistent with the type checking in other policies


skomis-mm (approved):

augustoproiete (approved):
LGTM

PR #1826, opened 2023-01-26
use Dictionary instead of Hashtable in MessageTemplateCache

SimonCropp:


src/Serilog/Core/Pipeline/MessageTemplateCache.cs
```diff
@@ -36,9 +36,10 @@ public MessageTemplate Parse(string messageTemplate)
 
         // ReSharper disable once InconsistentlySynchronizedField
         // ignored warning because this is by design
-        var result = (MessageTemplate?)_templates[messageTemplate];
-        if (result != null)
+        if (_templates.TryGetValue(messageTemplate, out var result))
+        {
```
skomis-mm, line 40: Thread safety guaranties for two data structures are different:
[Dictionary<,>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-7.0#thread-safety):

> A [Dictionary<TKey,TValue>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-7.0) can support multiple readers concurrently, as long as the collection is not modified.

[Hashtable](https://learn.microsoft.com/en-us/dotnet/api/system.collections.hashtable?view=net-7.0#thread-safety):
> It is thread safe for multi-thread use when only **one of the threads perform write** (update)

That was the main motivation to use `Hashtable`, otherwise read operation should have been placed into the `lock` section.

nblumhardt:
@SimonCropp no great worry, but would it be possible to send PRs from your own fork rather than branches in this repo? With a few contributors hacking at things it gets a bit hectic to figure out what's what :-)

src/Serilog/Core/Pipeline/MessageTemplateCache.cs
SimonCropp: @skomis-mm i did not know that. thanks so much for the info

SimonCropp: @nblumhardt do u want a PR with a comment containing the above info?

SimonCropp:
@nblumhardt happy to. although i prefer branches in the main repo since it is easier to collaborate on PRs. i find the branch noise is mitigated by having "auto delete merged branches" enabled for the project

src/Serilog/Core/Pipeline/MessageTemplateCache.cs
nblumhardt: Awesome idea, thanks :-)

nblumhardt:
:+1: let's keep the current approach for the time being; "allow edits from maintainers" works to enable roughly the same collaboration options.

For longer-lived subprojects, creating branches here is fair game though too.

src/Serilog/Core/Pipeline/MessageTemplateCache.cs
SimonCropp: @nblumhardt  https://github.com/serilog/serilog/pull/1828

PR #1830, opened 2023-01-27
Leverage dictionary TryAdd and items constructor

SimonCropp:
* try add. leverages a custom path since the internal implementation combines existence check and the insert. so avoids one of the two lookups. and eventually less code when we burn only target runtimes
 * use the items constructor. less code and leverages an internal `AddRange...`
 
 ```
 // It is likely that the passed-in dictionary is Dictionary<TKey,TValue>. When this is the case,
// avoid the enumerator allocation and overhead by looping through the entries array directly.
// We only do this when dictionary is Dictionary<TKey,TValue> and not a subclass, to maintain
// back-compat with subclasses that may have overridden the enumerator behavior.
```

it effectively does an array copy of the internal dictionary entries. this should speed up the hot path since the copy is done in `Emit(LogEvent)`

also avoids enumeration and the value lookup in the source 

```
foreach (var key in _properties.Keys)
            properties.Add(key, _properties[key]);
```

### CODE REVIEW

**Comment by SimonCropp** [comment]
@sungam3r thoughts?

**Review by sungam3r — APPROVED** [review]
Just run `dotnet test -c Release -f net7.0 --filter "FullyQualifiedName=Serilog.PerformanceTests.Harness.Pipeline"` before and after changes to verify intents. Optimization is nothing without speedup confirmation.

**Comment by sungam3r** [comment]
Also you may have to copy assets folder here 
изображение
if you have issue with signing in BenchmarksDotNet.

**Comment by SimonCropp** [comment]
@sungam3r i get

```
No test matches the given testcase filter `FullyQualifiedName=Serilog.PerformanceTests.Harness.Pipeline` in C:\Code\serilog\serilog\test\Serilog.ApprovalTests\bin\Release\net7.0\Serilog.ApprovalTests.dll
```

SimonCropp:
@sungam3r thanks

profiling just the copy 


before

| EmitLogEvent | 382.7 ns | 0.46 ns | 0.38 ns | 0.0601 |     504 B |


after

| EmitLogEvent | 128.5 ns | 0.66 ns | 0.52 ns | 0.0601 |     504 B |

Issue #1226, opened 2018-10-02
Consider taking inheritance into account in Destructure.ByTransforming


ArXen42:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [ ] Bug
- [x] New Feature

From `Serilog.Configuration.LoggerDestructuringConfiguration.ByTransforming<TValue>` ([link](https://github.com/serilog/serilog/blob/5b0569ac42363dffced9854b0406ca338248f5fc/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs#L121)):
``` csharp
var policy = new ProjectedDestructuringPolicy(t => t == typeof(TValue), // <-- type comparison
      o => transformation((TValue)o));
```

Such behaviour does not allow using `ByTransforming` with interfaces or base classes. For example, it won't work with Entity Framework models.

Replacing `t == typeof(TValue)` with something like `t.IsAssignableFrom(typeof(TValue))` would remove the need to write custom policies for described cases.

Introducing optional boolean parameter to enable new behaviour can preserve backward compatibility.

nblumhardt:
Hi! Thanks for the suggestion; I think you can achieve what you're after using `Destructure.ByTransformingWhere(...)` - does that help?

ArXen42:
I'm blind, thanks. Didn't think about using this method with desired predicate.
Seems like there is no reason to change `ByTransforming` behaviour, then.

PR #1833, opened 2023-01-27
add some type safety to ProjectedDestructuringPolicy

SimonCropp:
and adds support for (fixes a bug) for inherited types when using ByTransforming


skomis-mm:
It will break existing loggers. Consider you have some object inherited from `T`, but have additional properties that can throw an exception (e.g. `HttpRequest`). 

Exact matching for `T` is by design for `ByTransforming`. We had several discussions/issues on this: #1226 , #1512 and some others

SimonCropp:
@skomis-mm thanks for the context

PR #1842, opened 2023-01-30
Use correct overload with char in JsonFormatter

sungam3r:
Also get rid of `string.Format` in `WriteRenderings` and `WriteProperties`.


SimonCropp (approved):

SimonCropp:
are there no tests for this code path?

sungam3r:
I don't know. I just reviewed changes from #1839 .

PR #1863, opened 2023-02-25
move from net461 to net462

SimonCropp:
also unlocks updating xunit.runner.visualstudio since it dropped net461


sungam3r (approved):

augustoproiete (approved):
LGTM

PR #1865, opened 2023-02-28
simplify build scripts

SimonCropp:
* fixes on mac and linux that we were missing test runs on net7 and net6
 * removes redundant restore = faster CI
 * remove redundant `dotnet --info`. we already get that from the dotnet install script in appveyor.yml
 * avoids extra build+restore that happen when u test if u omit  `--no-build --no-restore` = faster CI
 * less code

### DIFF
--- a/Build.ps1
+++ b/Build.ps1
@@ -30,6 +30,6 @@ if($LASTEXITCODE -ne 0) { exit 2 }
 
 Write-Output "build: Testing"
 
-& dotnet test --configuration Release --no-build --no-restore
+& dotnet test  test\Serilog.Tests --configuration Release --no-build --no-restore
 
 if($LASTEXITCODE -ne 0) { exit 3 }
--- a/Build.ps1
+++ b/Build.ps1
@@ -1,15 +1,11 @@
 Write-Output "build: Build started"
 
-& dotnet --info
-
 Push-Location $PSScriptRoot
 
 if(Test-Path .\artifacts) {
     Write-Output "build: Cleaning .\artifacts"
     Remove-Item .\artifacts -Force -Recurse
 }
-
-& dotnet restore --no-cache
 
 $branch = @{ $true = $env:APPVEYOR_REPO_BRANCH; $false = $(git symbolic-ref --short -q HEAD) }[$NULL -ne $env:APPVEYOR_REPO_BRANCH];
 $revision = @{ $true = "{0:00000}" -f [convert]::ToInt32("0" + $env:APPVEYOR_BUILD_NUMBER, 10); $false = "local" }[$NULL -ne $env:APPVEYOR_BUILD_NUMBER];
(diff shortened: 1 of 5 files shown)


sungam3r:
👍  But the main question from me so far - why to continue using appveyor at all? GitHub actions CI seems more simple and powerful.

SimonCropp:
@sungam3r 

>  GitHub actions CI seems more simple and powerful.

in what way. the build scripts serilog uses are doing custom versions. the same script would need to be written in GH actions

sungam3r:
Of course any scripting framework can doing custom versions. Another question is - is it necessary to do such versioning at all? You can use MinVer or something like that, which will versioning builds without any manual scripting at all.

```
$branch = @{ $true = $env:APPVEYOR_REPO_BRANCH; $false = $(git symbolic-ref --short -q HEAD) }[$NULL -ne $env:APPVEYOR_REPO_BRANCH];
$revision = @{ $true = "{0:00000}" -f [convert]::ToInt32("0" + $env:APPVEYOR_BUILD_NUMBER, 10); $false = "local" }[$NULL -ne $env:APPVEYOR_BUILD_NUMBER];
$suffix = @{ $true = ""; $false = "$($branch.Substring(0, [math]::Min(10,$branch.Length)))-$revision"}[$branch -eq "main" -and $revision -ne "local"]
$commitHash = $(git rev-parse --short HEAD)
$buildSuffix = @{ $true = "$($suffix)-$($commitHash)"; $false = "$($branch)-$($commitHash)" }[$suffix -ne ""]
```
If you mean this, then I would first wonder if it is worth deleting all this and use more modern/simpler tools. MinVer, for example, does the job with only single nuget dependency and works for any branch.

In addition, it seems to me to be natural to believe that the GitHub Actions ecosystem is obviously better suited (than any external system) for the code placed on GitHub itself. Development, reviewing and CI/CD occur in one space so there is no need to jump over external site to dive into build/test logs.

I honestly say that I do not know appveyor so much as to expertly discuss its advantages and disadvantages (compared to other CI/CD systems), but what I can say for sure is that I saw a number of migrations from appveyor to GitHub Actions last years. I did not see the reverse transitions.

sungam3r:
I may be wrong but also it seems that appveyor provides more limited resources for builds performing them rather slowly and putting in the queue.

SimonCropp:
> If you mean this, then I would first wonder if it is worth deleting all this and use more modern/simpler tools. MinVer, for example, does the job with only single nuget dependency and works for any branch.

agreed. and i was planning on suggesting that soon. but i am trying to do incremental changes.
but that decision is not related to AppVeyor/GHActions since the same complexity exist on both

> In addition, it seems to me to be natural to believe that the GitHub Actions ecosystem is obviously better suited (than any external system) for the code placed on GitHub itself. Development, reviewing and CI/CD occur in one space so there is no need to jump over external site to dive into build/test logs.

I actually find AppVeyor simpler to navigate (and specifically to find tests) than GHActions. Take this build for example https://github.com/bUnit-dev/bUnit/actions/runs/4290650005, or the specific job https://github.com/bUnit-dev/bUnit/actions/runs/4290650005/jobs/7474883650, i cant see how to find the test results. 
compared to AppVeyor where it is clearly visible for the equivalent https://ci.appveyor.com/project/serilog/serilog/builds/46361629 https://ci.appveyor.com/project/serilog/serilog/builds/46361629/job/iae48v8rq5pcy37w/tests

> I may be wrong but also it seems that appveyor provides more limited resources for builds performing them rather slowly and putting in the queue.

yep agreed. GHActions is better at this. in AppVeyor you need to pay. 25$ a month for an additional concurrent job

I have ~120 projects that use AppVeyor for builds, and most of those also use GH actions for a few other small tasks (like building documentation). I also work on ~10 projects that use GH actions for builds. IMO AppVeyor is still better for build+test runs. I suspect that will eventually change given GH actions are consistently evolving, but they are not there yet.

Again just my preference. If @nblumhardt wants to move to GHActions i am happy to help with the transition

sungam3r:
I agree that AppVeyor has nicer tests visualization.

sungam3r:
https://www.reddit.com/r/dotnet/comments/mul8po/github_action_to_visualize_test_results_in_github/

nblumhardt:
No strong opinion around Actions vs AppVeyor here, but at this point in time I don't think a switch would achieve much other than churn.

PR #1872, opened 2023-03-04
Optimize AddPropertyIfAbsent

sungam3r:
No actual API changes, only a few fixes for NRT annotations.
Eliminates allocation of `LogEventProperty` wrapper for `LogEventPropertyValue`.

`dotnet test -c Release -f net7.0 --filter "FullyQualifiedName=Serilog.PerformanceTests.Harness.LogContextEnrichment"`

``` ini

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22621.1265/22H2/2022Update/SunValley2)
.NET SDK=7.0.201
  [Host]     : .NET 7.0.3 (7.0.323.6910), X64 RyuJIT AVX2
  DefaultJob : .NET 7.0.3 (7.0.323.6910), X64 RyuJIT AVX2


```

before 

|               Method |       Mean |     Error |    StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|--------------------- |-----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
|                 Bare |   6.864 ns | 0.1439 ns | 0.1346 ns |  1.00 |    0.00 |      - |         - |          NA |
|         PushProperty |  61.231 ns | 1.2249 ns | 1.3107 ns |  8.94 |    0.28 | 0.0296 |     248 B |          NA |
|   PushPropertyNested | 118.867 ns | 2.4084 ns | 2.6769 ns | 17.34 |    0.62 | 0.0591 |     496 B |          NA |
| PushPropertyEnriched | 106.414 ns | 1.7542 ns | 1.6409 ns | 15.51 |    0.36 | 0.0362 |     304 B |          NA |

after

|               Method |       Mean |     Error |    StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|--------------------- |-----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
|                 Bare |   6.810 ns | 0.0722 ns | 0.0603 ns |  1.00 |    0.00 |      - |         - |          NA |
|         PushProperty |  62.395 ns | 1.2570 ns | 2.5393 ns |  9.52 |    0.26 | 0.0296 |     248 B |          NA |
|   PushPropertyNested | 128.455 ns | 2.5705 ns | 4.6351 ns | 19.13 |    0.87 | 0.0591 |     496 B |          NA |
| PushPropertyEnriched |  93.311 ns | 1.8405 ns | 2.1909 ns | 13.74 |    0.41 | 0.0296 |     248 B |          NA |

See `Allocated` column - 18% less heap allocation for case when event is logged with some pushed property.


src/Serilog/Capturing/MessageTemplateProcessor.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Serilog.Capturing;
 
-class MessageTemplateProcessor : ILogEventPropertyFactory
+class MessageTemplateProcessor : ILogEventPropertyFactory, ILogEventPropertyValueFactory
```
sungam3r, line 17: I think this interface was just missed.

src/Serilog/Events/LogEvent.cs
```diff
@@ -192,3 +194,18 @@ internal LogEvent Copy()
             properties);
     }
 }
+
+internal static class LogEventExtensions
```
sungam3r, line 198: @nblumhardt For now it's `internal` but it may be changed to public for use in dependent packages. For example, in serilog-extensions-logging, see `EnrichAndCreateScopeItem`: 

```
var property = propertyFactory.CreateProperty(key, value, destructureObject);
logEvent.AddPropertyIfAbsent(property);
```

src/Serilog/Core/ILogEventPropertyValueFactory.cs
```diff
@@ -15,5 +15,5 @@ public interface ILogEventPropertyValueFactory
     /// then the value will be converted to a structure; otherwise, unknown types will
     /// be converted to scalars, which are generally stored as strings.</param>
     /// <returns>The value.</returns>
-    LogEventPropertyValue CreatePropertyValue(object value, bool destructureObjects = false);
+    LogEventPropertyValue CreatePropertyValue(object? value, bool destructureObjects = false);
```
nblumhardt, line 18: Wow, nice find! 👍

src/Serilog/Events/LogEvent.cs
```diff
@@ -192,3 +194,18 @@ internal LogEvent Copy()
...
+{
+    public static void AddPropertyIfAbsent(this LogEvent evt, ILogEventPropertyFactory factory, string name, object? value, bool destructureObjects = false)
```
nblumhardt, line 200: Since we already end up adding an `internal` property to `LogEvent`, we could just skip that and make this overload of `AddPropertyIfAbsent` an internal method on `LogEvent` itself, for now?

I can see how dependent libraries would benefit from making this accessible, but in that case we'd have to expose _something_ on `LogEvent` anyway, so we might as well keep everything tight/encapsulated for now.

sungam3r: done

PR #1873, opened 2023-03-04
Make LogEventProperty struct

sungam3r:
In light of upcoming v3.


SimonCropp:
can we use `record struct`?

nblumhardt:
This would completely (binary) break the entire existing ecosystem, we can't make this change unfortunately.

sungam3r:
Let this PR will still be open for some time. One way or another, more and more changes are gradually taking place for version 3. It is possible that the time will come when the next change will already be so breaking (at the cost of tempting improvements) that it will not make sense to take care of any backward compatibility. In the end, the change of major version is the only moment when such changes are acceptable. If not now, then never.

SimonCropp:
what is the perf impact of this? so the value can be compared to the breaking impact

nblumhardt:
@SimonCropp it's a fair question!

I don't think we'd be giving enough weight to what breaking Serilog's compatibility with existing sinks would mean, though. It could end up completely destroying this project. A bit like the "Apache log4net signing key rotation" but an order of magnitude worse :-)

With this change, all 500+ existing sinks on NuGet would either work **only with v2**, or **only with v3**, meaning you couldn't update your application from v2 to v3 until all of the sinks it relies on have been recompiled. It'd split the Serilog ecosystem into those who'd have to stay on v2, those who were forced to move to v3, and the largest group - the ones stuck in the middle who would port to some other library instead because the combination of sinks they need wouldn't line up with either version (even temporarily).

There are other ways to make this change though, that don't break the ecosystem. I've spiked out pieces of one option over the past couple of years, but it's a lot of work. Essentially, we define a new `LogEventData` struct and pipeline separately, and allow `LoggerConfiguration` to plug older sinks/enrichers etc. into it by converting from `LogEvent` to `LogEventData` on the way in, and back again on the way out.

I was really enthusiastic about this for the perf possibilities it opens up, but then hit the realization that asynchronous/buffered sinks end up doing all the boxing work anyway, and in many cases we end up allocating more than we would have done in the current design.

For example, imagine you have a file sink behind `Serilog.Sinks.Async`, and a Seq (or some other network) sink at the same time. Both of these need to maintain a heap-based queue of events. Instead of each needing a reference to the bulky `LogEvent` class instance that's already provided by Serilog, they both end up with allocations for the struct version's fields, plus a lot more copying etc. on the way in.

I ended up at the conclusion that while we'd do better on synthetic benchmarks, it's uncertain that would translate to actual improvements in end-to-end performance for real apps, and might end up making performance worse in a lot of cases.

It'd be cool to explore again, though, and there could always be tactics/nuances that I missed :-)

sungam3r:
Close?

nblumhardt:
👍 thanks again @sungam3r.

PR #1875, opened 2023-03-04
Cache common scalar values

### LINKED ISSUE #1777: Cache common scalar values
Nice! Seems like a low-risk win 👍 

My only thought was whether there'd be any advantage to keeping the static property entirely internal, but seems reasonable to expose it and encourage similar usage in other places.

`true` and `false` might also be good candidates for this treatment.

_Originally posted by @nblumhardt in https://github.com/serilog/serilog/pull/1774#pullrequestreview-1203370943_

sungam3r:
fixes #1777

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -135,8 +135,18 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
             }
         }
 
-        if (value is string)
-            return new ScalarValue(value);
+        if (value is string s)
+            return s == string.Empty ? ScalarValue.EmptyString : new ScalarValue(value);
+
+        if (value is bool b)
+            return b ? ScalarValue.True : ScalarValue.False;
+
+        if (value is int i)
+        {
+            if (i == 0) return ScalarValue.Zero;
+            if (i == 1) return ScalarValue.One;
+            if (i == -1) return ScalarValue.MinusOne;
+        }
 
         foreach (var scalarConversionPolicy in _scalarConversionPolicies)
         {
@@ -165,7 +175,15 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
         if (TryConvertCompilerGeneratedType(value, type, destructuring, out var compilerGeneratedResult))
             return compilerGeneratedResult;
 
-        return new ScalarValue(value.ToString() ?? "");
+        var stringified = value.ToString();
+
+        if (stringified == null)
+            return ScalarValue.EmptyString; // ScalarValue.Null; ?
+
+        if (stringified == string.Empty)
+            return ScalarValue.EmptyString;
+
+        return new ScalarValue(stringified);
     }
 
     bool TryConvertEnumerable(object value, Type type, Destructuring destructuring, [NotNullWhen(true)] out LogEventPropertyValue? result)
--- a/src/Serilog/Events/ScalarValue.cs
+++ b/src/Serilog/Events/ScalarValue.cs
@@ -25,6 +25,36 @@ public class ScalarValue : LogEventPropertyValue
     public static ScalarValue Null { get; } = new(null);
 
     /// <summary>
+    /// Scalar value representing empty string.
+    /// </summary>
+    public static ScalarValue EmptyString { get; } = new(string.Empty);
+
+    /// <summary>
+    /// Scalar value representing <see langword="true"/>.
+    /// </summary>
+    public static ScalarValue True { get; } = new(true);
+
+    /// <summary>
+    /// Scalar value representing <see langword="false"/>.
+    /// </summary>
+    public static ScalarValue False { get; } = new(false);
+
+    /// <summary>
+    /// Scalar value representing 0 integer value.
+    /// </summary>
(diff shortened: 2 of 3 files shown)


src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -165,7 +175,15 @@ LogEventPropertyValue CreatePropertyValue(object? value, Destructuring destructu
         if (TryConvertCompilerGeneratedType(value, type, destructuring, out var compilerGeneratedResult))
             return compilerGeneratedResult;
 
-        return new ScalarValue(value.ToString() ?? "");
+        var stringified = value.ToString();
+
+        if (stringified == null)
+            return ScalarValue.EmptyString; // ScalarValue.Null; ?
```
sungam3r, line 181: 🤔 ? Why so returning empty string? I tried to change it to return `ScalarValue.Null` but some tests failed. I have a suspicion that this was done by design a long time ago and now it can be dangerous to change behavior here. Honestly returning `null` here seems more reasonable.

nblumhardt:
I just ran:

```sql
select
  sum(length(keys(@Properties))) as total,
  count(@Properties[?] = -1) as minus_1,
  count(@Properties[?] = 0) as zero,
  count(@Properties[?] = 1) as plus_1,
  count(@Properties[?] = '') as empty,
  count(@Properties[?] = null) as nul,
  count(@Properties[?] = true) as tru,
  count(@Properties[?] = false) as fals
from stream 
```

over our last week's raw prod log data, and got:

total | minus_1 | zero | plus_1 | empty | nul | tru | fals
-- | -- | -- | -- | -- | -- | -- | --
8486554 | 0 | 116 | 1975 | 0 | 10706 | 614 | 191

The logs are our own app logs (system is ~6 years old?) using ASP.NET Core.

Out of the 8.5M property values we've collected, shockingly, we don't have any -1 or `''` values.

`0`, `true`, and `false` are so uncommon as to be practically absent, while the `1` case is still only around 1/4250 frequency.

Not sure how representative this is, but given the most frequent `null` case is only 10.7K values, perhaps we should reevaluate?

Anyone watching along keen to spin that query up over a few hours (or days) data in Seq and see whether the picture is similar?

sungam3r:
Rebased on dev. What is verdict here?

nblumhardt:
I'd be inclined to skip this one - doesn't seem compelling enough to justify including it.

sungam3r:
Some logs from our production for `IsAuthenticated` boolean property
изображение

As I said - experience may vary. In your case you have no enough logs with booleans, ints, etc., OK.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -165,7 +175,15 @@ LogEventPropertyValue CreatePropertyValue(object? value, Destructuring destructu
...
+
+        if (stringified == string.Empty)
```
celluj34, line 183: A similar check is already happening at line 138. IMO we should move the string checks to the bottom and consolidate to something like `var stringified = value is string ? (string)value : value.ToString()`.

src/Serilog/Capturing/PropertyValueConverter.cs
celluj34: Is this even possible? I suppose some class could override `ToString` method, but that seems silly. As an outsider, returning null seems more logical.

src/Serilog/Capturing/PropertyValueConverter.cs
sungam3r: Similar but not the same. Check on line 138 serves as fast return **before** calling `value.ToString()`.

sungam3r:
rebased on dev

src/Serilog/Capturing/PropertyValueConverter.cs
celluj34: What I meant was, it would be silly for a class to override ToString to return `null`, but I'm sure it's being done somewhere for a reason I am not aware of.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -138,8 +138,18 @@ LogEventPropertyValue CreatePropertyValue(object? value, Destructuring destructu
...
-            return new ScalarValue(value);
+        if (value is string s)
+            return s == string.Empty ? ScalarValue.EmptyString : new ScalarValue(value);
+
+        if (value is bool b)
+            return b ? ScalarValue.True : ScalarValue.False;
+
+        if (value is int i)
+        {
+            if (i == 0) return ScalarValue.Zero;
```
celluj34, line 149: We could use a switch expression here, that would eliminate two comparisons if `i == -1`.

sungam3r: Switch expression should return value for each case.

nblumhardt:
A middle ground might be to add these into `PrimitiveScalarConversionPolicy`, and skip the public constants?

https://github.com/serilog/serilog/blob/dev/src/Serilog/Policies/PrimitiveScalarConversionPolicy.cs#L17

Doing it in there would potentially enable using `switch` to give the compiler an easier task optimizing things.

I'd be inclined to drop `-1`, since it's much rarer than 32-bit int 0 and 1.

PR #1878, opened 2023-03-10
Avoid creating SafeAggregateSink wrapper around empty list

sungam3r:


src/Serilog/Core/Sinks/SafeAggregateSink.cs
```diff
@@ -18,6 +18,8 @@ class SafeAggregateSink : ILogEventSink
 {
     readonly ILogEventSink[] _sinks;
 
+    internal static SafeAggregateSink Empty = new(new ILogEventSink[0]);
```
nblumhardt, line 21: This appears to be unused?

src/Serilog/LoggerConfiguration.cs
```diff
@@ -128,11 +128,17 @@ public Logger CreateLogger()
...
+            sink = new SafeAggregateSink(_logEventSinks);
 
         var auditing = _auditSinks.Any();
         if (auditing)
-            sink = new AggregateSink(new[] { sink }.Concat(_auditSinks));
+        {
+            sink = new AggregateSink(sink == null ? _auditSinks : new[] { sink }.Concat(_auditSinks));
+        }
+
+        sink ??= SafeAggregateSink.Empty;
```
SimonCropp, line 141: @nblumhardt isnt it used here?

src/Serilog/Core/Sinks/SafeAggregateSink.cs
sungam3r: `sink ??= SafeAggregateSink.Empty;` in another file.

src/Serilog/LoggerConfiguration.cs
nblumhardt: Ah, my bad :-) sorry!

nblumhardt:
In the majority of applications, where there are a nonzero number of sinks, wouldn't this just lead to an additional/wasted/long-lived instance of `SafeAggregateSink` and an empty array?

sungam3r:
I removed field.

nblumhardt:
LGTM! 👍

PR #1881, opened 2023-03-26
use System.Memory on legacy runtimes

SimonCropp:
the plan is this will allow us to use span APIs in more places without conditional compilation


SimonCropp:
ping @sungam3r

src/Serilog/Core/LevelOverrideMap.cs
```diff
@@ -58,18 +58,15 @@ public LevelOverride(string context, LoggingLevelSwitch levelSwitch)
     }
 
     public void GetEffectiveLevel(
-#if FEATURE_SPAN
         ReadOnlySpan<char> context,
```
sungam3r, line 61: I do not understand why span is used here. At the first glance it can be safely changed to string.

sungam3r:
I'm fine with new `System.Memory` conditional dependency. We used the same approach in GraphQL.NET.

sungam3r:
Not applicable?

PR #1882, opened 2023-03-26
remove MessageTemplateToken.StartIndex

SimonCropp:
@nblumhardt any ideas what MessageTemplateToken.StartIndex is for? it doesnt seem to be used in the code. and i cant see how it would add value to a consumer of serilog


nblumhardt:
These were added very early on, as part of an effort to create a ReSharper plug-in for Serilog:

https://github.com/serilog/serilog/pull/249

I am not sure what kinds of consumers might be broken now; it might not hurt to make the removal, but consider bringing these back if we find breakage during beta/compatibility testing?

PR #1884, opened 2023-03-27
Make Alignment and LevelOverride readonly

sungam3r:


src/Serilog/Parsing/PropertyToken.cs
```diff
@@ -32,7 +32,7 @@ public sealed class PropertyToken : MessageTemplateToken
     /// <param name="startIndex">The token's start index in the template.</param>
     /// <exception cref="ArgumentNullException">When <paramref name="propertyName"/> is <code>null</code></exception>
     /// <exception cref="ArgumentNullException">When <paramref name="rawText"/> is <code>null</code></exception>
-    public PropertyToken(string propertyName, string rawText, string? format = null, Alignment? alignment = null, Destructuring destructuring = Destructuring.Default, int startIndex = -1)
+    public PropertyToken(string propertyName, string rawText, string? format = null, in Alignment? alignment = null, Destructuring destructuring = Destructuring.Default, int startIndex = -1)
```
nblumhardt, line 35: This one would be a binary breaking change (see https://github.com/dotnet/docs/issues/3717#issuecomment-345094376), but given that it's very unlikely to be called by sink code or otherwise, it could be in the "test and revert if necessary during beta" category.

nblumhardt (approved):

src/Serilog/Parsing/PropertyToken.cs
sungam3r: ok

PR #1889, opened 2023-03-30
Reinstate `LoggerSinkConfiguration.Sink(ILogEventSink, LogEventLevel)`

nblumhardt:
I just switched Seq over to Serilog 3 dev and the latest versions of our dependencies, but because of some transitive dependencies in the build I hit:

```
System.MissingMethodException: Method not found: 'Serilog.LoggerConfiguration Serilog.Configuration.LoggerSinkConfiguration.Sink(Serilog.Core.ILogEventSink, Serilog.Events.LogEventLevel)'.
   at Serilog.SeqLoggerConfigurationExtensions.<>c__DisplayClass3_0.<Seq>b__0(LoggerSinkConfiguration wt)
   at Serilog.Configuration.LoggerSinkConfiguration.Wrap(LoggerSinkConfiguration loggerSinkConfiguration, Func`2 wrapSink, Action`1 configureWrappedSink, LogEventLevel restrictedToMinimumLevel, LoggingLevelSwitch levelSwitch)
   at Serilog.Configuration.LoggerSinkConfiguration.Conditional(Func`2 condition, Action`1 configureSink)
   at Serilog.SeqLoggerConfigurationExtensions.Seq(LoggerSinkConfiguration loggerSinkConfiguration, String serverUrl, LogEventLevel restrictedToMinimumLevel, Int32 batchPostingLimit, Nullable`1 period, String apiKey, String bufferBaseFilename, Nullable`1 bufferSizeLimitBytes, Nullable`1 eventBodyLimitBytes, LoggingLevelSwitch controlLevelSwitch, HttpMessageHandler messageHandler, Nullable`1 retainedInvalidPayloadsLimitBytes, Int32 queueSizeLimit)
   at Seq.Build.Program.Main(String[] rawArgs) in C:\projects\seq\ci\Seq.Build\Program.cs:line 33
```

We speculatively removed it in #1800.

I think this is going to be a common experience; for the cost of one overload that wasn't already `[Obsolete]` I think walking this change back is a good trade-off.

I've left the other removal from #1800 alone as we don't have any evidence of breakage at this point.

### DIFF
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -37,6 +37,7 @@ public class LoggerSinkConfiguration
     /// <seealso cref="Sink(ILogEventSink, LogEventLevel, LoggingLevelSwitch?)"/>
     /// <returns>Configuration object allowing method chaining.</returns>
     /// <remarks>Sink configuration methods that specify <paramref name="restrictedToMinimumLevel"/> should also specify <see cref="LoggingLevelSwitch"/>.</remarks>
+    [EditorBrowsable(EditorBrowsableState.Never)]
     public LoggerConfiguration Sink(
         ILogEventSink logEventSink,
         LogEventLevel restrictedToMinimumLevel)
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -33,6 +33,22 @@ public class LoggerSinkConfiguration
     /// </summary>
     /// <param name="logEventSink">The sink.</param>
     /// <param name="restrictedToMinimumLevel">The minimum level for
+    /// events passed through the sink.</param>
+    /// <seealso cref="Sink(ILogEventSink, LogEventLevel, LoggingLevelSwitch?)"/>
+    /// <returns>Configuration object allowing method chaining.</returns>
+    /// <remarks>Sink configuration methods that specify <paramref name="restrictedToMinimumLevel"/> should also specify <see cref="LoggingLevelSwitch"/>.</remarks>
+    public LoggerConfiguration Sink(
+        ILogEventSink logEventSink,
+        LogEventLevel restrictedToMinimumLevel)
+    {
+        return Sink(logEventSink, restrictedToMinimumLevel, null);
+    }
+
+    /// <summary>
+    /// Write log events to the specified <see cref="ILogEventSink"/>.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


augustoproiete (approved):
LGTM

test/Serilog.ApprovalTests/ApiApprovalTests.cs
```diff
@@ -16,6 +16,11 @@ public void PublicApi_Should_Not_Change_Unintentionally()
                 ExcludeAttributes = new[] { "System.Diagnostics.DebuggerDisplayAttribute" },
             });
 
-        publicApi.ShouldMatchApproved(options => options.WithFilenameGenerator((_, _, fileType, fileExtension) => $"{assembly.GetName().Name!}.{fileType}.{fileExtension}"));
+        publicApi.ShouldMatchApproved(options =>
+        {
+            // Comment this line out to view the failure as a diff. Leave it here so that CI builds don't hang when this test fails.
+            options.NoDiff();
```
sungam3r, line 22: How does hang look like? I do not use `NoDiff` in other repos and did not see issues.

src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -28,6 +28,23 @@ internal LoggerSinkConfiguration(LoggerConfiguration loggerConfiguration, Action
...
+    /// <summary>
+    /// Write log events to the specified <see cref="ILogEventSink"/>.
+    /// </summary>
+    /// <param name="logEventSink">The sink.</param>
+    /// <param name="restrictedToMinimumLevel">The minimum level for
+    /// events passed through the sink.</param>
+    /// <seealso cref="Sink(ILogEventSink, LogEventLevel, LoggingLevelSwitch?)"/>
+    /// <returns>Configuration object allowing method chaining.</returns>
+    /// <remarks>Sink configuration methods that specify <paramref name="restrictedToMinimumLevel"/> should also specify <see cref="LoggingLevelSwitch"/>.</remarks>
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
sungam3r, line 40: Why not obsolete?

nblumhardt: It's impossible to stop the compiler selecting it when doing overload resolution, and thus marking it obsolete results in a spurious warning.

test/Serilog.ApprovalTests/ApiApprovalTests.cs
nblumhardt: That's a good question, I may have jumped the gun here, based on experience with other tools that launch windowed executables for this.

PR #1899, opened 2023-05-01
Fix issues found by Roslynator

sungam3r:


src/Serilog/Rendering/Padding.cs
```diff
@@ -21,7 +21,7 @@ static class Padding
     /// <summary>
     /// Writes the provided value to the output, applying direction-based padding when <paramref name="alignment"/> is provided.
     /// </summary>
-    public static void Apply(TextWriter output, string value, in Alignment? alignment)
```
sungam3r, line 24: https://github.com/JosefPihrt/Roslynator/blob/main/docs/analyzers/RCS1242.md
https://github.com/dotnet/runtime/issues/23969
https://github.com/dotnet/corefx/pull/24997

SimonCropp:
my personal preference is to avoid roslyn analysers. been burned too many times with poor performance and bugs. 

when i do use them i only enable them for build. so disable realtime checking

src/Serilog/Debugging/LoggingFailedException.cs
```diff
@@ -20,6 +20,13 @@ namespace Serilog.Debugging;
 /// </summary>
 public class LoggingFailedException : Exception
 {
+    /// <summary>
+    /// Construct a <see cref="LoggingFailedException"/> to communicate a logging failure.
+    /// </summary>
+    public LoggingFailedException()
```
nblumhardt, line 26: Default exception constructors usually set a default message.

src/Serilog/Rendering/Padding.cs
nblumhardt: Is making `Alignment` readonly a suitable alternative? (The `in` keyword was added here thanks to advice from a competing analyzer, IIRC 🤷 )

test/Serilog.ApprovalTests/Serilog.approved.txt
```diff
@@ -830,7 +832,7 @@ namespace Serilog.Parsing
     }
     public sealed class PropertyToken : Serilog.Parsing.MessageTemplateToken
     {
-        public PropertyToken(string propertyName, string rawText, string? format = null, in Serilog.Parsing.Alignment? alignment = null, Serilog.Parsing.Destructuring destructuring = 0) { }
```
nblumhardt, line 833: Ditto here.

nblumhardt (commented):
I like some of the fixes/suggestions in here, but like @SimonCropp I'd lean towards keeping the solution lighter (and not having to potentially suppress another set of false-positives/unnecessary hints); Roslynator is installable as an IDE extension so perhaps that's enough, in this case?

src/Serilog/Debugging/LoggingFailedException.cs
SimonCropp: why does an exception need a default constructor?

nblumhardt: I've never really liked the guideline, but this is the source of the convention: https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#include-three-constructors-in-custom-exception-classes

test/Serilog.ApprovalTests/Serilog.approved.txt
sungam3r: `Alignment` is a readonly struct and works well with `in`m but `Nullable<Alignment` **is not** a readonly struct so `in` creates defensive copy as I understand.

> The in keyword was added here thanks to advice from a competing analyzer, IIRC 

No, I added in myself when making struct readonly.

src/Serilog/Rendering/Padding.cs
sungam3r: > Is making Alignment readonly a suitable alternative? 

`Alignment` is already is readonly struct.

sungam3r:
rebased on dev

sungam3r:
rebased on dev

sungam3r:
Rebased on dev. @nblumhardt Anything else to do here?

test/Serilog.ApprovalTests/Serilog.approved.txt
nblumhardt: Adding or removing `in` from a parameter is a breaking change according to https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules, I'd like to remove this but might need to wait for v5.

nblumhardt:
A couple of items still look outstanding. Any plans to continue this one? I am fine closing it otherwise. Cheers!

PR #1902, opened 2023-05-03
Accept/pass through the standard `levelSwitch` option in `WriteTo.Logger()`

nblumhardt:
This overload was introduced in #1890, so there's no compatibility concern here.


augustoproiete (approved):
LGTM

SimonCropp:
note that this is a breaking change

nblumhardt:
@SimonCropp in this case no, the signature being changed is a completely new one that's never appeared pre-3.0.

SimonCropp:
@nblumhardt ahhh. thanks for the correction

PR #1905, opened 2023-05-05
add net47

### LINKED ISSUE #1820: What min netframework should serilog 3 support?

SimonCropp:
fixes #1820

### DIFF
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -15,29 +15,26 @@
     <IsTrimmable>true</IsTrimmable>
     <NoWarn>$(NoWarn);CS1437;CS1570</NoWarn>
     <PolySharpIncludeRuntimeSupportedAttributes>true</PolySharpIncludeRuntimeSupportedAttributes>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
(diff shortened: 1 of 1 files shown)


src/Serilog/Serilog.csproj
```diff
@@ -24,16 +24,22 @@
...
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net47' " />
+
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net471' ">
     <DefineConstants>$(DefineConstants);FEATURE_ITUPLE</DefineConstants>
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
   </PropertyGroup>
+
   <ItemGroup Condition=" '$(TargetFramework)' == 'net471' ">
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
nblumhardt, line 39: I don't think this is actually needed for `net471`? (I notice it's missing from the added `net47` so seems like we should get rid of it from both).

nblumhardt: I'll commit an edit so that we can finalize this one :-)

nblumhardt:
Thanks @SimonCropp!

PR #1911, opened 2023-05-26
Use `JsonValueFormatter` to implement classic `JsonFormatter`

nblumhardt:
Since 3.0 removes all the obsolete `JsonFormatter` virtual methods, we can now seal this and remove the duplicated functionality that existed solely for subclass use.

Feeling like this one's a decent win :-)

### DIFF
--- a/src/Serilog/Formatting/Json/JsonFormatter.cs
+++ b/src/Serilog/Formatting/Json/JsonFormatter.cs
@@ -80,9 +80,15 @@ public sealed class JsonFormatter : ITextFormatter
         if (logEvent.Properties.Count != 0)
         {
             output.Write(",\"Properties\":{");
+
+            char? propertyDelimiter = null;
             foreach (var property in logEvent.Properties)
             {
-                output.Write(',');
+                if (propertyDelimiter != null)
+                    output.Write(propertyDelimiter.Value);
+                else
+                    propertyDelimiter = ',';
+
                 JsonValueFormatter.WriteQuotedJsonString(property.Key, output);
                 output.Write(':');
                 _jsonValueFormatter.Format(property.Value, output);
@@ -126,8 +132,9 @@ public sealed class JsonFormatter : ITextFormatter
         {
             if (propertyDelimiter != null)
                 output.Write(propertyDelimiter.Value);
+            else
+                propertyDelimiter = ',';
 
-            propertyDelimiter = ',';
             output.Write('"');
             output.Write(propertyFormats.Key);
             output.Write("\":[");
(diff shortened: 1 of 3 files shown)


src/Serilog/Formatting/Json/JsonFormatter.cs
```diff
@@ -190,22 +147,24 @@ void WriteRenderingsValues(IGrouping<string, PropertyToken>[] tokensWithFormat,
...
+            output.Write('"');
+            output.Write(property.Key);
+            output.Write("\":");
+            _jsonValueFormatter.Format(property.Value, output);
+            precedingDelimiter = ',';
         }
+        output.Write('}');
     }
 
+
```
sungam3r, line 167:
```suggestion
```

src/Serilog/Formatting/Json/JsonFormatter.cs
```diff
@@ -316,141 +218,7 @@ void WriteJsonProperty(string name, object? value, ref char? precedingDelimiter,
         output.Write('"');
         output.Write(name);
         output.Write("\":");
-        WriteLiteral(value, output);
+        _jsonValueFormatter.Format(new ScalarValue(value), output);
```
sungam3r, line 221: Allocation of `ScalarValue` on each property, not good.

sungam3r: I suggest to add `internal` to `protected virtual void FormatLiteralValue(object? value, TextWriter output)` and call it directly from here.

sungam3r: Sorry. Not from here, but from other places
изображение

Then you need one more `WriteJsonProperty` private method.

nblumhardt: Good suggestion, thanks; I've internalized the idea that `JsonFormatter` is the legacy approach, but since we're still shipping it I should pay more attention 😅 - will follow up 👍

nblumhardt: Fixed; managed to clobber quite a lot of (now) unnecessary code, too 👍

nblumhardt:
image

😎 

Think this is ready to go.

nblumhardt:
Odd, tests are all green for me in Rider, will dig in 🧐

nblumhardt:
Test runner wasn't rebuilding the solution 😅  - CI should be all green now.

nblumhardt:
Sending this one through; tests are green, feedback addressed - hopefully no observable behavioral differences.

We're still yet to see whether any significant proportion of the ecosystem depends on subclassing this type, but winding back the removal of virtual extension points would not be any harder atop the changes in this PR.

PR #1922, opened 2023-06-19
Modernize the README a little more

nblumhardt:


README.md
```diff
@@ -58,35 +58,39 @@ Supporting structured data doesn't mean giving up text: when Serilog writes even
...
+    .WriteTo.File("log.txt",
+        rollingInterval: RollingInterval.Day,
+        rollOnFileSizeLimit: true)
+    .CreateLogger();
+
+try
+{
+    // Your program here...
+    const string name = "Serilog";
+    Log.Information("Hello, {Name}!", name);
```
augustoproiete, line 85: Would be useful to throw an exception here to show it working?
Or perhaps a comment

Because  `Log.Information` would never throw an exception, so it's a little odd to see.

```suggestion
    Log.Information("Hello, {Name}!", name);
    throw new InvalidOperationException("Oops...");
```

Issue #1929, opened 2023-06-28
Access sinks


bdovaz:
**Is your feature request related to a problem? Please describe.**

@nblumhardt I am using `Serilog.Settings.Configuration` and the `Serilog.Sinks.Graylog` sink.

The problem I have is that it seems that the use case of mixing and reading the appsettings configuration and then configuring or overriding a sink from code is not covered.

Graylog's sink allows for example to set up a transport factory which is obviously something that cannot be set up in a json file: https://github.com/whir1/serilog-sinks-graylog/blob/master/src/Serilog.Sinks.Graylog.Core/GraylogSinkOptions.cs#L211C28-L211C28

**Describe the solution you'd like**

Although the ideal solution requires more changes in the Graylog repository, this Serilog Core repository should at least expose information that so far is not exposed: https://github.com/serilog/serilog/blob/3016aa31772f882d294ed356d13207d6f55901c0/src/Serilog/LoggerConfiguration.cs#L22

Proposal:

```csharp
public class LoggerConfiguration
{
    public IEnumerable<ILogEventSink> Sinks => _logEventSinks;
}
```

And then:

```csharp
var loggerConfiguration = new LoggerConfiguration()
        .ReadFrom.Configuration(configuration);

var graylogSink = loggerConfiguration.Sinks.OfType<GraylogSink>().First();
```

**Describe alternatives you've considered**

Right now I'm doing it through reflection:

```csharp
FieldInfo? logEventSinksFieldInfo = loggerConfiguration.GetType().GetField("_logEventSinks", BindingFlags.NonPublic | BindingFlags.Instance);

List<ILogEventSink> logEventSinks = (List<ILogEventSink>)logEventSinksFieldInfo!.GetValue(loggerConfiguration)!;
```

**Additional context**

sungam3r:
Agree. This is a common problem of design with an implicit configuration of sinks from configuration. As I understand Graylog repository **may** provide needed overload accepting delegate to set `Func<ITransport> TransportFactory` but it will not help in case of using `ReadFrom.Configuration(configuration)`. I should note that configuration subsystem of Serilog supports some non-trivial features like constructing delegates so with some effort you **may** achive your goal but I agree that this it is not convenient to do and personally I would not.

Also note that sinks are often considered to be immutable after creating. You provided `Func<ITransport> TransportFactory` property that exists on options object passed into Graylog sink. `GraylogSink` **may** use this options object only in ctor so further modification of it does nothing.

[...]

PR #1930, opened 2023-06-29
Introduce LoggerSinkSetupConfiguration

sungam3r:
Draft demo to address #1929.


nblumhardt:
I don't think #1929 will be merged; it's remaining for discussion only, so I'll close this here (open PRs take some review time on each triage pass). Thanks for taking a look!

PR #1936, opened 2023-07-07
Update upgrading instructions in README.md

nblumhardt:
I don't think there are many people upgrading from 1.x to 2.0 these days 😅


augustoproiete (approved):
LGTM

Issue #1944, opened 2023-08-08
Only case sensitive check for minimumLevel override


Phil91:
**Description**
Serilog currently only supports case sensitive matches when overriding the log level for a specific namespace, when for example using Serilog with Serilog.Settings.Configuration and setting env variables for a docker image the Namespaces can't be all upper case.

**Reproduction**
Setting up a test project with Serilog and Serilog.Settings.Configuration installed, configure Serilog for example like this:

``` C#
host.UseSerilog((context, configuration) =>
        {
            configuration
                .WriteTo.Console(new JsonFormatter())
                .ReadFrom.Configuration(context.Configuration);
        })
```

add a appsettings file with the following configuration:

``` json
"Serilog": {
    "Using":  [ "Serilog.Sinks.Console" ],
    "MinimumLevel": {
      "Default": "Information",
      "Override": {
        "Test.Nmspc": "Information"
      }
    },
    "WriteTo": [
      { "Name": "Console" }
    ],
    "Enrich": [
      "FromLogContext"
    ],
    "Properties": {
      "Application": "Org.Eclipse.TractusX.Portal.Backend.Administration.Service"
    }
  },
```

Then execute the application with the following env var: SERILOG__MINIMUMLEVEL__OVERRIDE__TEST.NMSPC=Debug

The minimum logLevel for Loggers within that namespace will be Information

If needed I can provide an example repo.

**Expected behavior**
The minimum log level for logger instances creates within the Test.Nmspc Namespace should be Debug.

**Relevant package, tooling and runtime versions**
- Serilog 3.0.1
- Serilog.Settings.Configuration 3.4.0
- aspnet core 6 & 7 (both tested)
- docker image with environment variables (tested locally with a secret file as well on a win 10 machine)

**Additional context**
I've already provided a PR with a change for the case insensitive check.

Phil91:
> Duplicate #1335 - would still love to see a case-insensitive match.

@tillig didn't saw that. thanks for linking.

tillig:
Somewhat related - serilog/serilog-settings-configuration#386 found that the parser for the log level enum in configuration was inconsistently case-sensitive and case-insensitive, and that got resolved to be case-insensitive everywhere. That makes things just slightly more inconsistent: enum parsing is case-insensitive, but the keys are case-sensitive.

PureKrome:
Why can't this be a 'flag' ? Give the developer the option?

nblumhardt:
[...]

PR #1945, opened 2023-08-08
fix(LevelOverrideMap): change levelSwitch check

Phil91:
changed the levelSwitch check from case sensitive namespace checks to a case insensitve check

Refs: #1944

### DIFF
--- a/src/Serilog/Core/LevelOverrideMap.cs
+++ b/src/Serilog/Core/LevelOverrideMap.cs
@@ -68,7 +68,7 @@ class LevelOverrideMap
     {
         foreach (var levelOverride in _overrides)
         {
-            if (context.StartsWith(levelOverride.Context) &&
+            if (context.StartsWith(levelOverride.Context, StringComparison.CurrentCultureIgnoreCase) &&
                 (context.Length == levelOverride.Context.Length || context[levelOverride.Context.Length] == '.'))
             {
                 minimumLevel = LevelAlias.Minimum;
--- a/src/Serilog/Core/LevelOverrideMap.cs
+++ b/src/Serilog/Core/LevelOverrideMap.cs
@@ -68,7 +68,7 @@ class LevelOverrideMap
     {
         foreach (var levelOverride in _overrides)
         {
-            if (context.StartsWith(levelOverride.Context, StringComparison.CurrentCultureIgnoreCase) &&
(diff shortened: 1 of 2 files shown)


src/Serilog/Core/LevelOverrideMap.cs
```diff
@@ -68,7 +68,7 @@ public LevelOverride(string context, LoggingLevelSwitch levelSwitch)
     {
         foreach (var levelOverride in _overrides)
         {
-            if (context.StartsWith(levelOverride.Context) &&
+            if (context.StartsWith(levelOverride.Context, StringComparison.CurrentCultureIgnoreCase) &&
```
bartelink, line 71: would OrdinalIgnoreCase work? Introducing a culture dependency is not something I'd do lightly

bartelink: That was a reasonable choice.
I guess the [perf of switching to ordinal](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/String.Comparison.cs,23804789ea4c9c0e) would be a good thing too, if we assume that the values, being programmatic identifiers, should hopefully not require culture-sensitivity.

nblumhardt: We steer clear of the default/current culture whenever possible, because it can lead to surprising and unexpected deployment-time behavior changes. Most culture-specific API use in Serilog either relies on either ordinal/"invariant" culture operations, if the task is programmatic and wouldn't generally be expected to vary in behavior based on the _user's_ culture, or it accepts an `IFormatProvider` or equivalent.

In this case, the programmer's culture and the eventual deployment environment culture might not bear any relationship to each other at all (e.g. the program and its config files might be written in Australia, but run on a machine in Germany), so ordinal/invariant comparisons seem like the best choice.

Phil91: @nblumhardt  thanks for the explanation

bartelink: >Australia, but run on a machine in Germany

Missed opportunity to say "in Austria" :P

nblumhardt:
Sorry about leaving this one dangling, @Phil91. As you'd have spotted, Travis dug in with benchmarks around this change and the issue was eventually resolved in #2045. Thanks for the issue, and for the first version of the PR all the same!

Issue #1946, opened 2023-08-21
MessageTemplateCache performance improvement


epeshk:
Currently, `MessageTemplateCache` uses `HashTable` (or `Dictionary<T>/ConcurrentDictionary<T>` in alternative implementations from `Serilog.PerformanceTests.Support`) with the `string` key to cache `MessageTemplate` objects.

These hash maps are initialized with default by-value comparer, and every cache lookup requires a linear scan of the `string messageTemplate` to compute the hash code (hash codes for strings aren't cached in current .NET implementations). This hash code computation is inefficient, especially for long message templates.

Since message template string objects are typically reused, and even constant for the cases like: `_logger.Info("Log message, elapsed: {ElapsedTime}", ...)`, this cache may compare strings by-reference, instead of by-value, skipping the slow hash code computation.

I don't know if there any cases when by-value caching is useful, so I tested this proposal with two-level caching: by-reference, then by-value.

### MessageTemplateCacheBenchmark_Cached benchmark for Hashtable-based MessageTemplateCache

Before optimization:
```
|    Method | Items | MaxDegreeOfParallelism |      Mean |    Error |   StdDev |   Gen0 | Allocated |
|---------- |------ |----------------------- |----------:|---------:|---------:|-------:|----------:|
| Hashtable |    10 |                     -1 |  29.34 us | 0.127 us | 0.119 us | 0.4578 |   7.77 KB |
| Hashtable |    10 |                      1 |  72.42 us | 0.379 us | 0.354 us |      - |   1.56 KB |
| Hashtable |    20 |                     -1 |  35.69 us | 0.069 us | 0.065 us | 0.4883 |   8.17 KB |
| Hashtable |    20 |                      1 | 145.06 us | 0.719 us | 0.638 us |      - |   1.56 KB |
| Hashtable |    50 |                     -1 |  43.35 us | 0.020 us | 0.018 us | 0.4883 |      8 KB |
| Hashtable |    50 |                      1 | 363.45 us | 0.809 us | 0.717 us |      - |   1.56 KB |
| Hashtable |   100 |                     -1 |  60.83 us | 0.188 us | 0.176 us | 0.4883 |    8.5 KB |
| Hashtable |   100 |                      1 | 734.14 us | 3.448 us | 2.879 us |      - |   1.56 KB |
| Hashtable |  1000 |                     -1 |  61.83 us | 0.256 us | 0.240 us | 0.4883 |   8.51 KB |
[...]

PR #1947, opened 2023-08-21
By reference string comparison in template cache

epeshk:
Related issue: https://github.com/serilog/serilog/issues/1946


nblumhardt:
How would the numbers come out if we ditched the fall-through to the by-value cache?

There's a good case to be made for reducing memory use by dropping the second dictionary; anyone who is formatting message "templates" dynamically is probably not actually using message templates and is likely to be generating unique strings anyway, so the situations that will actually benefit from the second layer of caching are pretty limited.

epeshk:
Yes, I agree that it's better to remove by-value cache completely

nblumhardt:
LGTM 👍 

Interested whether anyone else can spot gotchas I might have missed so will leave this a few more days.

PR #1948, opened 2023-08-21
Removed temporary array allocations for properties

epeshk:
Similar to https://github.com/serilog/serilog/pull/1779, but uses on-stack allocation instead of pooling.

### DIFF
--- a/src/Serilog/Capturing/PropertyBinder.cs
+++ b/src/Serilog/Capturing/PropertyBinder.cs
@@ -34,9 +34,13 @@ class PropertyBinder
     /// represented in the message template.</param>
     /// <returns>A list of properties; if the template is malformed then
     /// this will be empty.</returns>
-    public EventProperty[] ConstructProperties(MessageTemplate messageTemplate, object?[]? messageTemplateParameters)
+#if FEATURE_SPAN
+    public EventProperty[] ConstructProperties(MessageTemplate messageTemplate, ReadOnlySpan<object?> messageTemplateParameters)
+#else
+    public EventProperty[] ConstructProperties(MessageTemplate messageTemplate, object?[] messageTemplateParameters)
+#endif
     {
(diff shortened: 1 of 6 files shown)


epeshk:
I'm not sure it's worth (and possible) to optimize the default `ILogger` interface implementation before `params ReadOnlySpan<T>` support is released in C#.

https://github.com/dotnet/csharplang/blob/main/proposals/params-span.md

nblumhardt:
I think this is the way to go - looks like CI produced some failures, though.

src/Serilog/Capturing/PropertyBinder.cs
```diff
@@ -34,9 +34,13 @@ public PropertyBinder(PropertyValueConverter valueConverter)
...
     /// this will be empty.</returns>
-    public EventProperty[] ConstructProperties(MessageTemplate messageTemplate, object?[]? messageTemplateParameters)
+#if FEATURE_SPAN
+    public EventProperty[] ConstructProperties(MessageTemplate messageTemplate, ReadOnlySpan<object?> messageTemplateParameters)
+#else
+    public EventProperty[] ConstructProperties(MessageTemplate messageTemplate, object?[] messageTemplateParameters)
+#endif
     {
-        if (messageTemplateParameters == null || messageTemplateParameters.Length == 0)
+        if (messageTemplateParameters.Length == 0)
```
nblumhardt, line 43: May be worth including a few quick speculative tests to check that removing the `== null` test here is safe even when callers without strict null checking enabled call various logger methods, e.g.:

```csharp
[Fact]
public void NullMessageTemplateParametersDoNotBreakBinding()
{
    var log = new LoggerConfiguration().WriteTo.Sink(new CollectingSink()).CreateLogger();
    log.Information("test", (object?[]?) null);
    log.BindMessageTemplate("test", (object?[]?)null, out _, out _);
}
```

?

nblumhardt (approved):

PR #1959, opened 2023-09-30
PropertyValueConverter optimization: avoiding hashset lookup for primitive types

epeshk:
Optimization for `PropertyValueConverter`. Skips hashset lookup in `SimpleScalarConversionPolicy` for primitive types. The [primitive types](https://learn.microsoft.com/en-us/dotnet/api/system.type.isprimitive?view=net-7.0#remarks) are Boolean, Byte, SByte, Int16, UInt16, Int32, UInt32, Int64, UInt64, IntPtr, UIntPtr, Char, Double, and Single.

It additionally handles `sbyte`, `nint`, `nuint`, but these types could be excluded with additional check without performance penalty. I'm not sure if `sbyte` was originally excluded on purpose.

```cs
type.IsPrimitive && type != typeof(sbyte) && type != typeof(IntPtr) && type != typeof(UIntPtr)
// or
type.IsPrimitive && Type.GetTypeCode(type) is not TypeCode.Object and not TypeCode.SByte
```

Before:

|        Method |     Mean |   Error |  StdDev |   Gen0 | Allocated |
|-------------- |---------:|--------:|--------:|-------:|----------:|
| IntProperties | 268.5 ns | 1.74 ns | 1.63 ns | 0.0296 |     496 B |

After:

|        Method |     Mean |   Error |  StdDev |   Gen0 | Allocated |
|-------------- |---------:|--------:|--------:|-------:|----------:|
| IntProperties | 195.1 ns | 3.42 ns | 3.20 ns | 0.0296 |     496 B |

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -22,10 +22,7 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
 {
     static readonly HashSet<Type> BuiltInScalarTypes = new()
     {
-        typeof(bool),
-        typeof(char),
-        typeof(byte), typeof(short), typeof(ushort), typeof(int), typeof(uint),
-        typeof(long), typeof(ulong), typeof(float), typeof(double), typeof(decimal),
+        typeof(decimal),
         typeof(string),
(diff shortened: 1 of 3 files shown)


nblumhardt:
Looks like a nice easy win 👍

PR #1960, opened 2023-10-02
Added missing DynamicallyAccessedMembers in PropertyValueConverter

epeshk:
Fixes build without `FEATURE_ITUPLE` constant defined.

> `Error IL2070 : 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicFields' in call to 'System.Type.GetFields(BindingFlags)'. The parameter 'type' of method 'Serilog.Capturing.PropertyValueConverter.TryConvertValueTuple(Object, Type, Destructuring, out LogEventPropertyValue)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.`


nblumhardt:
Thanks! 👍

PR #1962, opened 2023-10-09
Removed temporary string allocations for rendering string in ScalarValue

Jakimar:
Currenly when rendering ScalarValue for MessageTemplate it's always escape double quotes with "string.Replace" which leads to creating new string, we can avoid this by using Span's and manually replacing double quotes.

Before:

|                         Method |       Mean |     Error |    StdDev |   Gen0 | Allocated |
|------------------------------- |-----------:|----------:|----------:|-------:|----------:|
|       TemplateWithNoProperties |   2.100 ns | 0.1175 ns | 0.1722 ns |      - |         - |
|   TemplateWithVariedProperties | 167.852 ns | 1.0368 ns | 1.4869 ns | 0.0153 |      96 B |
| TemplateWithQuotesEscapeString | 682.964 ns | 5.0617 ns | 7.4194 ns | 0.1326 |     832 B |

After:
|                         Method |       Mean |     Error |    StdDev |   Gen0 | Allocated |
|------------------------------- |-----------:|----------:|----------:|-------:|----------:|
|       TemplateWithNoProperties |   2.045 ns | 0.0375 ns | 0.0537 ns |      - |         - |
|   TemplateWithVariedProperties | 183.667 ns | 2.7135 ns | 3.9774 ns | 0.0153 |      96 B |
| TemplateWithQuotesEscapeString | 296.243 ns | 2.3111 ns | 3.3875 ns |      - |         - |

### DIFF
--- a/src/Serilog/Events/ScalarValue.cs
+++ b/src/Serilog/Events/ScalarValue.cs
@@ -67,9 +67,13 @@ public class ScalarValue : LogEventPropertyValue
         {
             if (format != "l")
             {
+#if FEATURE_SPAN
+                MessageEscaper.WriteString(output, s);
+#else
                 output.Write('"');
                 output.Write(s.Replace("\"", "\\\""));
                 output.Write('"');
+#endif
             }
             else
             {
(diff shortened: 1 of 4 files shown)


nblumhardt:
Thanks for the PR. This part of the codebase is currently in "maintenance mode" while better formatting routines are being developed in _Serilog.Formatting.Compact_ and _Serilog.Expressions_.

I think leaving things as simple as possible here would be preferable, so that we don't create more inertia preserving the status quo.

This kind of approach seems like it would be valuable for those other libraries, though!

Issue #1963, opened 2023-10-09
ReusableStringWriter: Potentional high memory usage\leak


Jakimar:
**Description**
When working on https://github.com/serilog/serilog/pull/1962 i did notice what in ReusableStringWriter underlying buffer of StringBuilder can grow as much as biggest message we log, so in case we log some really big message, like more than 32_000 symbols(or even >1_000_000) we are gonna stuck with StringBuilder buffer this size forever even if we never log anything this big for application lifetime.

**Expected behavior**
If "StringBuilder.Capacity > 32748" or any other reasonable buffer size threshold we dispose ReusableStringWriter, something like this:
```

    /// <summary>
    /// Clear this instance and prepare it for reuse in the future.
    /// </summary>
    protected override void Dispose(bool disposing)
    {
        // We don't call base.Dispose because all it does is mark the writer as closed so it can't be
        // written to and we want to keep it open as reusable writer.
        var sb = GetStringBuilder();
        if (sb.Capacity > 32748)
        {
            base.Dispose();
            return;
        }
        sb.Clear();
        _pooledWriter = this;
    }
```

PR #1964, opened 2023-10-09
ReusableStringWriter: Dispose instance with too big buffer

Jakimar:
Fixes https://github.com/serilog/serilog/issues/1963 
StringBuilderCapacityThreshold is subject to discuss

### DIFF
--- a/src/Serilog/Rendering/ReusableStringWriter.cs
+++ b/src/Serilog/Rendering/ReusableStringWriter.cs
@@ -9,6 +9,11 @@ class ReusableStringWriter : StringWriter
     static ReusableStringWriter? _pooledWriter;
 
     /// <summary>
+    /// Max capacity of StringBuilder we keep for next using
+    /// </summary>
+    const int StringBuilderCapacityThreshold = 32678;
+
+    /// <summary>
     /// Gets already created StringWriter if there is one available or creates a new one.
     /// </summary>
     /// <param name="formatProvider"></param>
@@ -34,9 +39,15 @@ class ReusableStringWriter : StringWriter
     /// </summary>
     protected override void Dispose(bool disposing)
     {
+        var sb = GetStringBuilder();
+        if (sb.Capacity > StringBuilderCapacityThreshold)
+        {
+            base.Dispose();
+            return;
+        }
         // We don't call base.Dispose because all it does is mark the writer as closed so it can't be
         // written to and we want to keep it open as reusable writer.
-        GetStringBuilder().Clear();
+        sb.Clear();
         _pooledWriter = this;
     }
 }


nblumhardt:
Makes sense, and the default seems reasonable - thanks 👍

src/Serilog/Rendering/ReusableStringWriter.cs
```diff
@@ -8,6 +8,11 @@ class ReusableStringWriter : StringWriter
     [ThreadStatic]
     static ReusableStringWriter? _pooledWriter;
 
+    /// <summary>
+    /// Max capacity of StringBuilder we keep for next using
+    /// </summary>
+    const int StringBuilderCapacityThreshold = 32678;
```
bartelink, line 14: (This is a great insight in general - not actually criticising, just having a little side natter to myself!)
My eyes! should be 32768 :)
I'd also vote for more like 8192 if it was to be changed - even that is a monster size message for any normal usage. I know LOH threshold is 85KB but over time trends like https://maoni0.medium.com/dynamically-adapting-to-application-sizes-2d72fcb6f1ea will surely 'other' sizes in this territory?
And if you want to see real pedantry in this space: https://github.com/jet/propulsion/pull/234/files#diff-2b8832d4b8d2a4bcf3625a5670b5b06d5d286aa1053eaa81a19f006e6b089ef9L33

Jakimar: Oh my bad, it's just a typo, i was expecting some opinions on size threshold before merging so i didn't really double checked everything. StringBuilder buffer can contains a lot of smaller chunks up to 8000 symbols each( https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Text/StringBuilder.cs,71 ), so for LOH point of view it's really does not matter 8192 or 32768.

Jakimar: I was wrong https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Text/StringBuilder.cs,2433
MaxChunkSize can be size of instered string if it's bigger than 8000. So yeah, 8192 can be better in some cases.

nblumhardt: Sharp eyes @bartelink - sorry about my trigger-happy merge; I'll zip back in and YOLO-edit the constant 😅 

I'm on board with tuning that constant up or down if you think it can be improved. My first thought was just to avoid "I logged a GB string and now I've wasted 1 GB RAM" types of scenarios. The optimal value would probably depend on the intended resource consumption and behavior of the app 🤔

Issue #1966, opened 2023-10-10
Potential V4 nits/breaking changes?


bartelink:
Minor nits from perusing the source; most are likely breaking changes best kept for v4

- TFMs
  - maybe a note somewhere about having a `net6.0` TFM but explaining that no egregious later ones will be added see (see reasoning in https://github.com/serilog/serilog-sinks-console/pull/145#issuecomment-1709862581)
  - remove net5.0 TFM (see above)
  - remove `net7.0` TFM (atm it has identical DefineConstants values to `net6.0`)
  - Remove netcoreapp3.1, net5.0 from tests
- README
  - Is there a stance/policy on copyright years (atm readme has 2020)
  - `var log = new LoggerConfiguration()` should gain a `using`
  - `Log.CloseAndFlush();` in static example
    -  should probably gain `a `// Normally in a `try/finally`; ensures File, Async etc sinks have been flushed before program exits` 
    - or, more tersely '// ensure buffering or async Sinks are flushed before app exits'
  - move gitter mention further down the list? or remove?
  - > We welcome bug reports and suggestions through our
   
     We welcome reproducable bug reports and detailed feature requests through our [issue tracker](https://github.com/serilog/serilog/issues) here on GitHub; please note the other avenues are much better or for quick questions or seeking usage help 
- 🤔 remove changes.md given it stops dead at 2.10 (or sync with GH release notes)
- AssemblyVersion [for 3.x] is `2.0`, maybe shift to 4.0 in v4?
- on https://github.com/serilog/serilog/wiki/Getting-Started, in the sample app
  - maybe lose `and Windows Phone 8+.` (maybe add some Blazor mention?)
  - use a top level `namespace SerilogExample`
  - consider using `await Log.CloseAndFlushAsync()`
  - (if this makes sense, ping me and I can do the edits)
- 914M on https://www.nuget.org/packages/Serilog/ is insane (currently # 8 all time, should be 7 soon) - not long to go to 1B!  

(Happy to do PRs for any that get agreed; equally happy for someone else to do them...)

nblumhardt:
These all look good, just dropping thoughts in here to try unblocking as many as possible.

> TFMs

This one will need a standalone ticket for discussion, keen to simplify things where possible but also eager to avoid covering all the ground covered in the last thread over again :-)

> Is there a stance/policy on copyright years (atm readme has 2020)

I think we've moved away from specifying the year, "Copyright &copy; Serilog Contributors" is preferred

[...]

PR #1969, opened 2023-10-21
chore(docs): Markdown housekeeping

bartelink:
related: https://github.com/serilog/serilog/issues/1966
- Remove year (2020) from copyright message on readme (note this propagates to nuget)
- Remove mention of gitter (it's not well-monitored, so better to have people use GitHub issues and/or PRs to to/fro ideas)
- Added NuGet package copyright (see far bottom right of https://www.nuget.org/packages/equinox)

Open issues:
- [ ] Some files are missing copyright headers. Many have incomplete years.
   
   I feel the best answer is to remove them in favor of having one copyright year range in the NOTICE. This seems to be valid based on https://www.apache.org/legal/apply-license.html#copy-per-file and general SO searching. If this is not agreed, perhaps I could remove them from the test files only?
   
   Failing that, I can add the header to all source files.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,6 +7,7 @@
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
+    <Copyright>Copyright © 2013-23 Serilog Contributors</Copyright>
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
     <Reference Include="System" />
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ catch (Exception ex)
 }
 finally
 {
-    await Log.CloseAndFlushAsync();
+    await Log.CloseAndFlushAsync(); // ensure buffering or async Sinks are flushed before app exits
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -122,4 +122,4 @@ When contributing please keep in mind our [Code of Conduct](CODE_OF_CONDUCT.md).
 | `dev`  | [![Build status](https://ci.appveyor.com/api/projects/status/b9rm3l7kduryjgcj/branch/dev?svg=true)](https://ci.appveyor.com/project/serilog/serilog/branch/dev)       |
 | `main` | [![Build status](https://ci.appveyor.com/api/projects/status/b9rm3l7kduryjgcj/branch/master?svg=true)](https://ci.appveyor.com/project/serilog/serilog/branch/master) |
 
-_Serilog is copyright &copy; 2013-2020 Serilog Contributors - Provided under the [Apache License, Version 2.0](http://apache.org/licenses/LICENSE-2.0.html). Needle and thread logo a derivative of work by [Kenneth Appiah](http://www.kensets.com/)._
+_Serilog is copyright &copy; Serilog Contributors - Provided under the [Apache License, Version 2.0](http://apache.org/licenses/LICENSE-2.0.html). Needle and thread logo a derivative of work by [Kenneth Appiah](http://www.kensets.com/)._
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ catch (Exception ex)
 }
 finally
 {
-    await Log.CloseAndFlushAsync(); // ensure buffering or async Sinks are flushed before app exits
+    await Log.CloseAndFlushAsync(); // ensure all logs written before app exits
 }
 ```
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -91,7 +91,7 @@ catch (Exception ex)
 }
 finally
 {
-    await Log.CloseAndFlushAsync();
+    await Log.CloseAndFlushAsync(); // ensure buffering or async Sinks are flushed before app exits
```
bartelink, line 94: Open to removing this given its a lot of words...

Directory.Build.props
```diff
@@ -7,10 +7,11 @@
     <CheckEolTargetFramework>false</CheckEolTargetFramework>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
+    <Copyright>Copyright © 2013-23 Serilog Contributors</Copyright>
```
bartelink, line 10: This flows up to the nuget landing page/ package metadata (and I think the FILEVERSIONINFO on the DLL)

README.md
```diff
@@ -103,12 +103,15 @@ To learn more about Serilog, check out the [documentation](https://github.com/se
...
 Serilog has an active and helpful community who are happy to help point you in the right direction or work through any issues you might encounter. You can get in touch via:
 
- * [Stack Overflow](http://stackoverflow.com/questions/tagged/serilog) &mdash; this is the best place to start if you have a question
+ * [Stack Overflow](http://stackoverflow.com/questions/tagged/serilog) &mdash; this is the best place to start if you have a question. Many track the `serilog` tag there.
  * The [#serilog tag on Twitter](https://twitter.com/search?q=%23serilog)
  * [Serilog-related courses on Pluralsight](https://www.pluralsight.com/search/?q=serilog)
 
 We welcome bug reports and suggestions through our [issue tracker](https://github.com/serilog/serilog/issues) here on GitHub.
 
+We welcome reproducible bug reports and detailed feature requests through [our GitHub issue tracker](https://github.com/serilog/serilog/issues);
```
bartelink, line 112: trying to temper the above statement to avoid drive by questions - can defintiely be worded better, so open to suggestions

Directory.Build.props
bartelink: Hm the Authors in the package (i.e., `<Authors>Serilog Contributors</Authors>`) might get appended in some renderings?

bartelink: Seems they are complementary per https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#copyright, so leaving it that way

README.md
nblumhardt: LGTM, going to remove the older version above?

bartelink: Done, silly me!

bartelink:
Thoughts on the open issue (in the overview) ? (i.e. add missing copyright headers in impl/test files and/or a copyright to the LICENSE and/or adjust the years per file) ?

README.md
nblumhardt: LGTM, maybe we could shorten it to `// ensure all logs are written before the app exits` ?

nblumhardt:
RE the copyright headers, I think sorting out the year range (&copy; in the source files where they appear, and full range in the LICENSE file) would be my preference; no super strong feelings there though.

README.md
bartelink: That works, done

bartelink:
Not sure what sorting out the year range would mean; make them consistent and then update them every year?
Ensure they reflect the time of last edit per file?
Or strip off the date in the headers?

Happy to do anything of the following that you want
0. just leave it
1. add copyright and year range in LICENSE
2. remove years in file headers, leaving it as `Copyright © Serilog Contributors` or maybe `Copyright 2013-` ...
3. add headers to all source files missing them
4. add headers to all test files missing them
5. remove all headers from source files

bartelink:
@nblumhardt  can you have a peep at File sink PR 258 please?
Happy to continue reviewing until xmldoc aligns with impl, but perhaps it should just be called off as the motivation/need seems dubious at this point

nblumhardt:
Thanks for the follow-up. I think even just (2) - using `Copyright © Serilog Contributors`, and (3) add them consistently to all source files would do the trick, if that sounds good to you?

bartelink:
@nblumhardt thanks for the response - will merge this and do it as a fresh PR another time

One final question: do I remove headers from test files (a number 6 option I didnt propose!), or apply them consistently same as for source files (i.e. 4) ?

nblumhardt:
I'm fine either way on the test files - they don't tend to "travel" like the source files do (e.g. https://github.com/Azure/serilog-sinks-azuredataexplorer/issues/16).

bartelink:
Hm, having seen that I'll apply them to the test files too, unless that becomes a truly huge change..

Issue #1970, opened 2023-10-21
Medium term strategy for Target Framework Monikers


bartelink:
**Atm Serilog 3.X targets `net462;net471;netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0` and will continue to do so.**

This was arrived at organically. Going forward, the TL;DR of the policy of what to target is:
- we support `net462`, `net471` and `netstandard2.0` where possible.
- Where there are compelling features that dictate targeting newer TFMs, we favor the lowest even numbered TFM >= `net6.0`.
- Odd-numbered target frameworks get trimmed on major version releases (`netstandard2.1` and `net5.0` included).

**This implies that Serilog 4.X will likely target the following TFMs: `net462;net471;netstandard2.0;net6.0`.**

----

## Reasoning

This issue replaces #1966 and comment threads such as https://github.com/serilog/serilog-sinks-console/pull/145#issuecomment-1709862581.

There's been [requests to support `net8.0` etc, inferring based on the above broad range](https://github.com/serilog/serilog-sinks-console/pull/145#issuecomment-1709862581). If we have a clear strategy in the core package and repo, the other sinks can follow. [Longer version] 

## Production assemblies

For _production_ assemblies, the following rules apply for picking TFMs, the following rules should apply:

- In general, if `netstandard2.0` is insufficient, we target the lowest even numbered `netX.0` TFM that will work. i.e. a light sink would target `net6.0` (as opposed to targeting `net12.0` just because it's there)
- non-LTS TFMs should not be targeted unless there's a reason:
  - e.g. if .NET 9.0 adds a new API that lets us do something significantly faster etc
- on a major version bump, we trim outdated non-LTS targets (e.g. if `net10.0` is out, we remove `net7.0` and/or `net9.0` targeting
- net462, net471 and netstandard <= 2.0 won't be aggressively trimmed
- `netstandard2.1` and `net5.0` will be considered as non-LTS (so will be replaced with `net6.0` on next major)

For this repo, that means V4 should:
- remove `netstandard2.1`, `net5.0` (ideally they would have been removed in V3 but the ship has sailed)
- if V `3.10` had added a `net9.0` target (because it needed a new API that was not in `net8.0`), V4 would remove that and replace it with `net10.0`

For the Console sink, the above rules would imply:
- it should target `netstandard2.0`.
- it should not target `netstandard2.1`, `netcoreapp3.1`, `net5.0`
[...]

PR #1971, opened 2023-10-21
chore: Drop test coverage for out of support .NET Core vers

bartelink:
Removes test coverage for unsupported fameworks per https://github.com/serilog/serilog/issues/1970

### DIFF
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -28,10 +28,6 @@
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net47' ">
-    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
-  </PropertyGroup>
-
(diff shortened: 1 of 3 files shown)


test/TestDummies/TestDummies.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;net462</TargetFrameworks>
```
bartelink, line 4: can't think of a specific reason a NS2.1 variant would add value? Or is the aim to pull in the NS2.1 variant of the prod assembly? (And if that's the aim, should there be a net48 one added instead of removing this?)

src/Serilog/Serilog.csproj
```diff
@@ -28,10 +28,6 @@
     <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net47' ">
```
bartelink, line 31: seems dead?

nblumhardt: Looks like it :+1:

PR #1977, opened 2023-11-10
Don't stack overflow when disposing `ReusableStringWriter`

nblumhardt:
Fixes https://github.com/serilog/serilog-sinks-console/issues/148


SeppPenner:
Thank you for the quick fix, this totally broke our system 😅

mrtristan:
commenting with keywords `StackOverflow` and `StackOverflowException` so people searching may find this like i was trying 😄

PR #1981, opened 2023-11-14
Update to .NET 8.0

eslamisepehr:
Hello 👋

Update Projects to .NET 8.0.

### DIFF
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -4,7 +4,7 @@
     <VersionPrefix>3.1.2</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net462;net471</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


src/Serilog/Serilog.csproj
```diff
@@ -4,7 +4,7 @@
     <VersionPrefix>3.1.2</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net462;net471</TargetFrameworks>
-    <TargetFrameworks>$(TargetFrameworks);netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworks);netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0;net8.0</TargetFrameworks>
```
bartelink, line 7: The main project should not need to target net8.0
(it does make sense to add it to test and/or samples though)
see https://github.com/serilog/serilog/issues/1970

nblumhardt: I think we should hold making ad hoc TFM changes for now, given the desire to review the strategy for TFMs across the board (but thanks all the same @eslamisepehr!).

PR #1999, opened 2024-01-14
target net8 to reduce assembly +pdb size

SimonCropp:
When using a polyfill, libraries should target latest TFMs. the reason for this is that no polyfills are required in the latest, therefor the resulting dll+pdb are smaller and hence faster to load+uses less memory

in this case net7 serilog dll+pdb is 206KB. in net 8 they are 185KB

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "7.0.102",
+    "version": "8.0.101",
     "allowPrerelease": false,
     "rollForward": "latestFeature"
   }
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -4,7 +4,7 @@
     <VersionPrefix>3.1.2</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net462;net471</TargetFrameworks>
-    <TargetFrameworks>$(TargetFrameworks);netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworks);netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0;net8.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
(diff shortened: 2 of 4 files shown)


SimonCropp:
IMO reduce assembly +pdb size, and associated load time an mem usage, is justification for adding a new target

bartelink:
That's not unreasonable, but perhaps with some more analysis we might be able to trim it to some degree
i.e. what's the rule that prevents that list having net7.0 through net19.0 in 10y and compiles taking ages? The prompting issue for this was people screaming for new releases of Console sink "supporting" net7.0 and net8.0 because it only "supported" net5.0 due to that being the newest TFM-specific build that happened to be included

Has there been a significant size reduction in the 7-8 timeframe? Or was it in net7.0? Do we anticipate more in net8.0?
How much of the size difference is for the PDB vs the DLL?
Does the memory usage actually change, or is it down to some format change that does not manifest at runtime (some change in padding rules?)

i.e. can we map this to a rule change such as `net6.0` + supported LTS FWs at time of release of any major version, but we trim STS versions, and unsupported LTSes (but  always leave `net6.0` and `netstandard2.0`)

bartelink:
Ha, sorry man; I've spent way too much time on this little pet peeve!

SimonCropp:
yep i can tell compile time must have greatly effected you with your significant contributions https://github.com/serilog/serilog/graphs/contributors

bartelink:
Touche! (I do have contribs in the Async Sink I wish I now wish I could uncontribute)
My daily lang is F# hence my oversensitivity to compile time as something remotely relevant....

nblumhardt:
>  net6.0 + supported LTS FWs at time of release of any major version, but we trim STS versions, and unsupported LTSes ...

... actually sounds like a pretty good rule of thumb to me; the latest LTS would always be free of polyfill bloat, and the people who care most are likely moving forward with the framework, there.

@SimonCropp we've been brewing a TFM cleanup for a while (and trying to come up with something that works reasonably consistently across various sinks, etc.) - I think what @bartelink's suggesting is that we hadn't considered this angle and might want to bake it into the scheme so that when we do the same in the other projects there's some clear explanation to link back to.

Would you be keen to reopen and work through the other details in this PR (perhaps with the v4 major bump)?

PR #2014, opened 2024-02-22
Move "missing named property" self log to respect enrichment

SimonCropp:
The current location where named properties are checked is too early. it does not respect properties provided via enrichment.
Hence it incorrectly writes `Required properties not provided for:` to the selflog on every log call

### DIFF
--- a/src/Serilog/Capturing/PropertyBinder.cs
+++ b/src/Serilog/Capturing/PropertyBinder.cs
@@ -42,9 +42,6 @@ class PropertyBinder
     {
         if (messageTemplateParameters.Length == 0)
         {
-            if (messageTemplate.NamedProperties != null || messageTemplate.PositionalProperties != null)
-                SelfLog.WriteLine("Required properties not provided for: {0}", messageTemplate);
-
             return NoProperties;
         }
 
--- a/src/Serilog/Core/Logger.cs
+++ b/src/Serilog/Core/Logger.cs
@@ -478,6 +478,19 @@ public sealed class Logger : ILogger, ILogEventSink, IDisposable
             SelfLog.WriteLine("Exception {0} caught while enriching {1} with {2}.", ex, logEvent, _enricher);
         }
 
+        var eventProperties = logEvent.Properties;
+        var namedProperties = logEvent.MessageTemplate.NamedProperties;
+        if(namedProperties != null)
+        {
+            foreach (var property in namedProperties)
+            {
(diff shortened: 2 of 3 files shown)


nblumhardt:
Hi Simon! Thanks for the PR. I think this is incorrect, though - the message template needs to be valid at the call site. If you want to output properties added by enrichment then the output template would be the place to do that.

There have been a couple of discussions around this in the past; for me the most important consideration is robustness - I don't want logging statements in my app to suddenly become invalid because enrichment was changed somewhere further down the pipeline (or when I move that code between apps).

The tooling consideration is also important, too - Rider, ReSharper, SerilogAnalyzer, and VS with plug-ins (possibly by default now, too?) will report missing parameter warnings if you write a Serilog logging statement without providing all of the properties mentioned in the template.

Is there a scenario we're not handling well using the output template mechanism?

SimonCropp:
@nblumhardt we have enrichers that add some common properties (username, application, machine name, git sha) to all messages.

and at app startup we log a message with `"Logger created for Application: '{Application}' by machine named '{Machine}'."`

we force that message to warning. so that it forces then entry to go to seq, so that the seq logging level switch is ensured to be applied

so that message infers the Application and Machine from the enriched properties.

At the moment that warning message triggers the selfliog. IMO incorrectly since the properties are correctly inserted into the message

it seems you are implying that properties that come from enrichment should not be used inside the message template. this would seem strange to me given it works seamlessly at the moment

nblumhardt:
Thanks for the follow-up.

> it seems you are implying that properties that come from enrichment should not be used inside the message template.
> this would seem strange to me given it works seamlessly at the moment

It doesn't work seamlessly, though - you'll get a warning in `SelfLog`, SerilogAnalyzer/Rider etc. will show a squiggly, and if you change your enricher setup or try to add more arguments to your logging statement, everything will fall apart:

```csharp
Log.Information("Logger created for Application: '{Application}' by machine named '{Machine}' version {Version}.", app.Version)
```

SimonCropp:
> you'll get a warning in SelfLog, SerilogAnalyzer/Rider etc. will show a squiggly

 * very few people are aware of SelfLog. none of my team knew about it. and i only did due to my contributions to serilog. I wonder if people should be required to define a SelfLog as part of config.
 * `SerilogAnalyzer/Rider etc. will show a squiggly` weird. i dont get that. perhaps something wrong with my setup
 

happy to work around by explicitly passing in the props. Will close this. Thanks so much for helping me out

PR #2016, opened 2024-02-24
Update TFMs for Serilog 4.0

nblumhardt:
Aiming for the scheme discussed in #1970; depends on #2015.

### DIFF
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -5,7 +5,7 @@
     <Authors>Serilog Contributors</Authors>
     <!-- .NET Framework version targeting is frozen at these two TFMs. -->
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net462;net471</TargetFrameworks>
-    <!-- Policy is to trim TFM-specific builds to `netstandard2.0`,
+    <!-- Policy is to trim TFM-specific builds to `netstandard2.0`, `net6.0`,
     all active LTS versions, and optionally the latest RTM version, when releasing new
     major Serilog versions. -->
     <TargetFrameworks>$(TargetFrameworks);netstandard2.0;net6.0;net8.0</TargetFrameworks>
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -3,8 +3,10 @@
     <Description>Simple .NET logging with fully-structured events</Description>
     <VersionPrefix>4.0.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
+    <!-- .NET Framework version support is frozen at these two TFMs. -->
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net462;net471</TargetFrameworks>
-    <TargetFrameworks>$(TargetFrameworks);netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0</TargetFrameworks>
+    <!-- Policy is to support `netstandard2.0`, all active LTS versions, and the latest RTM version. -->
+    <TargetFrameworks>$(TargetFrameworks);netstandard2.0;net6.0;net8.0</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>serilog;logging;semantic;structured</PackageTags>
     <PackageIcon>icon.png</PackageIcon>
@@ -34,19 +36,11 @@
 
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' " />
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.1' ">
-    <DefineConstants>$(DefineConstants);FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN;FEATURE_ITUPLE;FEATURE_ASYNCDISPOSABLE;FEATURE_DICTIONARYTRYADD</DefineConstants>
-  </PropertyGroup>
-
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net5.0' ">
-    <DefineConstants>$(DefineConstants);FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN;FEATURE_ITUPLE;FEATURE_ASYNCDISPOSABLE;FEATURE_WRITE_STRINGBUILDER;FEATURE_TOHEXSTRING;FEATURE_DICTIONARYTRYADD</DefineConstants>
-  </PropertyGroup>
-
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
     <DefineConstants>$(DefineConstants);FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN;FEATURE_ITUPLE;FEATURE_DATE_AND_TIME_ONLY;FEATURE_ASYNCDISPOSABLE;FEATURE_WRITE_STRINGBUILDER;FEATURE_TOHEXSTRING;FEATURE_DICTIONARYTRYADD</DefineConstants>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net7.0' ">
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net8.0' ">
     <DefineConstants>$(DefineConstants);FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN;FEATURE_ITUPLE;FEATURE_DATE_AND_TIME_ONLY;FEATURE_ASYNCDISPOSABLE;FEATURE_WRITE_STRINGBUILDER;FEATURE_TOHEXSTRING;FEATURE_DICTIONARYTRYADD</DefineConstants>
   </PropertyGroup>
 
@@ -69,10 +63,6 @@
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" />
-  </ItemGroup>
-
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.1' ">
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.2" />
   </ItemGroup>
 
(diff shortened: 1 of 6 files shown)


src/Serilog/Serilog.csproj
```diff
@@ -1,16 +1,17 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Description>Simple .NET logging with fully-structured events</Description>
-    <VersionPrefix>3.1.2</VersionPrefix>
+    <VersionPrefix>4.0.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
+    <!-- .NET Framework version support is frozen at these two TFMs. -->
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net462;net471</TargetFrameworks>
-    <TargetFrameworks>$(TargetFrameworks);netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0</TargetFrameworks>
+    <!-- Policy is to support `netstandard2.0`, all active LTS versions, and the latest RTM version. -->
```
bartelink, line 8: calling this 'support' is what kicked off this debate about busywork, so language is critical here
though this is a bit long...
```suggestion
    <!-- Policy is to trim TFM-specific builds to `netstandard2.0`, `net6.0` and active LTSes when releasing major versions -->
```

src/Serilog/Serilog.csproj
```diff
@@ -35,19 +36,11 @@
...
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net5.0' ">
-    <DefineConstants>$(DefineConstants);FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN;FEATURE_ITUPLE;FEATURE_ASYNCDISPOSABLE;FEATURE_WRITE_STRINGBUILDER;FEATURE_TOHEXSTRING;FEATURE_DICTIONARYTRYADD</DefineConstants>
-  </PropertyGroup>
-
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
     <DefineConstants>$(DefineConstants);FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN;FEATURE_ITUPLE;FEATURE_DATE_AND_TIME_ONLY;FEATURE_ASYNCDISPOSABLE;FEATURE_WRITE_STRINGBUILDER;FEATURE_TOHEXSTRING;FEATURE_DICTIONARYTRYADD</DefineConstants>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net7.0' ">
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net8.0' ">
```
bartelink, line 43: this flagset is identical to net6.0, which means there is no reason for it to exists save for:
- potentially different packing stuff that might make the DLL smaller (but by how much?)
- potential perf benefit because the compiler can perhaps generate slightly different code targeting new intrinsics or something else unspecific

If it was me, I wouldnt bother actually adding net8.0 here without having a reason, and "it can't hurt and the DLL seems to be 16K smaller" is a bit weak for me

But it's not me, so this is just me attempting to stir debate (and/or some might claim, troll.... :trollface: )

test/Serilog.Tests/Serilog.Tests.csproj
```diff
@@ -32,4 +35,7 @@
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net7.0' ">
     <DefineConstants>$(DefineConstants);FEATURE_DEFAULT_INTERFACE;FEATURE_SPAN;FEATURE_DATE_AND_TIME_ONLY;FEATURE_ITUPLE;FEATURE_ASYNCDISPOSABLE</DefineConstants>
```
bartelink, line 36: perhaps make a DefineConstantsGeNet60 and refer to that in the 3 supported ones?
that makes it easy to see a) if there is variance b) the specifics of that variance where one of the later ones needs a new flag due to stressing some new API specific to a later RTM

src/Serilog/Serilog.csproj
Numpsy: >  the DLL seems to be 16K smaller

On that subject, and refing #1999 - Is there actually anything in here that needs pollyfilling in .NET 6.0 anyway?
(A quick tinker with the current dev branch suggests that it builds with the PolySharp NuGet package only being referenced in the .NET FW and Standard build targets)

src/Serilog/Serilog.csproj
nblumhardt: Thanks for the review; I'll merge #2015 and address those items here 👍

src/Serilog/Serilog.csproj
nblumhardt: I think the benefit here is just having a simple policy that will work consistently across the majority of the Serilog projects and discourage projects from drifting into odd/arbitrary TFM lists.

We don't use anything .NET 8 specific in this repo (yet) but there's enough new surface are in 7 and 8 that we'll likely consume some of it at some point.

Not a strongly-held opinion, but also keen just to move on - we're hitting diminishing returns on more time investment on this I think :-)

test/Serilog.Tests/Serilog.Tests.csproj
nblumhardt: Prefer avoiding this if possible, just in favor of a simpler setup.

bartelink: Makes sense ;)

src/Serilog/Serilog.csproj
```diff
@@ -3,8 +3,12 @@
     <Description>Simple .NET logging with fully-structured events</Description>
     <VersionPrefix>4.0.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
+    <!-- .NET Framework version targeting is frozen at these two TFMs. -->
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net462;net471</TargetFrameworks>
-    <TargetFrameworks>$(TargetFrameworks);netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0</TargetFrameworks>
+    <!-- Policy is to trim TFM-specific builds to `netstandard2.0`,
```
bartelink, line 8:
```suggestion
    <!-- Policy is to trim TFM-specific builds to `netstandard2.0`, `net6.0`,
```
I think `net6.0` should also be on the 'we always compile for it' list - reasoning being that it provides a stable base.

Otherwise, when it gets dropped, you're forced to choose:
- update your minimum to `net8.0` because Serilog forced you to
- make accommodations (add shims) so you can retarget to `netstandard2.0`

Where netstandard2.0 is not feasible, it's also the fixed point of stability that NS2.0 normally is

src/Serilog/Serilog.csproj
bartelink: If adding 8.0 preemptively makes someone happy, fine by me - the main thing is that there comes a time when it gets trimmed.

But if I can (continue to be) a very pedantic evil's advocate, I'm not sure it's that great an example to be seeding things with - if other lbs take 6+8, you know that people will feel the instinct to add 10.0 to follow the pattern (which is where we started on this) - and at that point, net6.0 will be in the baseline set, and net8.0 and net10.0 will be in LTS.

Only in late 2026 do we get to net8.0 being removed, why which time the it looks like you just add LTSes to the list.

But to be fair, the policy wording is also clear, and deviations can be picked up in code review.

nblumhardt: It's tricky to balance the need for a clear/simple policy with the unpredictability of what we might want from future .NETs... I don't think it'll be a big burden to see a handful of LTSes in there - but we can reevaluate if it gets out of hand :-)

I find that if I'm not targeting the latest TFM by default while editing code, it's too easy to miss the existence of new APIs via autocomplete etc., so some amount of wanting 8 there is a quality-of-life thing.

bartelink: That's a good point I hadn't considered - the only bit that seems unresolved to me is whether `net6.0` targeting is going to be trimmed in the `net10.0` timeframe.

My general policy atm is to target `net6.0` only and be conservative in adding new TFM-specific builds until it hurts. But then my libs are more I/O bound in nature than Serilog, so it's kinda apples and oranges. Eventually removing `net6.0` from the list makes life a tiny bit more interesting than I'd like it to be. (Having said that, I'm writing libs that should prb=obably target otel, not serilog in the long term)

Can I be so rude as to page you regarding this @andrewlock ?

TL;DR the proposed policy is that on rollign the major version of Serilog going forward, that the TMFs list gets trimmed to `netstandard2.0` + the currently supported LTSes and (optionally) the latest STS if there's something relevant in there that someone has a specific use for in realising a feature (or significant perf win)

nblumhardt: Just in case you missed it @bartelink, I merged your suggestion to add `net6.0` to the always-included list alongside `netstandard2.0`.

Give it a few years and it's hard to predict what will be optimal, but I think we're on the same page RE `net6.0` being the "new `netstandard`" for the foreseeable future 👍

bartelink: Doh! Yes, my bad; I looked at it wrong, and was interpreting as if you'd opted not to have `net6.0` on the permanent list (without any specific reason).
And you're def not wrong with caution about throwing around forward looking statements!

src/Serilog/Serilog.csproj
```diff
@@ -3,8 +3,12 @@
     <Description>Simple .NET logging with fully-structured events</Description>
     <VersionPrefix>4.0.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
+    <!-- .NET Framework version targeting is frozen at these two TFMs. -->
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net462;net471</TargetFrameworks>
-    <TargetFrameworks>$(TargetFrameworks);netstandard2.1;netstandard2.0;net5.0;net6.0;net7.0</TargetFrameworks>
+    <!-- Policy is to trim TFM-specific builds to `netstandard2.0`, `net6.0`,
```
bartelink, line 8: reying to shoehorn it onto two lines:
```suggestion
    <!-- Policy is to trim TFM-specific builds when upping Serilog major version number to:
    `netstandard2.0`, `net6.0`, active LTS versions and (optionally) the latest RTM -->
```

bartelink: Either way, this all looks pretty complete and concise, and I'd say it's ready to merge.
It could be argued that a note in a wiki or the README might make sense, but I think #1970 (which this PR should resolve) and the text above should be enough - we just need a snippet at the ready saying
"discussing the TFM list has been flogged to death in #1970; policy is to drive the TFM list based on how Serilog's src/Serilog.csproj for the foreseeable" :D

nblumhardt: 👍 diminishing returns on this, we can still revisit further along in the process :-)

PR #2017, opened 2024-02-25
also target net8

SimonCropp:
since serlog uses a polyfill, all TFMs should be listed. the reason is that if a target is omitted, that target will be unnecessarily larger due to redundant polyfills.

 * net7 138kb
 * net8 129kb

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "8.0.201",
     "allowPrerelease": false,
     "rollForward": "latestFeature"
   }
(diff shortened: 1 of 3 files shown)


SimonCropp:
it also (often) has performance implications. since often polyfills do not have the same optimisations as the actual implementation, omitting a target have result in using those slower polyfill implementations over the actual implementation.

Numpsy:
[Are there any polyfills actually being used in .NET 6/7 ?](https://github.com/serilog/serilog/pull/2016#discussion_r1501389784) - i.e. is there presently any need for PolySharp to be attached to those builds in the first place?)

SimonCropp:
@Numpsy yeah u could probably condition the reference out. but you would probably eventually want to add it back in, eg when u want newer compiler features that need attributes

Numpsy:
fwiw, If i build it locally then the .NET 8 dll is 129KB, unless I add the ```<DefineConstants>``` to match .NET 6/7, at which point it's 137KB

SimonCropp:
guess i was mistaken then

PR #2018, opened 2024-03-01
Add `LogEvent.UnstableAssembleFromParts()`

### LINKED ISSUE #2010: Add `LogEvent.UnstableAssembleFromParts()`
**Is your feature request related to a problem? Please describe.**

Libraries that interop with Serilog frequently need to construct `LogEvent` instances from pre-created parts, such as `MessageTemplate`s, timestamps, and lists of properties.

For example, we have one in _Serilog.Extensions.Logging_:

https://github.com/serilog/serilog-extensions-logging/blob/dev/src/Serilog.Extensions.Logging/Extensions/Logging/SerilogLogger.cs#L165

And in _SerilogTracing_ a very similar scenario occurs:

https://github.com/serilog-tracing/serilog-tracing/blob/dev/src/SerilogTracing/Interop/ActivityConvert.cs#L83

This has a significant performance cost because the public `LogEvent` constructors require allocations for the inbound `LogEventProperty` instances, an allocation for the internal dictionary, and copying/population to occur.

**Describe the solution you'd like**

```csharp
var evt = LogEvent.UnstableAssembleFromParts(timestamp, ...);
```

This method, in particular, would directly accept a `Dictionary<string, LogEventPropertyValue> properties`, and this would be directly used to set the `_properties` field.

Since this would be for use only in some fairly sophisticated use cases, I think we could avoid having to enumerate the dictionary by requiring pre-validation of property names (just `!string.IsNullOrEmpty()` by the caller.

The name includes `Unstable` because it's expected that the signature will change as the internal structure of `LogEvent` changes, and callers would need to be updated appropriately.

**Describe alternatives you've considered**

Various weaker versions, such as accepting enumerators of `KeyValuePair<string, LogEventPropertyValue>`, which end up with only marginally better safety and substantially worse performance.

**Additional context**

Might be worth considering this a 4.0 addition, so it's clear which version callers requiring this should depend on.

nblumhardt:
Fixes #2010

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,7 @@ root = true
 
 [*]
 trim_trailing_whitespace = true
-insert_final_newline = true
+insert_final_newline = false
 indent_style = space
 indent_size = 4
 charset = utf-8
--- a/src/Serilog/Events/LogEvent.cs
+++ b/src/Serilog/Events/LogEvent.cs
@@ -226,6 +226,40 @@ public class LogEvent
         _properties.Remove(propertyName);
     }
 
+    /// <summary>
+    /// Construct a <see cref="LogEvent"/> using pre-allocated values for internal fields. Normally,
+    /// the <see cref="LogEvent"/> constructor allocates a dictionary to back <see cref="Properties"/>,
+    /// so that this is not unexpectedly shared. This is unnecessary in many integration scenarios,
+    /// leading to an additional nontrivial <see cref="Dictionary{TKey,TValue}"/> allocation. This
+    /// method allows specialized callers to avoid that overhead.
+    /// </summary>
+    /// <remarks>
+    /// Because this method exposes parameters that essentially map 1:1 with internal fields of <see cref="LogEvent"/>,
+    /// the parameter list may change across major Serilog versions.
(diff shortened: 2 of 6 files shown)


nblumhardt:
Keen to roll this through and update _Serilog.Extensions.Logging_ upstream; will hit the button tomorrow but reviews at any later point still welcome :)

.editorconfig
```diff
@@ -2,7 +2,7 @@ root = true
 
 [*]
 trim_trailing_whitespace = true
-insert_final_newline = true
+insert_final_newline = false
```
bartelink, line 5: controversial!
a stack of the other projects have it the way it was?
(While I got sold on it a few years back as being the right thing, but it still irks me to this day, and obviously some IDE / tool generated files don't respect it and hence cause null diffs all the time)

nblumhardt: In this one it's the public API surface area test, which depends on a diff. If I try to edit the "approved" side, the inserted newline causes a test failure. Irked me for too long now :-)

sungam3r: https://github.com/serilog/serilog/pull/2028

Issue #1994, opened 2024-01-03
Destructure MyDictionary as Dictionary


zhangzw218:
**Is your feature request related to a problem? Please describe.**
I had a custom dictionary that showed errors when writing ElasticSearch, and in the end I suspected that it was not an ElasticSearch problem, but a problem with the dictionary format.
ElasticSearch Error Log:
```
following: {"type":"mapper_parsing_exception","reason":"failed to parse field [fields.Dic.Value] of type [long] in document with id '0BYPz4wBG91kum5oOCHj'. Preview of field's value: 'BB'","caused_by":{"type":"illegal_argument_exception","reason":"For input string: \"BB\""}}
```
There is a custom dictionary that is parsed into a list of KeyValuePair, I want to know if I can handle it as a dictionary?
[1896](https://github.com/serilog/serilog/issues/1896)
I'm using the Abp project, which has a very broad class:
```
[Serializable]
public class ExtraPropertyDictionary : Dictionary<string, object>
{
    public ExtraPropertyDictionary()
    {
    }
}

public class SendEsModel
{
    public SendEsModel()
    {
        ExtraProperties = new Dictionary<string, object>();
        ExtraProperties2 = new ExtraPropertyDictionary();
        ExtraProperties3 = new TestDictionary();
    }
    public Dictionary<string,object> ExtraProperties { get; }
    public ExtraPropertyDictionary ExtraProperties2 { get; }
    public MyDictionary ExtraProperties3 { get; }
}

```
When I try to print the log, I get List<KeyValuePair>:
```

    public object ToEsLogError()
    {
        var dic = new SendEsModel();
        dic.ExtraProperties.Add("LongA", 1L);
        dic.ExtraProperties.Add("StringB", "BB");
        dic.ExtraProperties.Add("Name", "ExtraProperties");

        dic.ExtraProperties2.Add("LongA", 1L);
        dic.ExtraProperties2.Add("StringB", "BB");
        dic.ExtraProperties2.Add("Name", "ExtraProperties2");

        dic.ExtraProperties3.Add("LongA", 1L);
        dic.ExtraProperties3.Add("StringB", "BB");
        dic.ExtraProperties3.Add("Name", "ExtraProperties3");
        var resultObj = new
        {
            CurrentDate = DateTime.Now,
            Dictionary = dic
        };
// Dic1:{"LongA": 1, "StringB": "BB", "Name": "ExtraProperties"}
[...]

PR #2020, opened 2024-03-05
Cast a custom dictionary inheritance class MyDictionary : Dictionary<…

zhangzw218:
[1994](https://github.com/serilog/serilog/issues/1994)
I just want to solve some custom extension dictionary types that can be successfully serialized according to the dictionary.
To prevent the problem from spreading, I added a field that lets users decide for themselves which classes can be successfully identified as dictionaries


nblumhardt:
Thanks for getting in touch.

Implementing a custom `IDesctructuringPolicy` would be the way to achieve this.

Hope this helps,
Nick

nblumhardt:
(The policy can be plugged into your own code without the need for modifications in Serilog.)

zhangzw218:
Thank you very much. I did not know that IDestructuringPolicy existed before, but now I have found the corresponding location.
@nblumhardt

PR #2024, opened 2024-03-08
Add WhenNoOperator for destructuring

### LINKED ISSUE #2023: Redaction of sensitive information
We have the same issue as in https://github.com/serilog/serilog/issues/1649, therefore I'm not posting the full template here again.

We need to ensure that our generated data does not leak information. We can not do anything on `ToString` and also can't attribute it. We were able to achieve this through the destructure policy since scalar conversion does not allow for `StructureValue` to be output. Any other transformation is also always only executed if the variable was marked for destruction. But this means, that we MUST ensure that every developer uses destructuring correctly (which is nearly impossible).

I already tried the `ILogEventEnricher` route without success because of the findings that Troy also had (only `ScalarValue` available). I would also love to be able to define what `Destructuring.Default` means for specific types (i.e. being able to force it to destructure). This would be very easily resolvable in `PropertyValueConverter`. The only idea I would be able to come up is inheriting from `PropertyValueConverter` and force that version to be used by the pipleine.

We need this to happen on all sinks. We currently use the `ILogger` from `LoggerConfiguration.CreateLogger()` and not any adapter. I COULD wrap the DI registered instance to wrap it (basically introducing an adapter), handling parameters before passing, but this seems just like a big overkill and I hope there is a better way using Serilog.

The main issue is still that it's not possible to use any custom `ILogEventPropertyFactory` and/or `ILogEventPropertyValueFactory` implementation since `MessageTemplateProcessor` and `PropertyValueConverter` are internal.

Falco20019:
Fixes #2023

I created a PR to show that it is aligning with the architecture of Serilog quite well and that it only needs minimal changes.

For my use-case, not only the type but also inherited types would need to use the fallback. But I wanted to start with this for discussion to see if it get‘s interest or if it gets declined. This could easily be also given to the call as parameter.

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -40,7 +40,8 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
     readonly bool _propagateExceptions;
     readonly IDictionary<Type, Destructuring> _fallbackDestructuring;
 
-    public PropertyValueConverter(int maximumDestructuringDepth,
+    public PropertyValueConverter(
+        int maximumDestructuringDepth,
         int maximumStringLength,
         int maximumCollectionCount,
         IEnumerable<Type> additionalScalarTypes,
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -38,15 +38,16 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
     readonly int _maximumStringLength;
     readonly int _maximumCollectionCount;
     readonly bool _propagateExceptions;
+    readonly IDictionary<Type, Destructuring> _fallbackDestructuring;
 
-    public PropertyValueConverter(
-        int maximumDestructuringDepth,
+    public PropertyValueConverter(int maximumDestructuringDepth,
         int maximumStringLength,
         int maximumCollectionCount,
         IEnumerable<Type> additionalScalarTypes,
         IEnumerable<Type> additionalDictionaryTypes,
         IEnumerable<IDestructuringPolicy> additionalDestructuringPolicies,
-        bool propagateExceptions)
+        bool propagateExceptions,
+        IDictionary<Type, Destructuring> fallbackDestructuring)
     {
         Guard.AgainstNull(additionalScalarTypes);
         Guard.AgainstNull(additionalDestructuringPolicies);
@@ -79,6 +80,7 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
 
         _dictionaryTypes = additionalDictionaryTypes.ToArray();
         _depthLimiter = new(maximumDestructuringDepth, this);
+        _fallbackDestructuring = fallbackDestructuring;
     }
 
     public LogEventProperty CreateProperty(string name, object? value, bool destructureObjects = false)
@@ -123,6 +125,12 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
         if (value == null)
             return ScalarValue.Null;
 
+        var type = value.GetType();
+        if (destructuring == Destructuring.Default && _fallbackDestructuring.TryGetValue(type, out var fallbackDestructuring))
+        {
+            destructuring = fallbackDestructuring;
+        }
+
         if (destructuring == Destructuring.Stringify)
         {
             return Stringify(value);
@@ -156,7 +164,6 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
             }
         }
 
-        var type = value.GetType();
         if (TryConvertEnumerable(value, type, destructuring, out var enumerableResult))
             return enumerableResult;
 
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Collections.Concurrent;
+
 namespace Serilog.Capturing;
 
 // Values in Serilog are simplified down into a lowest-common-denominator internal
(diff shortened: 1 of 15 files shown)


test/Serilog.Tests/Serilog.Tests.csproj
```diff
@@ -20,6 +20,10 @@
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework.TrimEnd(`0123456789`))' == 'net' ">
```
Falco20019, line 23: This is from the official documentation: https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-conditions?view=vs-2022

But could also be written as `$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net5.0'))` for our use-case.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -38,6 +40,8 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
     readonly int _maximumStringLength;
     readonly int _maximumCollectionCount;
     readonly bool _propagateExceptions;
+    readonly IDictionary<Type, DestructuringFallback> _fallbackDestructuring;
+    readonly ConcurrentDictionary<Type, Destructuring?> _fallbackDestructuringCache;
```
Falco20019, line 44: I'm not sure if a concurrent dictionary is necessary or if a regular one would be sufficient. I assumed that the converter is used multi-threaded.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -169,6 +185,61 @@ LogEventPropertyValue CreatePropertyValue(object? value, Destructuring destructu
         return new ScalarValue(value.ToString() ?? "");
     }
 
+    Destructuring? GetCachedFallbackDestructuring(
+#if !NET5_0
```
Falco20019, line 189: .NET 5 had a bug that it was missing the `DynamicallyAccessedMemberTypes.Interfaces` item. But `type.GetInterfaces()` was also not using it, so here is no problem not adding it. Any other .NET version (Framework and Core) has it and would result in Roslyn analyzer issues if left away.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -169,6 +185,61 @@ LogEventPropertyValue CreatePropertyValue(object? value, Destructuring destructu
...
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.Interfaces)]
+#endif
+        Type type)
+    {
+        if (_fallbackDestructuringCache.TryGetValue(type, out var fallbackDestructuring)) return fallbackDestructuring;
+
+        var fallback = GetFallbackDestructuring(type);
+        _fallbackDestructuringCache[type] = fallback;
```
Falco20019, line 197: This is only done for performance to not resolve the whole type hierarchy on every conversion. Overwriting it even on multi-threading would not lead to race conditions as the value would also be the same.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -169,6 +185,61 @@ LogEventPropertyValue CreatePropertyValue(object? value, Destructuring destructu
...
+        return fallback;
+    }
+
+    Destructuring? GetFallbackDestructuring(
+#if !NET5_0
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.Interfaces)]
+#endif
+        Type type)
+    {
+        // Check direct hit:
```
Falco20019, line 208: This is the quick-exit and COULD be merged with the base-type checks, but I preferred to have it seperately and before checking the interfaces as it would be the best-fit and most common case.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -169,6 +185,61 @@ LogEventPropertyValue CreatePropertyValue(object? value, Destructuring destructu
...
+        var destructuringFallback = GetFallbackDestructuring(type, true);
+        if (destructuringFallback != null) return destructuringFallback;
+
+        // Check interfaces:
+        foreach (var @interface in type.GetInterfaces())
```
Falco20019, line 213: This will return all interfaces in the full hierarchy, including those from base types.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -169,6 +185,61 @@ LogEventPropertyValue CreatePropertyValue(object? value, Destructuring destructu
...
+        {
+            destructuringFallback = GetFallbackDestructuring(@interface, false);
+            if (destructuringFallback != null) return destructuringFallback;
+        }
+
+        // Check base type hierarchy:
+        var baseType = type.BaseType;
```
Falco20019, line 220: As last resort, the type hierarchy is checked upwards.

src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
```diff
@@ -203,4 +206,26 @@ public LoggerConfiguration ToMaximumCollectionCount(int maximumCollectionCount)
...
     }
+
+    /// <summary>
+    /// If no explicit destructuring hint was given for the property, use the given
+    /// destructuring as fallback.
+    /// </summary>
+    /// <param name="destructuringFallback">The fallback destructuring.</param>
+    /// <param name="type">Type to define the fallback for.</typeparam>
+    /// <returns>Configuration object allowing method chaining.</returns>
+    public LoggerConfiguration WhenNoOperator(Type type, DestructuringFallback destructuringFallback)
```
Falco20019, line 217: It would be possible to have these use `Destructuring` and `bool` directly in the user-facing API, but I just used `DestructuringFallback` as forward-compatible version.

Falco20019 (commented):
Some information about my intentions in different locations.

nblumhardt:
https://github.com/serilog/serilog/issues/2023#issuecomment-2097672305 - thanks again :+1:

PR #2027, opened 2024-03-18
Finish trim annotation

agocke:
There's one trim warning that the analyzer didn't catch (bug) that the
AOT compiler does. Unfortunately, the warning isn't easily resolvable --
it implies that you can no longer use deconstructed compiler-generated
types (like anonymous types) through the various logging calls.

There's unfortunately no way to perfectly resolve this. This PR proposes
a behavioral change for trimming -- when trimming, compiler-generated
types will be logged like scalars, instead of being deconstructed. This
is not ideal. Libraries should not have significantly different behavior
when trimming without a warning or some other diagnostic being produced.

Unfortunately, the contract implied by the existing code is much too
complicated to be encoded in the trimmer. The contract is that types do
not need to preserve reflection metadata, except for compiler-generated
types, which need public properties. However, the definition of
compiler-generated types is not well-defined. Encoding a heuristic in
Serilog is one thing -- encoding the heuristic in the .NET runtime is
another.

Given the constraints, this seems like the best option.

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -163,8 +163,10 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
         if (TryConvertValueTuple(value, type, destructuring, out var tupleResult))
             return tupleResult;
 
-        if (TryConvertCompilerGeneratedType(value, type, destructuring, out var compilerGeneratedResult))
+#pragma warning disable IL2072 // analyzer doesn't support feature switches yet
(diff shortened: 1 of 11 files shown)


nblumhardt:
Thanks for forging ahead with this, @agocke; sorry it's taken so long to get eyes on it! Appreciate the effort - it's a rather nasty area to dig into, having you look into this helps enormously.

`dev` is now "v4" so it seems like a good point in time to introduce the switch :+1:

agocke:
I should note one more option -- I didn't realize how popular anonymous types are in C#, and this implementation kind-of silently breaks that scenario. An alternative would be to throw a NotSupportedException or similar in the case that you try to serialize an anonymous type. That would make it clear to the user that this case isn't supported.

nblumhardt:
See also #2068, switch name has changed slightly.

PR #2029, opened 2024-03-19
Remove null! usages

sungam3r:
They were introduced in #1955.


nblumhardt:
Thanks for taking a look at this @sungam3r; hope all is well with you. Reviews are a little behind on this repo currently :-)

These are suppressions for the Rider/ReSharper "Expression is always true or false according to nullable reference types annotations" inspection. We could use a comment for the suppression, but this is really noisy at each point we do it, or switch the inspection off at the file level, but this makes it less obvious that the checks are safety checks for non-NRT-checking callers.

nblumhardt:
Not planning to merge this one. Thanks again!

PR #2030, opened 2024-03-19
Use modern array expression

sungam3r:


nblumhardt:
Apologies, this change ended up piggy-backing on #2053, missed the open PR here. Thanks all the same!

PR #2034, opened 2024-03-22
add ITypeDestructuringPolicy

SimonCropp:
IMO it is better to do the type checking once and cache it

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -159,27 +159,30 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
         var type = value.GetType();
         if (destructuring == Destructuring.Destructure)
         {
-            var typeDestructuringPolicy = _typeDestructurers.GetOrAdd(type, _ =>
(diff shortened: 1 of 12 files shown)


nblumhardt:
Would be great to bench this.. `ConcurrentDictionary` used to be horribly slow, but recent versions sound much better, might be a nice win, though I suspect the results will depend heavily on the number of policies.

Assuming the results came out good, what do you think about skipping the new interface and just making this the way we handle `IDestructuringPolicy`? I.e. a successful `TryDestructure(o)` would imply `CanDestructure(o.Type)` and the first successful policy would end up in the dictionary and consulted exclusively on the next occurrence of that type.

I haven't seen too many destructuring policies that are conditional on value rather than type, but I'd guess they exist, or we'd just do this by default... Putting `[CacheableByType]` or something like that on the policy implementation, as an opt-in marker, would make the contract explicit :thinking:

nblumhardt:
Interesting in exploring this more if it is still on your radar, but closing as stale for now, need to keep the number of open PRs down to avoid long triage passes :-) Hope all's going well 👋

SimonCropp:
@nblumhardt ❤️ no worries at all

PR #2035, opened 2024-03-24
Add support for multi-dimensional arrays

### LINKED ISSUE #2019: Logging multidimensional 2d array fails silently
This fails silently:

```csharp
var a = new object[1, 1]{ { "a" } };
Log.Error("{@Value}", a);
```

Is this expected behavior? I would expect this:
`[["a"]]`

sungam3r:
fixes #2019

### DIFF
--- a/src/Serilog/Capturing/PropertyValueConverter.cs
+++ b/src/Serilog/Capturing/PropertyValueConverter.cs
@@ -216,26 +216,37 @@ partial class PropertyValueConverter : ILogEventPropertyFactory, ILogEventProper
                 }
                 else if (list is Array a && a.Rank > 1)
                 {
-                    int index = 0;
-                    var array = new LogEventPropertyValue[list.Count];
+                    var rows = new LogEventPropertyValue[a.GetLength(0)];
                     if (a.Rank == 2)
                     {
-                        for (int i = 0; i < a.GetLength(0); ++i)
-                            for (int j = 0; j < a.GetLength(1); ++j)
-                                array[index++] = _depthLimiter.CreatePropertyValue(a.GetValue(i, j), destructuring);
+                        for (int i = 0; i < rows.Length; ++i)
+                        {
+                            var columns = new LogEventPropertyValue[a.GetLength(1)];
+                            for (int j = 0; j < columns.Length; ++j)
+                                columns[j] = _depthLimiter.CreatePropertyValue(a.GetValue(i, j), destructuring);
+                            rows[i] = new SequenceValue(columns);
+                        }
                     }
                     else if (a.Rank == 3)
                     {
-                        for (int i = 0; i < a.GetLength(0); ++i)
-                            for (int j = 0; j < a.GetLength(1); ++j)
-                                for (int k = 0; k < a.GetLength(2); ++k)
-                                    array[index++] = _depthLimiter.CreatePropertyValue(a.GetValue(i, j, k), destructuring);
+                        for (int i = 0; i < rows.Length; ++i)
+                        {
+                            var columns = new LogEventPropertyValue[a.GetLength(1)];
+                            for (int j = 0; j < columns.Length; ++j)
+                            {
+                                var heights = new LogEventPropertyValue[a.GetLength(2)];
+                                for (int k = 0; k < heights.Length; ++k)
+                                    heights[k] = _depthLimiter.CreatePropertyValue(a.GetValue(i, j, k), destructuring);
+                                columns[j] = new SequenceValue(heights);
+                            }
+                            rows[i] = new SequenceValue(columns);
+                        }
                     }
                     else
                     {
                         throw new NotSupportedException("Serilog does not support multi-dimensional arrays with Rank > 3.");
                     }
-                    result = new SequenceValue(array);
+                    result = new SequenceValue(rows);
                 }
                 else
                 {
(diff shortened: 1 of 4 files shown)


src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -214,6 +214,29 @@ bool TryConvertEnumerable(object value, Type type, Destructuring destructuring,
                 {
                     result = SequenceValue.Empty;
                 }
+                else if (list is Array a && a.Rank > 1)
+                {
+                    int index = 0;
+                    var array = new LogEventPropertyValue[list.Count];
+                    if (a.Rank == 2)
+                    {
+                        for (int i = 0; i < a.GetLength(0); ++i)
```
sungam3r, line 223: I think JIT is smart enough in 2024 to inline all `GetLength` calls.

test/Serilog.Tests/Serilog.Tests.csproj
```diff
@@ -21,6 +21,7 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3" PrivateAssets="all" />
     <PackageReference Include="xunit" Version="2.4.2" />
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+    <PackageReference Include="Shouldly" Version="4.1.0" />
```
nblumhardt, line 24: Keen not to introduce a new inconsistent set of assertions into this project.

test/Serilog.Tests/Core/LoggerTests.cs
```diff
@@ -317,6 +318,54 @@ public void NullMessageTemplateParametersDoNotBreakBinding()
...
+        var evt = DelegatingSink.GetLogEvent(l =>
+        {
+            var a = new object[3, 2] { { "a", "b" }, { "c", "d" }, { "e", "f" } };
+            l.Error("{@Value}", a);
+        });
+
+        evt.Properties.Count.ShouldBe(1);
+        var arr = evt.Properties["Value"].ShouldBeOfType<SequenceValue>();
+        arr.Elements.Count.ShouldBe(6);
+        arr.LiteralValue().ShouldBe("[a,b,c,d,e,f]");
```
nblumhardt, line 334: Would expect `[[a,b],[c,d],[e,f]]` here I think?

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -214,6 +214,29 @@ bool TryConvertEnumerable(object value, Type type, Destructuring destructuring,
                 {
                     result = SequenceValue.Empty;
                 }
+                else if (list is Array a && a.Rank > 1)
```
nblumhardt, line 217: Perhaps this could be plugged in as an `IScalarConversionPolicy` instead? Since those are checked before this code is reached, we could avoid the extra condition here. We do this already to specialize conversion of byte arrays.

nblumhardt (commented):
Thanks for sending this. Keen to look at the `IScalarConversionPolicy` route, since it would keep some complexity out of the (already a bit complex) `PropertyValueConverter` class.

test/Serilog.Tests/Core/LoggerTests.cs
sungam3r: I did it intentionally since any array is just a sequence in terms of serilog regardless dimensions. Nevertheless I can change implementation to emit sequence of sequences if you like.

src/Serilog/Capturing/PropertyValueConverter.cs
sungam3r: `IScalarConversionPolicy` has no access to
1. _depthLimiter (can be solved)
2. `destructuring` boolean argument

test/Serilog.Tests/Serilog.Tests.csproj
sungam3r: done

test/Serilog.Tests/Core/LoggerTests.cs
sungam3r: done

src/Serilog/Capturing/PropertyValueConverter.cs
sungam3r: In my opinion, we should not use `IScalarConversionPolicy` here, because the determining factor whether or not using `IScalarConversionPolicy` is the type of object inside the array, regardless of the dimensions of the array.

src/Serilog/Capturing/PropertyValueConverter.cs
```diff
@@ -214,6 +214,40 @@ bool TryConvertEnumerable(object value, Type type, Destructuring destructuring,
...
+                                for (int k = 0; k < heights.Length; ++k)
+                                    heights[k] = _depthLimiter.CreatePropertyValue(a.GetValue(i, j, k), destructuring);
+                                columns[j] = new SequenceValue(heights);
+                            }
+                            rows[i] = new SequenceValue(columns);
+                        }
+                    }
+                    else
+                    {
+                        throw new NotSupportedException("Serilog does not support multi-dimensional arrays with Rank > 3.");
```
nblumhardt, line 247: Since this is not a usage error but just a limitation, perhaps we should just use

```
result null;
return false;
```

here?

nblumhardt: Think this is a good candidate to get into 4.0; shall I make these final tweaks and merge this, @sungam3r?

WAcry: Thank you for the work on supporting multi-dimensional arrays. I hope this feature can be merged soon.

I believe there’s no reason not to support higher-dimensional arrays. Based on this branch, I have further improved the implementation to allow support for arrays with more dimensions: https://github.com/serilog/serilog/pull/2132

cc @sungam3r @nblumhardt

PR #2037, opened 2024-03-31
Fix the Serilog Unit Test can fail randomly failing and Testing on more Runtimes

### LINKED ISSUE #2036: Serilog Unit Test can fail randomly and inspected because of mismatch on the TargetFramework and compiler directives
**Description**

The main Serilog Project is been build using [_net462;**net471**;netstandard2.0;net6.0;net8.0_ ](https://github.com/serilog/serilog/blob/4bd40f1241057ebeed2062811d4990236be48ba5/src/Serilog/Serilog.csproj#L7C60-L7C73) runtimes, but the test project is testing using [_net462;**net48**;netstandard2.0;net6.0;net8.0_ ](https://github.com/serilog/serilog/blob/4bd40f1241057ebeed2062811d4990236be48ba5/test/Serilog.Tests/Serilog.Tests.csproj#L4) runtimes, and some tests that depends on `ITuple` and in a compiler directives for this feature is not been set correctly.
This is making a test fail on Visual Studio randomly. 

**Expected behavior**

Unit Tests should be testing the same set or a bigger set of .NET Runtimes than the main project is been build. 
And testing the same featues and compiler directives that the main project is using.

Current `dev` branch. `net471` and `net48` are affected.

rafaelsc:
- Add missing framework from test and Compiler Constants

Fix #2036


bartelink:
What's the motivation behind the `net471`? It'd be good to have a clean one liner that explains how one decides what goes on the list - i.e. if there is a nnew netfw, do we add that to the matrix each time? what is the concern?

rafaelsc:
For some very odd reasons, probally a bug in VS2022 Preview _(that I'm using)_, and very hard to reproduce.

Using release and do some clean and rebuild on the solution and running the test after sometimes this test fail. 
image

This does not happens when using the `buid.ps1`

This 1 test (`PropertyValueConverterTests.EightPlusValueTupleElementsAreSupportedForCapturing()`) that changes the assertion depending on the framework version set by the compiler directive `FEATURE_ITUPLE`.

https://github.com/serilog/serilog/blob/4bd40f1241057ebeed2062811d4990236be48ba5/test/Serilog.Tests/Capturing/PropertyValueConverterTests.cs#L379

But I tested this change on this PR adding one more target framework and the compiler directive `FEATURE_ITUPLE`. 
Worked yesterday, but not today. And still failing randomly.
image

I belive that some times the test is using the `netstandard2` DLL insted of the `net471` DLL. 
The netstandard2 code does not compile with  `FEATURE_ITUPLE`, and the test have  `FEATURE_ITUPLE` doing the wrong assertions on that test.
I will look more this later.

rafaelsc:
When this issue happens the test code for net48 is loafing the net462 DLL. The bug is on VS2022 or MSBuild. 

This Issue and PR can be closed.

nblumhardt:
Thanks for closing the loop @rafaelsc :+1:

bartelink:
Thanks for taking the time to leave a record; much appreciated - hopefully nobody else will actually suffer having to chase down some future similar case like you did!

PR #2038, opened 2024-04-02
Performance Improvements - `CreateLogger()` and Logging Initialization

rafaelsc:
Changes:

- Lazy Initialization of the Configuration internals List's and Set's.
- Reuse the instances of List's and Set's Created on LoggerConfiguration after Logger creation.
- Lazy Inicialization of the Logger*Configuration.
- Avoid LINQ.
- Add Internal Empty Sink for Reduce Allocations and use the `EmptySink` when no sink was set.
- Avoid array alocation with params modifier. (**Public API Changes, adding a new Overload**)
- Prefer Collection Expression.
- Add a new Benchmark to test the `CreateLogger()` and the Serilog _initialization time_ and _memory usage_.

PR #2039, opened 2024-04-02
Improve Benchmark - Avoid Benchmark.Net `[GlobalSetup]` when is not needed.

rafaelsc:
- Use class `constructor` instead of Benchmark.Net `[GlobalSetup]`.
- Use C# collection expression. (Code will be easy to read and most of time gives a better performance)
- Remove needed `null!` after changing to class `constructor`. 
- Move some value initialization to static constructor.


nblumhardt:

PR #2040, opened 2024-04-02
Improve Benchmark - More Allocations benchmarks

rafaelsc:
- Add `AllocationsIgnoringEventsBenchmark` to check allocations when the LogEvent is been ignored.
- Improve `AllocationsBenchmark` with more combinations of types, and testing more log write overloads.


nblumhardt:

PR #2042, opened 2024-04-02
Improve Benchmark - Message template rendering benchmark

rafaelsc:
- Add more Benchmarks for many MessageTemplateParser, testing many other parts of the code.
- Refactor fields to be consistent to other Benchmarks


nblumhardt:

PR #2044, opened 2024-04-04
Build updates for compatibility

tillig:
These were issues I had to fix to get Serilog building/testing on a Mac as part of the benchmarking for #1944.

- The `.vscode` folder isn't a per-user setting the way `.vs` is. `.vscode` is how you provide a default build/test command for VS Code folks. I added that support so things like "Open Repo in VS Code" or other contributors can get going a little faster.
- The Powershell build script would `Push-Location` at the start but never `Pop-Location`, so you'd end up with a stack of history entries. Added a try/finally so at the end - failure or success - it will still pop back to where you were.
- The local branch build auto-generated semver in the Powershell build didn't like branches like `feature/xyz` with slashes. Added a fix for that so it'll replace slash with dash.
- VS Code notoriously uses the _first TFM_ listed in the .csproj file as the one for Intellisense and in-editor analyzers and such. Reordering the TFMs in the .csproj files to go newest-to-oldest instead of oldest-to-newest fixes some problems with Intellisense and contributing.
- The benchmarks in the `SourceContextMatchBenchmark` fixture caused some problems due to the targeting specifically of `netcoreapp2.1` and `netcoreapp3.1`. During the execution of the benchmarks, BenchmarkDotNet auto-generates an assembly that targets these frameworks so they can be run. This auto-generated assembly has `TreatWarningsAsErrors` turned on. In .NET 8, you get a NuGet warning when you try to target `netcoreapp2.1` because `Microsoft.NETCore.App` v2.1.0 has known security issues. All of this adds up to the auto-generated assembly not compiling and the benchmarks not running. Even if you get the `netcoreapp3.1` building, you still get issues because `System.Diagnostics.DiagnosticSource` writes warnings about how it hasn't been tested and isn't supported against `netcoreapp3.1`. The fix for this is to:
  - Target newer frameworks in the `SourceContextMatchBenchmark` so you don't get the security warnings.
  - Add the targeted frameworks to the benchmark project so you can catch any build warnings for that target framework _up front_ before the auto-generated assembly gets created.

I'm fine rolling any or all of this back, but this is what it took for me, on a Mac, to get this all working smoothly. Hopefully it can help other devs/contributors to get on board faster.


nblumhardt (approved):
Lovely, thank you 👍

PR #2045, opened 2024-04-05
Level override check is case-insensitive.

### LINKED ISSUE #1944: Only case sensitive check for minimumLevel override

tillig:
Fix #1944.

Changes `LevelOverrideMap.GetEffectiveLevel()` to use a case-insensitive comparison for context names, correcting some challenges with setting overrides through mechanisms like environment variables where ALL_CAPS may be used.

### DIFF
--- a/src/Serilog/Core/LevelOverrideMap.cs
+++ b/src/Serilog/Core/LevelOverrideMap.cs
@@ -68,7 +68,7 @@ class LevelOverrideMap
     {
         foreach (var levelOverride in _overrides)
         {
-            if (context.StartsWith(levelOverride.Context) &&
+            if (context.StartsWith(levelOverride.Context, StringComparison.OrdinalIgnoreCase) &&
                 (context.Length == levelOverride.Context.Length || context[levelOverride.Context.Length] == '.'))
             {
                 minimumLevel = LevelAlias.Minimum;
--- a/src/Serilog/Core/LevelOverrideMap.cs
+++ b/src/Serilog/Core/LevelOverrideMap.cs
@@ -68,7 +68,11 @@ class LevelOverrideMap
     {
         foreach (var levelOverride in _overrides)
         {
-            if ((context.Length == levelOverride.Context.Length || context[levelOverride.Context.Length] == '.') &&
+            if (
+                (
+                    context.Length == levelOverride.Context.Length ||
+                    (context.Length > levelOverride.Context.Length && context[levelOverride.Context.Length] == '.')
+                ) &&
                 context.StartsWith(levelOverride.Context, StringComparison.OrdinalIgnoreCase))
             {
                 minimumLevel = LevelAlias.Minimum;
--- a/src/Serilog/Core/LevelOverrideMap.cs
+++ b/src/Serilog/Core/LevelOverrideMap.cs
@@ -68,8 +68,8 @@ class LevelOverrideMap
     {
         foreach (var levelOverride in _overrides)
         {
-            if (context.StartsWith(levelOverride.Context, StringComparison.OrdinalIgnoreCase) &&
-                (context.Length == levelOverride.Context.Length || context[levelOverride.Context.Length] == '.'))
+            if ((context.Length == levelOverride.Context.Length || context[levelOverride.Context.Length] == '.') &&
+                context.StartsWith(levelOverride.Context, StringComparison.OrdinalIgnoreCase))
             {
                 minimumLevel = LevelAlias.Minimum;
                 levelSwitch = levelOverride.LevelSwitch;
--- a/test/Serilog.Tests/Core/LevelOverrideMapTests.cs
+++ b/test/Serilog.Tests/Core/LevelOverrideMapTests.cs
@@ -12,6 +12,9 @@ public class LevelOverrideMapTests
     [InlineData("MyApp.Api.Controllers.AboutController", true, Information)]
     [InlineData("MyApp.Api.Controllers.HomeController", true, Warning)]
     [InlineData("Api.Controllers.HomeController", false, LevelAlias.Minimum)]
+    [InlineData("serilog", false, LevelAlias.Minimum)]
+    [InlineData("myapp", true, Debug)]
+    [InlineData("myappsomething", false, LevelAlias.Minimum)]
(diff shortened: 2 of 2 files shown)


nblumhardt:
@serilog/maintainers I'm on board with this one but suspect it needs another set of eyes on it, if just to spot other downstream places this could break. I'll merge in a week otherwise :-)

Thanks @tillig!

src/Serilog/Core/LevelOverrideMap.cs
```diff
@@ -68,7 +68,7 @@ public LevelOverride(string context, LoggingLevelSwitch levelSwitch)
     {
         foreach (var levelOverride in _overrides)
         {
-            if (context.StartsWith(levelOverride.Context) &&
+            if (context.StartsWith(levelOverride.Context, StringComparison.OrdinalIgnoreCase) &&
                 (context.Length == levelOverride.Context.Length || context[levelOverride.Context.Length] == '.'))
```
bartelink, line 72: Surely the second part of this condition is more selective and cheaper to compute ?
i.e.
            var loLen = levelOverride.Context.Length;
            if ((context.Length == loLen || (context.Length > loLen && context[loLen] == '.'))
                && context.StartsWith(levelOverride.Context, StringComparison.OrdinalIgnoreCase))
                
(my C# is too rusty to convert that to an `is` expression here!)

nblumhardt: Oh, yes, good call @bartelink. The clause would need a slight modification:

```csharp
            if (context.Length >= levelOverride.Context.Length &&
                (context.Length == levelOverride.Context.Length || context[levelOverride.Context.Length] == '.') &&
                context.StartsWith(levelOverride.Context, StringComparison.OrdinalIgnoreCase))
```

@tillig I wonder if this would be enough to claw back a few % of the performance regression related to the case-sensitivity check? I'll dig in and take a look when I have a chance, if you don't beat me to it :-)

tillig: I'm out of town for a week or so, so it'll be a bit before I'm back here. Happy to make the change, just will be a bit.

tillig: I'm back and got this. Sorry for the delay.

tillig: Hmmm. It appears swapping those has started causing tests to fail. I'll have to look into it.

tillig: OK, I'm dumb, I missed the thing @nblumhardt mentioned.

I split the `==` and `>` so in the full match scenario it should be just slightly faster since it won't have to check for equality again. I'm sure it's on the order of microseconds, but 🤷 .

nblumhardt (approved):

nblumhardt:
Thank you @tillig! 👍

src/Serilog/Core/LevelOverrideMap.cs
```diff
@@ -68,8 +68,12 @@ public LevelOverride(string context, LoggingLevelSwitch levelSwitch)
     {
         foreach (var levelOverride in _overrides)
         {
-            if (context.StartsWith(levelOverride.Context) &&
-                (context.Length == levelOverride.Context.Length || context[levelOverride.Context.Length] == '.'))
+            if (
+                (
+                    context.Length == levelOverride.Context.Length ||
```
bartelink, line 73: Sorry for the necroposting on something that's already overkill, but I will forever be pondering whether using a String.Equals into the same length case will be cheaper, or whether `StartsWith` internally special cases this...

khellang: This is what the current `StartsWith` implementation does (when passing `OrdinalIgnoreCase`):

```csharp
case StringComparison.OrdinalIgnoreCase:
    if (this.Length < value.Length)
    {
        return false;
    }
    return Ordinal.EqualsIgnoreCase(ref this.GetRawStringData(), ref value.GetRawStringData(), value.Length);
```

And this is (effectively, I inlined a method call for brevity) what `Equals` does:

```csharp
case StringComparison.OrdinalIgnoreCase:
    if (a.Length != b.Length)
    {
        return false;
    }

    return Ordinal.EqualsIgnoreCase(ref a.GetRawStringData(), ref b.GetRawStringData(), b.Length);
```

In the case where the length is the same, they both call into the same machinery with the same length, so I think they should be equal here.

bartelink: Thanks Kristian, this has made my day (especially for saving Travis the lookup from an annoying drive by commenter!)

PR #2060, opened 2024-05-10
Redesign `LoggerSinkConfiguration.Wrap()`, and introduce `LoggerSinkConfiguration.CreateSink()`

nblumhardt:
Serilog sinks push for factory methods for sink construction and composition (`WriteTo.File()`), rather than an "OO" style of public sink types and constructors (`new FileSink(...)`).

The existing `LoggerSinkConfiguration.Wrap()` method is a helper used by wrapping sink implementations, such as _Serilog.Sinks.Async_, and the core `WriteTo.Conditional()` method, to enclose one or more sinks in a "wrapper" that modifies their behavior in some way.

The existing `Wrap()` method has three roles: providing access to the enclosed sinks created via their factory methods, delegating `IDisposable.Dispose()` to the enclosed sinks so that wrappers don't need to implement it explicitly, and applying a `LoggingLevelSwitch`/minimum level to the completed result.

This PR pulls out the two main responsibilities, so that it's now possible to create an `ILogEventSink` without constructing the full logger pipeline (previously only possible using a hack based on the `Wrap()` call), and it's separately possible to wrap a given sink instance.

Creating sinks is performed using `LoggerSinkConfiguration.CreateSink()`:

```csharp
var sink = new LoggerSinkConfiguration.CreateSink(wt => wt.File(...));
// `sink` is now a `FileSink` or similar
```

Wrapping uses a new overload of `LoggerSinkConfiguration.Wrap()`:

```csharp
var wrapper = LoggerSinkConfiguration.Wrap(
    enclosed => new Wrapper(enclosed),
    wt => wt.File(...));
```

Although it would be possible to simplify this further into `Wrap(sink, wrapper)`, i.e. fully-construct both sink and wrapper, and then just perform disposal delegation in `Wrap()`:

 * the equal types of the two sink arguments would make usage error-prone,
 * almost all usage needs to apply a `WriteTo` callback anyway, and
 * it's possible to use pre-constructed sinks and wrappers with this API if necessary.

Finally, the `LoggingLevelSwitch` and minimum level responsibilities are now handed off to `WriteTo.Sink()`. Since `CreateSink()` and `Wrap()` just return `ILogEventSink`, there's no need for either of those to have any interaction with levelling.

Why make these changes now? With the introduction of `WriteTo.Batched()`, code that previously relied on constructing `PeriodicBatchingSink` will be ported over to it. Unlike the older API, no sink class is exposed - batching is only available through the factory method. This makes porting nontrivial usage of PBS harder. `CreateSink()` fixes this.

### DIFF
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -340,7 +340,7 @@ public class LoggerSinkConfiguration
         configure(capturingLoggerSinkConfiguration);
 
         return sinksToWrap.Count == 1 ?
-            sinksToWrap.Single() :
+            sinksToWrap[0] :
             new DisposingAggregateSink(sinksToWrap);
     }
 }
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -246,12 +246,12 @@ public class LoggerSinkConfiguration
         Guard.AgainstNull(configureSink);
 
         // Level aliases and so on don't need to be accepted here; if the user wants both a condition and leveling, they
-        // can specify `restrictedToMinimumLevel` etc in the wrapped sink configuration.
-        return Wrap(this, s => new ConditionalSink(s, condition), configureSink);
+        // can specify `restrictedToMinimumLevel` etc. in the wrapped sink configuration.
+        return Sink(Wrap(s => new ConditionalSink(s, condition), configureSink));
     }
 
     /// <summary>
-    /// Helper method for wrapping sinks.
+    /// Helper method for constructing wrapper sinks.
     /// </summary>
     /// <param name="loggerSinkConfiguration">The parent sink configuration.</param>
     /// <param name="wrapSink">A function that allows for wrapping <see cref="ILogEventSink"/>s
@@ -265,6 +265,7 @@ public class LoggerSinkConfiguration
     /// <exception cref="ArgumentNullException">When <paramref name="loggerSinkConfiguration"/> is <code>null</code></exception>
     /// <exception cref="ArgumentNullException">When <paramref name="wrapSink"/> is <code>null</code></exception>
     /// <exception cref="ArgumentNullException">When <paramref name="configureWrappedSink"/> is <code>null</code></exception>
+    [Obsolete("Use the two-argument `Wrap()` overload to construct a wrapper, then use `WriteTo.Sink()` to add it to the configuration.")]
     public static LoggerConfiguration Wrap(
         LoggerSinkConfiguration loggerSinkConfiguration,
         Func<ILogEventSink, ILogEventSink> wrapSink,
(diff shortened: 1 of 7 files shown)


src/Serilog/Configuration/LoggerSinkConfiguration.cs
```diff
@@ -310,6 +311,36 @@ or IAsyncDisposable
...
+            sinksToWrap.Add);
+
+        // `WriteTo.Sink()` will return the capturing configuration; this ensures chained `WriteTo` gets back
+        // to the capturing sink configuration, enabling `WriteTo.X().WriteTo.Y()`.
+        capturingConfiguration.WriteTo = capturingLoggerSinkConfiguration;
+
+        configure(capturingLoggerSinkConfiguration);
+
+        return sinksToWrap.Count == 1 ?
+            sinksToWrap.Single() :
```
bartelink, line 343: perfectly cromulent code but the use of `Single`, rather than conveying intention is just distracting IMO
```suggestion
            sinksToWrap[0] :
```

test/Serilog.Tests/Configuration/LoggerSinkConfigurationTests.cs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    [Fact]
+    public void WrappingDoesNotPermitEnrichment()
+    {
+        var enclosed = new CollectingSink();
+        var propertyName = Some.String();
+        var wrapper = LoggerSinkConfiguration.Wrap(
+            s => new DelegatingSink(s.Emit),
+            w => w
```
bartelink, line 67:
```suggestion
            wt => wt
```
(isnt that the conventional arg name?

test/Serilog.Tests/Core/AuditSinkTests.cs
```diff
@@ -54,8 +54,6 @@ public void ExceptionsThrownByPropertyAccessorsAreNotPropagated()
             .CreateLogger();
 
         logger.Information("{@Value}", new ThrowingProperty());
```
bartelink, line 56:
```suggestion
        logger.Information("{@Value}", new ThrowingProperty());
        // (the implicit assertion is that the exception is handled and does not propagate)
```

test/Serilog.Tests/Configuration/LoggerSinkConfigurationTests.cs
```diff
@@ -0,0 +1,77 @@
...
+                .Sink(enclosed)
+                .Enrich.WithProperty(propertyName, 1));
```
bartelink, line 69: Is there any way to set a tripwire for this? While it seems obvious, I would bet money I've spent time expecting this to work late some evening! (obv that would have downsides too)

nblumhardt: I think it should be possible. After the creation/wrapping callback runs, some internal assertion could be implemented through the `LoggerConfiguration`-related types. May be quite a bit of code, perhaps out of scope for this PR, if consider this one to be a new face on existing functionality?

Issue #1384, opened 2019-11-12
Allow `.` and potentially other non-alphanumeric characters in message template property names


sungam3r:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [x] Bug
- [ ] New Feature

**What version of Serilog is affected?  Please list the related NuGet package.**

2.8.0

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [x] netCore 2.0
- [ ] netCore 1.0
- [ ] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**

```c#
Log.Information("Error {Rpc.Error} occured", "boom");
```
renders to `Error {Rpc.Error} occured` 

**Please describe the expected behavior?**

`Error "boom" occured`

-----
Relates to #404

I discovered that this is due to `IsValidInPropertyName` method inside `ParsePropertyToken`:
```c#
for (var i = 0; i < propertyName.Length; ++i)
{
      var c = propertyName[i];
      if (!IsValidInPropertyName(c))   <--- ???
           return new TextToken(rawText, first);
}
```

I originally used `LogContext.PushProperty("Rpc.Error", "boom")` and everything worked fine. I am interested in the question - why, when parsing a message template, knowingly narrow the set of valid characters of property name if the neighboring api (`Push`) allows you to add properties with any characters? This asymmetry led to an error when I moved the property name inside the format string and began to pass the value of the property in the params.

I read #404, but I don’t understand the motivation.
First, this is no structured data, just plain string.

> Dots within property names themselves aren't supported by many target sinks, so we chose the lowest common denominator here.

Second, I use Console and Kafka sink which both support dots in property names. However, this talk is not even about the dots. The template string does not work if you use any character other than an alphanumeric character or underscore - `:`, `-`, etc. What problem do you solve when you talk about _lowest common denominator_? And here I come back to the fact that `Push` allows you to pass an arbitrary property name (and that's cool). Therefore, as I see it, all this concern for the lowest common denominator does not make sense.

nblumhardt:
Thanks for the note!

[...]

PR #2063, opened 2024-05-17
`MessageTemplateParser` refactor incl. experiemental `.` support

nblumhardt:
This PR came out of, and includes, #1384. You can read about the reasons for considering `.` in message template property names, and the various caveats/disadvantages on that thread. The PR enables wider investigation and experimentation, and there is currently no guarantee that this will ever become a first-class, supported Serilog feature.

By setting the experimental and **unsupported** `AppContext` switch at program start-up:

```csharp
AppContext.SetSwitch("Serilog.Parsing.MessageTemplateParser.AcceptDottedPropertyNames", true)
```

Placeholders such as `{http.request.method}` will (should!) be accepted in message and output templates. Previously, these templates would have been regarded as invalid.

Dotted placeholders will be interpreted as literal property names, e.g. `"http.request.method"`, and not structured object paths like `{http: {request: {method: ...}}}`.

In the process of investigating this I cleaned up some message template parsing tests; these are _so old_ and could use a lot more work, especially translating long multi-assert tests into data-driven `[Theory]`s.

`MessageTemplateParser` itself has also been unchanged for a long time. It could benefit from porting to a `Span<T>`-based model, but even improving our use of vanilla `string` methods yielded a significant performance boost. Long runs of literal text (very common in real-world logs) now allocate dramatically less temporary memory during parsing.

**Before:**

```
BenchmarkDotNet=v0.13.5, OS=macOS 14.4.1 (23E224) [Darwin 23.4.0]
Apple M3 Pro, 1 CPU, 11 logical and 11 physical cores
.NET SDK=8.0.204
  [Host]     : .NET 8.0.4 (8.0.424.16909), Arm64 RyuJIT AdvSIMD
  DefaultJob : .NET 8.0.4 (8.0.424.16909), Arm64 RyuJIT AdvSIMD

|                       Method |      Mean |    Error |   StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|----------------------------- |----------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
|                EmptyTemplate |  46.98 ns | 0.175 ns | 0.164 ns |  1.00 |    0.00 | 0.0268 |     224 B |        1.00 |
| DefaultConsoleOutputTemplate | 475.56 ns | 3.038 ns | 2.842 ns | 10.12 |    0.08 | 0.2575 |    2160 B |        9.64 |
|              LiteralTextOnly | 171.53 ns | 0.223 ns | 0.209 ns |  3.65 |    0.01 | 0.1194 |    1000 B |        4.46 |
```

**After:**

```
BenchmarkDotNet=v0.13.5, OS=macOS 14.4.1 (23E224) [Darwin 23.4.0]
Apple M3 Pro, 1 CPU, 11 logical and 11 physical cores
.NET SDK=8.0.204
  [Host]     : .NET 8.0.4 (8.0.424.16909), Arm64 RyuJIT AdvSIMD
  DefaultJob : .NET 8.0.4 (8.0.424.16909), Arm64 RyuJIT AdvSIMD


|                       Method |      Mean |    Error |   StdDev | Ratio | RatioSD |   Gen0 |   Gen1 | Allocated | Alloc Ratio |
|----------------------------- |----------:|---------:|---------:|------:|--------:|-------:|-------:|----------:|------------:|
|                EmptyTemplate |  48.35 ns | 0.245 ns | 0.229 ns |  1.00 |    0.00 | 0.0277 |      - |     232 B |        1.00 |
| DefaultConsoleOutputTemplate | 432.28 ns | 2.591 ns | 2.423 ns |  8.94 |    0.06 | 0.2074 | 0.0005 |    1736 B |        7.48 |
|              LiteralTextOnly |  59.97 ns | 0.099 ns | 0.088 ns |  1.24 |    0.01 | 0.0343 |      - |     288 B |        1.24 |
```

Results were similar on .NET 6 on ARM, though much slower across the board.

The additional eight-byte allocation in the baseline `EmptyTemplate` case is the change in `MessageTemplate` memory layout due to the additional `bool _acceptDottedPropertyNames` member.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -331,3 +331,5 @@ ASALocalRun/
 *.orig
 
 *.received.txt
+
+.DS_Store
--- a/src/Serilog/Parsing/MessageTemplateParser.cs
+++ b/src/Serilog/Parsing/MessageTemplateParser.cs
@@ -267,40 +267,85 @@ public class MessageTemplateParser : IMessageTemplateParser
 
     static TextToken ParseTextToken(int startAt, string messageTemplate, out int next)
     {
-        var accum = new StringBuilder();
-        do
+        // If we encounter escape sequences like {{ or }}, the result is not a strict substring of the
+        // template. But, this requires allocating a StringBuilder, so we try to parse as far as we can first, and
+        // only allocate the StringBuilder/fall through to the slow string-building path if we actually need to.
+        // Most of the time we won't hit escapes, so we can get away with just a single Substring() allocation at
+        // the end.
+
+        var i = messageTemplate.IndexOfAny(['{', '}'], startAt);
+        if (i == -1)
         {
-            var nc = messageTemplate[startAt];
-            if (nc == '{')
+            // No more interesting characters in the template, everything left is text.
+            next = messageTemplate.Length;
+            return new(messageTemplate[startAt..]);
+        }
+
+        StringBuilder accum;
+        var ch = messageTemplate[i];
+        ++i;
+
+        // The character must be either `{` or `}`, since we found its index.
+        if (ch == '{')
+        {
+            if (i < messageTemplate.Length && messageTemplate[i] == '{')
             {
-                if (startAt + 1 < messageTemplate.Length &&
-                    messageTemplate[startAt + 1] == '{')
+                // Hit an escape sequence; ignore the second (duplicate) `{`, and push the rest onto the
+                // accumulator to start the slow path.
+                accum = new(messageTemplate, startAt, i - startAt, messageTemplate.Length - startAt);
+                ++i;
+            }
+            else
+            {
+                // Hit the start of a property token. We're done, no StringBuilder was required.
+                next = i - 1;
+                return next == startAt ? EmptyTextToken : new(messageTemplate.Substring(startAt, i - 1 - startAt));
+            }
+        }
+        else // ch == '}'
+        {
+            accum = new(messageTemplate, startAt, i - startAt, messageTemplate.Length - startAt);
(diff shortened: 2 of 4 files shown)


nblumhardt:
(Surprising actually, that this late in the game a couple of simple `string.IndexOf()` calls can > 3x performance and < 1/3x allocations in a fairly mainline scenario 😅 .)

nblumhardt:
Heads-up also @Mpdreamz - I know that `Elastic.Serilog.Sinks` currently maps some `PascalCase` property names to dotted ECS equivalents. It'd be interesting to consider what a future iteration supporting dotted property names in message templates directly would mean over there.

nblumhardt:
Updated the PR to consider `.` only, and ignore `-` for now.

cchapman-kxs:
/lgtm

Thanks for taking the first step in enabling easier log schema adherence!

I'm hoping I can find a chance soon to test this but no promises.

Mpdreamz:
Love this! Our ECS field based setters can handle both the original dotted key names as well as PascalCased ones.

cchapman-kxs:
Also verified simple console and file logging works as expected.  Sending to DataDog is using the `DurableHttpUsingFileSizeRolledBuffers` from https://github.com/FantasticFiasco/serilog-sinks-http - very much not a trivial example and many places that could have interfered with the intended result.

nblumhardt:
Thinking about opening a PR to stabilize this. Any experiences to share, good or bad?

PR #2089, opened 2024-07-22
Allow non-anonymous-type destructuring when trimmed

nblumhardt:
An old refactor pulled out structure capturing logic into a method incorrectly named `TryConvertCompilerGeneratedType`; this in fact is the catch-all for non-scalar/dictionary/enumerable types without an explicit `IDestructuringPolicy` configured.

The misleading name led to incorrect logic being implemented for skipping capturing of compiler-generated type properties when not supported: in this situation all structure capturing will be skipped.

This PR fixes the method name - `TryConvertStructure` - and the aforementioned trimming-related logic.

### DIFF
--- a/src/Serilog/Capturing/GettablePropertyFinder.cs
+++ b/src/Serilog/Capturing/GettablePropertyFinder.cs
@@ -14,9 +14,9 @@
 
 namespace Serilog.Capturing;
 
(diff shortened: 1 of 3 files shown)


nblumhardt:
Pushing this through to keep #2090 fresh

PR #2093, opened 2024-07-25
Drop macOS CI target; too slow to run successfully

nblumhardt:
More often than not, it takes the macOS runner too long to schedule a new thread for our tests to succeed even with generous timeouts. Not getting much value from the target, and developing on macOS half the time anyway, so better just skipping this in CI.


Numpsy:
Is that 'h' file supposed to be there?

PR #2094, opened 2024-07-25
Removed boxing and string allocations in MessageTemplateTextFormatter

epeshk:
This PR removes allocations from legacy (but still not deprecated in Console/File sinks) `MessageTemplateTextFormatter`. This may still be useful, since the timestamp is serialized for each log event.

Allocations:
 - `DateTimeOffset` boxing, by adding `ScalarValue.Render<T>` overload with constraint for `IFormattable`
 - `DateTimeOffset.ToString()` allocation, by using `ISpanFormattable.TryFormat` API

### Before

| Method         | Job      | Runtime  | Mean     | Error   | StdDev  | Gen0   | Allocated |
|--------------- |--------- |--------- |---------:|--------:|--------:|-------:|----------:|
| FormatToOutput | .NET 8.0 | .NET 8.0 | 184.9 ns | 1.93 ns | 1.81 ns | 0.0143 |     120 B |
| FormatToOutput | .NET 9.0 | .NET 9.0 | 141.3 ns | 1.50 ns | 1.40 ns | 0.0143 |     120 B |

### After

| Method         | Job      | Runtime  | Mean     | Error   | StdDev  | Allocated |
|--------------- |--------- |--------- |---------:|--------:|--------:|----------:|
| FormatToOutput | .NET 8.0 | .NET 8.0 | 190.2 ns | 1.70 ns | 1.59 ns |         - |
| FormatToOutput | .NET 9.0 | .NET 9.0 | 135.6 ns | 1.02 ns | 0.95 ns |         - |
```

### DIFF
--- a/src/Serilog/Events/ScalarValue.cs
+++ b/src/Serilog/Events/ScalarValue.cs
@@ -95,6 +95,43 @@ public class ScalarValue : LogEventPropertyValue
         }
     }
 
+#if FEATURE_SPAN
+    internal static void Render<T>(T value, TextWriter output, string? format = null, IFormatProvider? formatProvider = null) where T : struct, IFormattable, ISpanFormattable
+    {
+        Guard.AgainstNull(output);
+
+        var custom = (ICustomFormatter?)formatProvider?.GetFormat(typeof(ICustomFormatter));
+        if (custom != null)
+        {
+            WriteWithCustomFormatter(value, output, format, formatProvider, custom);
+            return;
+        }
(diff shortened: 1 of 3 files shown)


nblumhardt:
Looks like a great change, left a comment on the other version of this PR which I think will also apply here: https://github.com/serilog/serilog-expressions/pull/121

Cheers!

nblumhardt:
Thanks Evgeny! 👍

Issue #1791, opened 2022-12-29
SelfLog with Log Level


raV720:
Support for Log Levels (at least Error and Information) in SelfLog.
It would allow to write and distinguish if logged information is an error or diagnostic log.

There are sinks (for example: https://github.com/serilog-mssql/serilog-sinks-mssqlserver, https://github.com/saleem-mirza/serilog-sinks-sqlite) which are using SelfLog not only to log errors but also to log diagnostic informations. It is hard to detect if logged information is caused by error or is a simple diagnostic log.

So SelfLog methods may look like this:
```
public static void Enable(Action<string, SelfLogLevel> output)
public static void WriteLine(string format, SelfLogLevel level, object? arg0 = null, object? arg1 = null, object? arg2 = null)
```

nblumhardt:
Thanks for the suggestion! The current design of `SelfLog` assumes that it's only turned on when trying to diagnose a problem; although a lot of people do use it for monitoring, there are some gaps to fill before this would be a good experience.

I don't think we're aiming to create a mini logging framework around `SelfLog`, so I'd personally be inclined to leave this alone. Open to hearing other perspectives though.

raV720:
I understand it would be a logger inside logger, but I think changes would be really minor and worth benefits.
Its just matter of adding log level enum and pass it around two methods.

sungam3r:
Logger inside logger, yes, but not so complicated logger. I would say very primitive logger. I agree that changes should be minimal but they are likely breaking. I would add new `SelfLog`-like property and some switch to use either old `SelfLog` or the new one (+ optionally deprecate old `SelfLog`).

raV720:
> I would add new SelfLog-like property and some switch to use either old SelfLog or the new one (+ optionally deprecate old SelfLog).
In my opinion it would be "more breaking" and complicating the solution.
Simply adding log level is also breaking, but fixing it is very simple and obvious - it is just matter of setting appropriate log level for all WriteLine methods.

nblumhardt:
Another quicker and more efficient option might be to just send PRs to the two projects listed in the original report for a conditional "verbose diagnostics" switch - worth serious consideration, since landing changes in two sinks will be much faster and more effective than trying to land changes in the hundreds of others already written.

raV720:
> landing changes in two sinks will be much faster and more effective than trying to land changes in the hundreds of others already written

[...]

PR #2108, opened 2024-08-30
Failure listeners and fallback sinks

nblumhardt:
Implements https://github.com/serilog/serilog/issues/1791.

The core addition is a new interface, `ILoggingFailureListener`, through which sinks can report temporary or permanent logging failures.

The most direct way to consume this is hooked up as `WriteTo.Fallible()`, which installs a provided failure listener for a given sink:

```csharp
    .WriteTo.Fallible(
        wt => wt.Email(...),
        new MyFailureListener())
```

Failure listeners get enough information about failures that monitoring/metrics use cases should be well handled.

The common batching infrastructure has been updated to support failure listeners, so existing sinks that either a) throw exceptions synchronously on failure, or b) use the common batching infrastructure, will automatically support failure listeners.

Sinks that don't fall into either category can implement `ISetLoggingFailureListener` and receive a failure listener at start-up.

The second, higher-level way to consume the new functionality is through fallback chains: the `WriteTo.FallbackChain()` method accepts an ordered sequence of sinks to try, and using failure listeners, will pass failed writes on to the next sink in the chain:

```csharp
    .WriteTo.FallbackChain(
        // First try writing to Seq
        wt => wt.Seq(...),
        // If that fails, try email
        wt => wt.Email(...),
        // Finally, write events to a file
        wt => wt.File(...))
```

A few APIs get added and so the feature needs some careful scrutiny. It's possible to construe this as bloat, but it's also proven very hard to implement anything similar in a general/cross-cutting way outside the core, so I'm leaning towards pushing ahead with this. Keen to streamline and refine it as much as we can.


nblumhardt:
Making a note that _Serilog.Sinks.Async_ will need an update to support this if/when merged.

alsi-lawr:
@nblumhardt thanks for the effort that's gone into this, it's greatly appreciated.

If I understand correctly, would there essentially be 2 _kinds_ of sinks that can be wrapped with a fallback chain:
1. Sinks that explicitly implement the `ISetFailureLoggingSink` interface get passed whatever failure listener configured during logger configuration, and then must explicitly call the `OnLoggingFailed` method based on the `LogFailureKind`
2. Sinks that don't implement the new `ISetFailureLoggingSink` interface get minimal support through exceptions only

Is there also support for state capture? Is that the goal of the `object sender` parameter of the `OnLoggingFailed` method?

src/Serilog/Core/ILoggingFailureListener.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <param name="kind">An indication of the failure kind.</param>
+    /// <param name="message">A description of the failure.</param>
+    /// <param name="events">Events associated with the failure.</param>
+    /// <param name="exception">The exception triggering the failure, if any.</param>
+    void OnLoggingFailed(
+        object sender,
+        LoggingFailureKind kind,
+        string message,
+        IReadOnlyCollection<LogEvent>? events,
+        Exception? exception);
```
alsi-lawr, line 36: Is there a reason that `Exception` is preferred over `ExceptionDispatchInfo` to preserve the stack trace?

nblumhardt: Thanks for taking a look!

If the failure is to be forwarded on to another Serilog sink, an `Exception` will be needed; `ExceptionDispatchInfo` is useful for re-throwing, but can't expose enough information to be useful for logging, unfortunately.

ArieGato:
Hi @nblumhardt,

Will it be possible to configure the Fallback sinks through configuration? Will it require a change in the Serilog.Settings.Configuration?

Cheers,
Arjan

nblumhardt:
Thanks for taking a look @ArieGato. The API used in the PR should work with the current feature set of _Serilog.Settings.Configuration_ 👍 

As an intermediate step, it should also be possible to configure fallbacks internally within your own sink configuration method, if you want to preserve configuration/API compatibility.

nblumhardt:
Planning to merge this after #2118 so we can try it out more broadly. It's a nice benefit of the switch to a built-in batching implementation that this should work with most of the existing ecosystem without any changes :-)

src/Serilog/Core/Sinks/Fallback/DelegatingLoggingFailureListener.cs
```diff
@@ -0,0 +1,29 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Serilog.Core.Sinks;
+
+sealed class DelegatingLoggingFailureListener(ILogEventSink sink) : ILoggingFailureListener
+{
+    public void OnLoggingFailed(object sender, LoggingFailureKind kind, string message, IReadOnlyCollection<LogEvent>? events, Exception? exception)
+    {
+        if (events is not null && kind is LoggingFailureKind.Final or LoggingFailureKind.Permanent)
```
nblumhardt, line 21: For batched sinks this will only occur after retries have been performed, causing a delay before the fallback sink will see the events.

By default I think this is okay, but we may wish to add a `RetryCount` or `DisableRetries` type of setting to `BatchingOptions` so that fallbacks are tried immediately when this behavior is required.

ArieGato:
> Thanks for taking a look @ArieGato. The API used in the PR should work with the current feature set of _Serilog.Settings.Configuration_ 👍
> 
> As an intermediate step, it should also be possible to configure fallbacks internally within your own sink configuration method, if you want to preserve configuration/API compatibility.

I'm trying to configure the fallback/fallback chain sinks through `appsettings.json`, but I can't get it to work. Can you give an example how I can achieve this?

nblumhardt:
Hi @ArieGato; I suspect I got this wrong, and the `params` usage in the API isn't yet supported by Serilog.Settings.Configuration. I'll open a ticket over there.

ArieGato:
@nblumhardt I did some investigating and I think the problem with configuring the FallackChain is partly due to the fact that the WriteTo.FallbackChain is not an extension method. This can be resolved in two ways:

1. Change the method to an extension method so that it is more inline with configuring Sinks.
2. Alter Serilog.Settings.Configuration to support normal extension methods.

What are your thoughts on this?

nblumhardt:
Thanks for taking a look @ArieGato. There's already a mechanism in Serilog.Settings.Configuration to deal with these:

https://github.com/serilog/serilog-settings-configuration/blob/dev/src/Serilog.Settings.Configuration/Settings/Configuration/SurrogateConfigurationMethods.cs

I'm a bit short on time for the next few weeks unfortunately, but it looks like it should be a fairly quick addition.

davidkarlsson:
What would adding support for this to `Serilog.Settings.Configuration` look like ideally? Wouldn't it be weird to try to add `FallbackChain` to `WriteTo` somehow since that key just takes an array of sinks? Should a separate "write to with fallbacks" key be added for this instead even though it then wouldn't match the Serilog api for it?

nblumhardt:
@davidkarlsson I think the current mapping of C# configuration onto JSON means that the `"FallbackChain"` would be under `"WriteTo"`, and then under that would be another level of `"WriteTo"`. It's a bit awkward, really needs some tinkering with :)

ArieGato:
> @nblumhardt I did some investigating and I think the problem with configuring the FallackChain is partly due to the fact that the WriteTo.FallbackChain is not an extension method. This can be resolved in two ways:
> 
> 1. Change the method to an extension method so that it is more inline with configuring Sinks.
> 2. Alter Serilog.Settings.Configuration to support normal non-extension methods.
> 
> What are your thoughts on this?

@nblumhardt  I finally got round adding support for FallbackChain and Fallable to Serilog.Settings.Configuration. 

https://github.com/serilog/serilog-settings-configuration/pull/474

Can you take a look at it?

PR #2127, opened 2024-10-18
An initial Actions-based CI build and publishing workflow

nblumhardt:
We'll soon need to make a range of releases in support of the .NET 9 wave (in particular, of _Serilog.AspNetCore_, _Serilog.Settings.Configuration_, and the _Serilog.Extensions.*_ packages).

This is a good time to continue our work on making all Serilog organization projects use consistent tooling.

This time around, I want to migrate CI to GitHub Actions, due to the easier management of permissions and faster build times.

Once this change is applied here, the `Directory.Build.props` and `ci.yml` files in this PR should be able to be copied verbatim to all other projects in the organization (with some initial removal of duplicates from CSPROJ files, and related clean-up).

The next step following this one will be to manage those files centrally, but in the meantime, just following the protocol of making changes here in `serilog/serilog` first, before taking them downstream to other Serilog projects, should be a decent improvement in our workflow.


nblumhardt:
Note that the automatic CI run on this PR is meaningless - needs to hit `dev` before Actions will be invoked.

PR #2141, opened 2024-11-27
Stabilize and fully-implement dotted property names in message templates

nblumhardt:
This PR continues on from #2063, which enables `{a.b}`-style placeholders in message templates. The major benefit of this is to support environments that have standardized on [OpenTelemetry](https://opentelemetry.io/docs/specs/semconv/), [DataDog](https://docs.datadoghq.com/standard-attributes/), or [ECS-style](https://www.elastic.co/guide/en/ecs/current/ecs-field-reference.html) naming schemes.

Dotted names, which are captured as _flat_ property names by Serilog, are compatible with a good slice of the Serilog ecosystem, but not all of it. Enough Serilog-based sinks, libraries and tooling will work with dotted names that these should be immediately usable by a majority of users. The process of finding and fixing problems when dots appear in property names may take some time, but can proceed incrementally with this enabling work done.

There's a small behavioral change here: in message templates, property names beginning with a number must now be completely numeric, so `{0_}` and `{0a}`, which were previously accepted, are now rejected. I'd expect the number of users that this might impact to be vanishingly small, but in the unlikely case someone is capturing data with these names, fallback capturing behavior will at still collect the events and properties.

### DIFF
--- a/appveyor-perftest.yml
+++ b/appveyor-perftest.yml
@@ -1,7 +1,0 @@
-version: '{build}'
-skip_tags: true
-image: Visual Studio 2019
-configuration: Release
-test: off
-build_script:
-- ps: ./RunPerfTests.ps1
--- a/src/Serilog/Parsing/MessageTemplateParser.cs
+++ b/src/Serilog/Parsing/MessageTemplateParser.cs
@@ -20,26 +20,13 @@ namespace Serilog.Parsing;
 /// </summary>
 public class MessageTemplateParser : IMessageTemplateParser
 {
-    /// <summary>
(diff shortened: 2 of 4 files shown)


cchapman-kxs:
Great to see this moving forward @nblumhardt!  Does this mean it is no longer behind a feature toggle and on by default?  Are there any functional differences beside the one you called about property names starting with a number?

nblumhardt:
Hi @cchapman-kxs! Yes, the featrue toggle is gone/now always "on". The tests here show all of the behavior (enabled and changed) that I'm aware of:

https://github.com/serilog/serilog/pull/2141/files

Issue #2011, opened 2024-02-08
Make `:lj` formatting the default/only option for message templates


nblumhardt:
Message templates processed directly via the `MessageTemplate` class, and therefore by default in various sinks, use quotes around embedded string values:

```csharp
Log.Information("Hello, {Name}!", "world");
// -> Hello, "world"!
```

This is surprising and usually not desired. Modern formatting options such as _Serilog.Sinks.Console_'s custom output formatter, and `ExpressionTemplate`, do away with this:

```csharp
Log.Information("Hello, {Name}!", "world");
// -> Hello, world!
```

Similarly, older formatters use a C#-like syntax for rendering objects - `{Name = "world"}`, while newer ones just produce JSON - `{"Name": "world"}`.

It's time to do away with the old convention and make the codebase behave consistently throughout, using the new formatting style.

hayer:
This would be very useful for me, specially used together with the `RenderedCompactJsonFormatter`.

nblumhardt:
I think the trick here will be to introduce Serilog.Expressions-based formatting APIs, and migrate away from the original.

PR #2143, opened 2024-12-02
Make :lj formatting the default/only option for message templates

hayer:
#2011


nblumhardt:
Thanks! I had a shot earlier and shied away from this because of potential breakage, but I can see you've thought about that too. I'll take another look.

hayer:
@nblumhardt, any input would be appreciated. 

I can't see any way to accomplish this using the `RenderedCompactJsonFormatter` sink.

nblumhardt:
Sorry about the slow turnaround, aiming to get to this ASAP.

You should be able to achieve an equivalent format to `RenderedCompactJsonFormatter`, with `:lj` formatting by default, using Serilog.Expressions:

https://nblumhardt.com/2021/06/customize-serilog-json-output/

Issue #739, opened 2016-05-13
Best Performance Practices for enriching messages with properties


corneliutusnea:
Hi guys,

What are the recommended practices for enriching messages with properties from a performance perspective? As far as I could find there are several ways of enriching message:

1: Using the local context
```
var disposable = new DisposableCollection();
disposable.Add(LogContext.PushProperty("property1", "value1"));
disposable.Add(LogContext.PushProperty("property2", "value2"));
using(disposable){
   _logger.Info("template {value}", someValue);
}
```
2: Using ForContext
```
_logger.ForContext("property1", "value1")
   .ForContext("property2","value2")
   .Info("template {value}", someValue);
```
3: Using PropertyEnrichers
```
_logger.ForContext(new PropertyEnricher[] {
   new PropertyEnricher("property1", "value1", true)
   , new PropertyEnricher("property2", "value2", true)
}).Info("template {value}", someValue);
```

Is there a recommended way in order to achieve maximum performance?
No 1 to me feels like the least performance because you have to hit the ThreadLocalStorage both writing those values in the context and reading them.

No 2 seems to be creating a new `Logger` internally for each call into the `.ForContext`. If I have 6-10 such properties instead of 2 that feels like a serious performance hit on the GC to create so many extra objects just to log a string.

No 3 seems to be the best as I could pre-create those `PropertyEnricher`s and keep them on my context higher up the stack and re-feed them to all the logging I have to do inside a context.
However No 3 in the implementation seems to be parsing/deconstructing the value every time it has to enrich a message so reusing it feels like it would not achieve anything really.
Maybe a combination of the `PropertyEnricher` and the `FixedPropertyEnricher` that seems to e caching the `LogEventProperty` so I can truly reuse that parsed/deconstructed property.

Thoughts?

nblumhardt:
Hey Corneliu!

API number three is the one to use; just create your own enricher optimised for your scenario:

```csharp
class RocketParametersEnricher : ILogEventEnricher
{
    readonly LogEventProperty _designation, _capacity;

    public RocketParametersEnricher(string designation, int capacity)
    {
        _designation = new LogEventProperty("Designation", new ScalarValue(designation));
        _capacity = new LogEventProperty("Capacity", new ScalarValue(capacity));
    }

    public void Enrich(LogEvent logEvent, ILogEventPropertyValueFactory unused)
    {
        logEvent.AddPropertyIfAbsent(_designation);
        logEvent.AddPropertyIfAbsent(_capacity);
[...]

PR #2149, opened 2024-12-19
Property factory as AddPropertyIfAbsent parameter

vanni-giachin:
In this PR I propose to set public this internal method overload:
```
LogEvent.AddPropertyIfAbsent(ILogEventPropertyFactory factory, string name, object? value, bool destructureObjects = false)
```

The goal is allow to write optimised `ILogEventEnricher`classes and also to be able to use `ILogEventPropertyFactory` to create properties.

In therm of performance this solution would be almost the same as the code suggested in https://github.com/serilog/serilog/issues/739 where properties are created only once, but with the benefit of using the property factory as intended by the `ILogEventEnricher` interface. 

Example:
```
class SampleEnricher(Foo foo) : ILogEventEnricher
{
    public void Enrich(LogEvent logEvent, ILogEventPropertyFactory propertyFactory)
    {
        // a LogEventProperty object is created by the factory only if missing
        logEvent.AddPropertyIfAbsent(propertyFactory, nameof(foo.Bar), foo.Bar);
    }
}
```

### DIFF
--- a/src/Serilog/Events/LogEvent.cs
+++ b/src/Serilog/Events/LogEvent.cs
@@ -204,8 +204,20 @@ public class LogEvent
 #endif
     }
 
-    internal void AddPropertyIfAbsent(ILogEventPropertyFactory factory, string name, object? value, bool destructureObjects = false)
+    /// <summary>
+    /// Add a property to the event if not already present.
+    /// </summary>
+    /// <param name="factory">Factory for creating the property to add to the event.</param>
+    /// <param name="name">The name of the property.</param>
+    /// <param name="value">The value of the property.</param>
+    /// <param name="destructureObjects">If <see langword="true"/>, and the value is a non-primitive, non-array type,
+    /// then the value will be converted to a structure; otherwise, unknown types will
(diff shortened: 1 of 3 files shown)


sungam3r:
As I can see from codebase `PropertyEnricher` allows you to call `AddPropertyIfAbsent` without creating `LogEventProperty`.

vanni-giachin:
> As I can see from codebase `PropertyEnricher` allows you to call `AddPropertyIfAbsent` without creating `LogEventProperty`.

@sungam3r yes, correct, `PropertyEnricher` calls the method I propose to turn public. In some scenarios I prefer to use a custom `ILogEventEnricher` since I have to enrich the log events with multiple correlated properties, like multiple fields of an object passed to the enricher constructor, that don't need to be recalculated every time a log is emitted.

sungam3r:
When workflow run you'll see one failed test. Run api approval test locally.

vanni-giachin:
> When workflow run you'll see one failed test. Run api approval test locally.

Ok. Fixed, thanks!

nblumhardt (approved):
LGTM, thanks 👍

PR #2157, opened 2025-02-10
Improve Serilog logo

kalfilla:


SimonCropp:
@nblumhardt lgtm

nblumhardt:
At least in Firefox where I'm viewing it, the PR'd version doesn't match the original.

SimonCropp:
i think the original was incorrectly resized. when while it might render ok, it can be opimzied. here is is in chrome

image

nblumhardt:
Yep, the new version's white outline is wrong.

nblumhardt:
Thanks for sending this @kalfilla. Unfortunately the cleaned-up version doesn't match the original, so I'll close this for now, but if you're interested in taking a shot at getting these aligned please feel free to jump back in. Cheers!

PR #2158, opened 2025-02-10
Use HTTPS in README.md images and links

TimHess:
Avoid mixed content issues and unnecessary redirects with https links and image tags


nblumhardt:
Good spotting - thanks! :-)

PR #2163, opened 2025-03-11
allow pushing a IEnumerable of ILogEventEnricher

SimonCropp:
given in some contexts i build up a list of ILogEventEnricher of a (runtime) variable size. So the array nature of push forces a redundant array alloc.

this PR adds a `params IEnumerable` overload.

I am not sure on having the Obsolete? the OverloadResolutionPriority will point the compiler to IEnumerable overload. so it is not a compile time breaking change. and it is not a runtime breaking change. thoughts?

### DIFF
--- a/src/Serilog/Context/LogContext.cs
+++ b/src/Serilog/Context/LogContext.cs
@@ -129,7 +129,7 @@ public static class LogContext
     /// <param name="enrichers">Enrichers to push onto the log context</param>
     /// <returns>A token that must be disposed, in order, to pop properties back off the stack.</returns>
     /// <exception cref="ArgumentNullException">When <paramref name="enrichers"/> is <code>null</code></exception>
-    [OverloadResolutionPriority(1) ]
+    [OverloadResolutionPriority(1)]
     public static IDisposable Push(params IEnumerable<ILogEventEnricher> enrichers)
     {
         Guard.AgainstNull(enrichers);
--- a/src/Serilog/Context/LogContext.cs
+++ b/src/Serilog/Context/LogContext.cs
@@ -66,6 +66,7 @@ public static class LogContext
     /// <param name="enricher">An enricher to push onto the log context</param>
     /// <returns>A token that must be disposed, in order, to pop properties back off the stack.</returns>
     /// <exception cref="ArgumentNullException">When <paramref name="enricher"/> is <code>null</code></exception>
+    [OverloadResolutionPriority(3)]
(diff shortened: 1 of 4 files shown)


nblumhardt:
Makes sense 👍 

I think since there's little chance we'd actually remove the original array version in the foreseeable future, and considering there might be scenarios in which it generates a warning (e.g. old compiler versions?), perhaps we drop the obsolete attribute?

Gave a bit of thought to whether we'd also want to provide a `params ReadOnlySpan` version, too; seems like it'd cover a slightly different set of cases and still be an option in a later version if we decided to include it then, so probably not, but just mentioning it in case you spot any conflicts/synergies with that hypothetical API.

SimonCropp:
> perhaps we drop the obsolete attribute

done

> Gave a bit of thought to whether we'd also want to provide a params ReadOnlySpan version

added

i also added some tests

SimonCropp:
i also removed the OverloadResolutionPriority. since if we want to keep the array varient around, there is no need to redirect

SimonCropp:
nudge

nblumhardt (commented):
Thanks! :+1: 

Looks like CI is failing - stack overflow?

src/Serilog/Context/LogContext.cs
```diff
@@ -88,15 +89,59 @@ public static IDisposable Push(ILogEventEnricher enricher)
     /// <param name="enrichers">Enrichers to push onto the log context</param>
     /// <returns>A token that must be disposed, in order, to pop properties back off the stack.</returns>
     /// <exception cref="ArgumentNullException">When <paramref name="enrichers"/> is <code>null</code></exception>
+    [OverloadResolutionPriority(0)]
     public static IDisposable Push(params ILogEventEnricher[] enrichers)
+    {
+        return Push(enrichers);
```
nblumhardt, line 95: If this ends up resolving as `IEnumerable` there will be a perf regression; the earlier version used numeric indexing because there's no struct enumerator for `IEnumerable`. Possible to keep `AsSpan` here to sort this out, plus make the target method easier to identify?

nblumhardt: If `AsSpan` is used, a guard against null arg will be needed.

nblumhardt: > If AsSpan is used, a guard against null arg will be needed.

Added this, LGTM, will merge if my changes come through green :-)

Issue #2140, opened 2024-11-27
ILC : Trim analysis error IL2072


guilhermelinosp:
Serilog.Capturing.PropertyValueConverter.CreatePropertyValue(Object,Destructuring,Int32):
'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'Serilog.Capturing.PropertyValueConverter.TryConvertCompilerGeneratedType(Object,Type,Destructuring,LogEventPropertyValue&)'. 

The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

JR-Morgan:
Are there any updates on this @agocke @nblumhardt ?

This warning is preventing me from trimming my apps that depend on Serilog. Is there an older version of Serilog that is correctly trimmable in .NET8?

agocke:
Lemme take a look. I think the code changed slightly and invalidated an assumption. Hopefully an easy fix.

JR-Morgan:
Ah nice, that was speedy ty,
After testing a little with [4.3.0-dev-02364](https://www.nuget.org/packages/Serilog/4.3.0-dev-02364), no more trimming warnings 🙌

---

Not too certain if this is still required/the best advice, but the Microsoft docs suggest setting up a trimming test project that would consume and attempt to trim Serilog, docs suggest it may catch more trimming warnings that `EnableTrimAnalyzer` may not...
https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming

JR-Morgan:
AH, I see this is already being discussed https://github.com/serilog/serilog/pull/2175

PR #2175, opened 2025-05-10
Fix AOT compatibility

agocke:
It looks like the code has drifted a bit and disrupted the feature switch that provided AOT compatibility. I've fixed the feature switch, added the AotTestApp to the solution to ensure that it builds when the solution does, configured the AotTestApp to publish on build so that it runs the IL analysis engine, and added a Serilog.props file to the NuGet package to enable the feature switch automatically whenever trimming is enabled.

This should be much more likely to catch any regressions.

Fixes https://github.com/serilog/serilog/issues/2140

### DIFF
--- a/Serilog.sln
+++ b/Serilog.sln
@@ -1,3 +1,4 @@
+﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio Version 17
 VisualStudioVersion = 17.5.33424.131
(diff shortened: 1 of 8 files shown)


nblumhardt:
Thanks @agocke! The new naming makes the purpose of the switch much clearer, looks good 👍

agocke:
Thanks! Any idea when this would appear in a nuget package?

nblumhardt:
`4.3.0-dev-02358` on NuGet includes this; I've also queued up #2177 to get an RTM out shortly. Cheers!

agocke:
Thanks for the fast response!

nblumhardt:
@agocke I just pulled the `-dev` version into our codebase; I haven't tweaked anything (and don't have any code to set the switch), but when running unit tests against .NET 9:

```csharp
  System.AppContext.TryGetSwitch("Serilog.Capturing.IsStructureValueSupported", out var isEnabled)
  // -> true

  isEnabled
  // -> false
```

This switches off structure value capturing. There's no AOT in our toolchain that I'm aware of.

I'm guessing that it's:

```xml
<Project>
    <ItemGroup>
        <RuntimeHostConfigurationOption
        Include="Serilog.Capturing.IsStructureValueSupported"
        Value="false" Trim="true" />
    </ItemGroup>
</Project>
```

It seems to me that `Value` should possibly be `true` by default, here?

agocke:
Ah, I think I actually made a mistake by setting that RuntimeHostConfigurationOption unconditionally. I think it has the right value, but it should be guarded by `Condition="'$(PublishTrimmed)' == 'true'`. It also probably needs to be set in the targets file, not the props file. Lemme try to work this up locally.

PR #2177, opened 2025-05-10
4.3.0 Release

nblumhardt:
* #2149 - `LogEvent.AddPropertyIfAbsent(ILogEventPropertyFactory, ...)` overload that helps avoid allocations (@vanni-giachin)
 * #2158 - use HTTPS in all README.md images and links (@TimHess)
 * #2163 - `LogContext.Push()` overloads accepting `IEnumerable<ILogEventEnricher>` and `ReadOnlySpan<ILogEventEnricher>` (@SimonCropp)
 * #2175, #2178, #2179 - fix AOT compatibility (@agocke)

PR #2178, opened 2025-05-14
Move trim option to targets file

agocke:
I believe this fixes the problem observed in https://github.com/serilog/serilog/pull/2175.

I tried to validate this locally, but I'm not sure how thorough I am. I'd appreciate instructions on how to run the test suite.

### DIFF
--- a/src/Serilog/Serilog.csproj
+++ b/src/Serilog/Serilog.csproj
@@ -48,7 +48,7 @@
   <ItemGroup>
     <None Include="../../assets/icon.png" Pack="true" Visible="false" PackagePath="/" />
     <None Include="../../README.md" Pack="true" Visible="false" PackagePath="/" />
-    <None Include="Serilog.props" Pack="true" Visible="false" PackagePath="/build" />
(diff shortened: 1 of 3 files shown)


nblumhardt:
Thanks! I'm not sure either; I'd have expected the unit tests to fail in CI, but it seems to have passed. I'll merge and try it again here.

nblumhardt:
Looks good in the untrimmed case, now (Seq's integration tests are all green).

Is your testing sufficient to suggest we've still got the AOT case covered? If so, I'll push the button on #2177 in a day or two :-)

I think this came out nicely; I wish I had a bit more time to dig into AOT and trimming, have to put it back on my list. Thanks for all your help!

agocke:
> Is your testing sufficient to suggest we've still got the AOT case covered? If so, I'll push the button on https://github.com/serilog/serilog/pull/2177 in a day or two :-)

Yeah, I think the AotTestApp should cover that. I just realized, I think I can improve the test coverage slightly.

PR #2179, opened 2025-05-14
Reference targets file from tests

agocke:
Add the targets file import to the other test projects. This should increase the test coverage to make it look closer to what a NuGet package reference would experience.


agocke:
Follow up to https://github.com/serilog/serilog/pull/2178

nblumhardt:
Looks good, thanks 👍

PR #2180, opened 2025-05-15
Add TypeDestructuringPolicy

SimonCropp:
Adds a type specific variant of ProjectedDestructuringPolicy for the known type scenation

Avoids a GetType, a delegate creation, and a delegate execution


nblumhardt:
This looks good, but its behavior will vary from the reflection version by matching subtypes, while the reflection one is exact. I think it's only a superficial problem - the two functions have the same name but different behavior - but it'd be good to think a bit more about how to fit this into the API.

`ByTransformingAssignableTo<T>()` is a mouthful, but we could take that path and potentially provide statically-typed and reflection-based versions of both?

A `ByTransforming<T>()` for exact types would still be pretty cheap since in `value is T && value.GetType() == typeof(T)` the first clause would exclude non-matches without `GetType()`.

SimonCropp:
@nblumhardt yep that sounds good. PR adjusted

nblumhardt (approved):
Will hold onto this for a couple of days in case a patch to 4.3.0 needs to go out from `main`; otherwise I'll merge this and bump the minor version. Cheers!

SimonCropp:
@nblumhardt thanks. and no rush

PR #2181, opened 2025-05-15
add some notnull to transform APIS

SimonCropp:


nblumhardt (approved):
Might have be a 5.0 change; need to figure out whether adding `notnull` is binary-breaking - first-glance intuition is that it's probably nonbreaking, but I'm not sure what the implications are for `Nullable<T>`/whether there are special cases that may break.

PR #2203, opened 2025-09-24
Remove backpressure from XMLDoc

timothycoleman:
Because backpressure is not applied, instead events are discarded when the queue is full.

The xmldoc for `BatchingOptions.QueueLimit` says that backpressure is applied when the limit is exceeded.

However, in `BatchingSink.cs` we see that no backpressure is applied

```
_queue.Writer.TryWrite(logEvent);
```

Instead the event is discarded (which is good).

To me it looks like a mistake in the XMLDoc, and this PR adjusts it.


nblumhardt:
Nice spotting - thanks! (Some leftovers from an earlier prototype, IIRC :-))

PR #2204, opened 2025-10-03
Add selective destructuring policy for logging specific properties

### LINKED ISSUE #2192: Add rendering options for destructuring complicated objects
## User Story

Hey, I suppose this is a bit of necromancy here of issue #908, but I was wondering if anyone else found a good way to expose certain parts of a property to be rendered while destructuring the whole?

We have a similar issue as @harvinders did within the original thread, but I figured just in case there were some differences I would create a fresh topic for discussion.

The context is logging complicated objects, without rendering them in full detail within the rendered message:

```json5
MyComplicatedObject: {
  Name: "SomeName"
  SimpleValue: "42",
  ImportantData: "Lots of Data..."
}
```

Ideally we want 

```C#
logger.Information("My object is {@MyComplicatedObject}")
```

To render as 

```text
"My object is { "Name": SomeName, "SimpleValue": 42 }"
```

But instead, it renders as 
```text
"My object is { 'Name': SomeName, 'SimpleValue': 42, 'ImportantData': 'Lots of Data...' }"
```

As expected, that `ImportantData` is important, so we can't afford to drop it. It's just not helpful to the 'at a glance' ideal use case for rendered messages.

Of course, you can use a `ForContext` or some other variation of enricher (Let's just pretend we have `.ForMyObject()`, but then you run into this issue:

```C#
logger.ForMyObject(myComplicatedObject).Information("My object is {MyObjectName");
```

Which means your JSON packet looks something along the lines of

```json5
MyComplicatedObject: {
  Name: "SomeName"
  SimpleValue: "42",
  ImportantData: "Lots of Data..."
}
MyObjectName: "SomeName"
```

I know this isn't a concern for some folks, and makes a lot of sense in many use cases, but to me, it seems preferable to find a way to not duplicate this data. It's nice to keep our packets smaller for high-intensity applications.

I want to display _parts_ of the structured object, but only parts.

## Suggestion

The suggestion is to implement some kind of partial reference, or destructured.

### Reference

```C#
logger.ForMyObject(myComplicatedObject).Information("My object is {%Name}");
```

```json5
MyComplicatedObject: {
  Name: "SomeName"
  SimpleValue: "42",
  ImportantData: "Lots of Data..."
}
```

```text
"My object is SomeName"
```

The `%` operator would search current context when used within the renderer and present it. If no context can be found, it could return "" or `null` depending on Serilog internal needs.

### Partial Destructure

In the destructuring policy, we could instead have some kind of overload when returning a object for destructuring.

Perhaps, some API like this:

```C#
public class MyComplicatedObjectDestructuringPolicy : IDestructuringPolicy
{
    /// <inheritdoc/>
    public bool TryDestructure(
        object property,
        ILogEventPropertyValueFactory propertyValueFactory,
        out LogEventPropertyValue? result
    )
    {
        if (property is not MyComplicatedObject myObject)
        {
            result = null;
            return false;
        }

        var destructuredObject = new
        {
            myObject.Name,
            myObject.SimpleValue,
        };
        
        var destructuredObjectAdditional = new
        {
            myObject.ImportantValue,
        };

        result = propertyValueFactory.CreatePropertyValue(destructuredObject, destructuredObjectAdditional, true);
        return true;
    }
}
```

## Final Thoughts

If there are better ways to achieve this, I'm fully for it and open to suggestions or partial implementations, but everything I've played around with so far hasn't been able to fit the bill in a performance friendly way.

We considered several solutions, but we're running out of steam.

For now, our two viable options seem:

- Have verbose rendered messages and let the readability suffer
- Have duplicated messages and let the packets suffer

Anybody have any ideas or suggestions for this feature request?

Nonanti:
## Summary

This PR implements issue #2192 by adding a new `SelectiveDestructuringPolicy<T>` class that allows users to specify which properties should be included when destructuring objects in Serilog logs.

## Problem

When logging complex objects, Serilog currently destructures the entire object, which can:
- Make log messages verbose
- Include unnecessary/sensitive details  
- Impact performance with large objects

## Solution

Introduced a new destructuring policy that allows selective property inclusion:

```csharp
    .Destructure.BySelecting<MyComplicatedObject>("Name", "Id", "Status")
    .WriteTo.Console()
    .CreateLogger();
```

## Implementation Details

- **`SelectiveDestructuringPolicy<T>`**: A generic destructuring policy that accepts property names to include
- **`BySelecting<T>` extension method**: Added to `LoggerDestructuringConfiguration` for easy configuration
- **Trimming/AOT support**: Includes proper `DynamicallyAccessedMembers` annotations for .NET 8+ trimming compatibility
- **Comprehensive tests**: Added unit tests covering various scenarios

## Test Coverage

The implementation includes tests for:
- Type-specific destructuring (only applies to specified type)
- Selective property inclusion
- Null property value handling
- Non-existent property handling
- Integration with LoggerConfiguration
- Edge cases (empty property list)

## Breaking Changes

None - this is purely additive functionality.

## Example Usage

```csharp
public class User
{
    public string Name { get; set; }
    public int Id { get; set; }
    public string Email { get; set; }
    public string PasswordHash { get; set; }
}

// Only log Name and Id, excluding sensitive data
logger
    .Destructure.BySelecting<User>("Name", "Id")
    .Information("User logged in {@User}", user);
// Output: User logged in {Name: "John", Id: 123}
```

Fixes #2192

### DIFF
--- a/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerDestructuringConfiguration.cs
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Diagnostics.CodeAnalysis;
+
 namespace Serilog.Configuration;
 
 /// <summary>
@@ -202,5 +204,21 @@ public class LoggerDestructuringConfiguration
 
         _setMaximumCollectionCount(maximumCollectionCount);
         return _loggerConfiguration;
+    }
+
+    /// <summary>
+    /// When destructuring objects of the specified type, only include the selected properties
+    /// in the log output. This is useful for logging only relevant parts of complex objects.
+    /// </summary>
+    /// <typeparam name="T">Type of objects to apply selective destructuring to.</typeparam>
+    /// <param name="propertyNames">Names of the properties to include when destructuring.</param>
+    /// <returns>Configuration object allowing method chaining.</returns>
+    /// <exception cref="ArgumentNullException">When <paramref name="propertyNames"/> is <code>null</code></exception>
+    public LoggerConfiguration BySelecting<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] T>(params string[] propertyNames)
+    {
+        Guard.AgainstNull(propertyNames);
+
+        var policy = new SelectiveDestructuringPolicy<T>(propertyNames);
+        return With(policy);
     }
 }
--- a/src/Serilog/Policies/SelectiveDestructuringPolicy.cs
+++ b/src/Serilog/Policies/SelectiveDestructuringPolicy.cs
@@ -1,0 +1,81 @@
+// Copyright 2013-2015 Serilog Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+
+namespace Serilog.Policies;
+
+/// <summary>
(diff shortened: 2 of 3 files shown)


nblumhardt:
Hi! Thanks for the PR.

The linked issue is actually for a slightly different requirement - selective rendering, rather than selective capturing.

The functionality in this PR is already covered by:

```csharp
    .Destructure.ByTransforming<MyComplicatedObject>(c => new {c.Name, c.Id, c.Status})
    // ...
```

There's definitely room to improve the current API, e.g. accepting an `Expression<T>` and pulling out the targeted property names, instead of constructing-then-reflecting, would be interesting to explore), but it's not certain that there's enough benefit there to motivate carrying around the additional code.

Cheers!

Issue #2194, opened 2025-08-08
IL2067 trim analysis on Serilog.Capturing.PropertyValueConverter.CreateStructureValue


Mpdreamz:
When trying to publish an AOT console app using Serilog `4.3.0` and `.NET 9`. 

I am seeing the following IL2067 trim analysis on `Serilog.Capturing.PropertyValueConverter.CreateStructureValue`

> 
> ecs-aot-smoketest failed with 2 error(s) (58.2s) → examples/ecs-aot-smoketest/bin/release/net9.0/osx-arm64/ecs-aot-smoketest.dll
>     ILC : Trim analysis error IL2067: Serilog.Capturing.PropertyValueConverter.TryConvertStructure(Object,Type,Destructuring,StructureValue&): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicProperties' in call to 'Serilog.Capturing.PropertyValueConverter.CreateStructureValue(Object,Type,Boolean)'. The parameter 'type' of method 'Serilog.Capturing.PropertyValueConverter.TryConvertStructure(Object,Type,Destructuring,StructureValue&)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
>     /Users/mpdreamz/.nuget/packages/microsoft.dotnet.ilcompiler/9.0.7/build/Microsoft.NETCore.Native.targets(317,5): error MSB3073: The command ""/Users/mpdreamz/.nuget/packages/runtime.osx-arm64.microsoft.dotnet.ilcompiler/9.0.7/tools/ilc" @"obj/release/net9.0/osx-arm64/native/ecs-aot-smoketest.ilc.rsp"" exited with code -1.

Link to Serilog.Sinks.Elasticsearch PR: https://github.com/elastic/ecs-dotnet/pull/529

nblumhardt:
Thanks for the report, @Mpdreamz 👍

agocke:
OK, I think I managed to recreate a situation similar to this. The problem is the transitive, rather than direct, reference to Serilog.

So the fix is to add an explicit reference to Serilog with version 4.3.0. The problem is that NuGet doesn't add the builds/targets to transitive references.

nblumhardt:
Thanks for digging in, @agocke! Guess we need a README section for this 👍

Numpsy:
> The problem is that NuGet doesn't add the builds/targets to transitive references.

[...]

PR #2214, opened 2026-01-20
Package Serilog.targets as 'buildTransitive' as well as 'build'

Numpsy:
refs https://github.com/serilog/serilog/issues/2194

This should result in the targets file being imported when Serilog is either a direct or transitive dependency of the consuming project, rather than only when it's a direct dependency


Numpsy:
This looks right when I build it locally:

image

PR #2215, opened 2026-01-26
Inline TraceId and SpanId JSON string formatting

SimonCropp:
Replaces usage of JsonValueFormatter.WriteQuotedJsonString for TraceId and SpanId with direct writing of quoted string values.

activitytraceid is only either numbers or lower-case hexadecimal https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activitytraceid.createfromstring?view=net-8.0#exceptions

same for activityspanid
https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.activityspanid.createfromstring?view=net-8.0#exceptions

### DIFF
--- a/src/Serilog/Formatting/Json/JsonFormatter.cs
+++ b/src/Serilog/Formatting/Json/JsonFormatter.cs
@@ -74,13 +74,17 @@ public sealed class JsonFormatter : ITextFormatter
         if (logEvent.TraceId != null)
         {
             output.Write(",\"TraceId\":");
-            JsonValueFormatter.WriteQuotedJsonString(logEvent.TraceId.ToString()!, output);
+            output.Write('\"');
+            output.Write(logEvent.TraceId.ToString()!);
+            output.Write('\"');
         }
 
         if (logEvent.SpanId != null)
         {
             output.Write(",\"SpanId\":");
-            JsonValueFormatter.WriteQuotedJsonString(logEvent.SpanId.ToString()!, output);
+            output.Write('\"');
+            output.Write(logEvent.SpanId.ToString()!);
+            output.Write('\"');
         }
 
         if (logEvent.Exception != null)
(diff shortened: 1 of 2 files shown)


ite-jankl:
Why was this change made? The previous behavior is not wrong, and always results in the same result, right? The previous code looks more descriptive/definitely sanitizing. Is this purely for better performance?

SimonCropp:
@ite-jankl 
WriteQuotedJsonString is relatively expensive. checking char by char to possibly escape and quote.

when we know something doesnt not contain anything that needs to be escaped, we should avoid it
```
   public static void WriteQuotedJsonString(string str, TextWriter output)
    {
        output.Write('\"');

        var cleanSegmentStart = 0;
        var anyEscaped = false;

        for (var i = 0; i < str.Length; ++i)
        {
            var c = str[i];
            if (c is < (char)32 or '\\' or '"')
            {
                anyEscaped = true;

#if FEATURE_SPAN
                output.Write(str.AsSpan().Slice(cleanSegmentStart, i - cleanSegmentStart));
#else
                output.Write(str.Substring(cleanSegmentStart, i - cleanSegmentStart));
#endif
                cleanSegmentStart = i + 1;

                switch (c)
                {
                    case '"':
                        output.Write("\\\"");
                        break;
                    case '\\':
                        output.Write("\\\\");
                        break;
                    case '\n':
                        output.Write("\\n");
                        break;
                    case '\r':
                        output.Write("\\r");
                        break;
                    case '\f':
                        output.Write("\\f");
                        break;
                    case '\t':
                        output.Write("\\t");
                        break;
                    default:
                        output.Write("\\u");
                        output.Write(((int)c).ToString("X4"));
                        break;
                }
            }
        }

        if (anyEscaped)
        {
            if (cleanSegmentStart != str.Length)
#if FEATURE_SPAN
                output.Write(str.AsSpan().Slice(cleanSegmentStart));
#else
                output.Write(str.Substring(cleanSegmentStart));
#endif
        }
        else
        {
            output.Write(str);
        }

        output.Write('\"');
    }
```

ite-jankl:
I see, thank you for the explanation and details

Issue #2211, opened 2025-12-18
Minor optimization in MessageTemplate positional token validation


its-kaktos:
After reviewing MessageTemplate.cs, I noticed a section that checks whether all propertyTokens are positional. It seems possible to break out of the loop early once a non-positional property is found, which might simplify the logic and avoid unnecessary iterations.
Conceptually, it could look something like this:
```cs
 var allPositional = true;
  var anyPositional = false;
  foreach (var propertyToken in propertyTokens)
  {
      if (propertyToken.IsPositional)
      {
           allPositional = false;
           break;
       }
  }
```

The section that Im talking about is:
https://github.com/serilog/serilog/blob/7f3a5bfec73dba1c3b29780b85a53b0fe2f7c52b/src/Serilog/Events/MessageTemplate.cs#L56:L64

If this approach is valid and acceptable, I would be happy to submit a pull request.

HarryYardley:
Opened a fresh PR for this on current dev, based on the earlier #2218/#2219 approach, with behavior preserved and tests passing.

nblumhardt:
Mixed ordinal/positional templates are very rare, there probably isn't much real-world gain here. Changing this behavior requires introducing new branches to a hot loop, so while we might gain something in rare cases, it's more likely we'll lose perf in the general case.

Closing as not planned, but thanks for raising this.

PR #2219, opened 2026-02-03
Optimize positional token validation with early exit

dipeshchaudhari3255:
Related to #2211

## What
Optimizes positional token validation in `MessageTemplate` by short-circuiting
the loop when a non-positional token is encountered.

## Why
The previous logic iterated through all tokens even when the final result
was already known. This improves readability and avoids unnecessary iterations
without changing behavior.

## Notes
- No behavioral change
- No allocations added
- All existing tests pass


dipeshchaudhari3255:
@serilog/maintainers PTAL when you have a moment. Thanks!

PR #2222, opened 2026-02-14
Emit SelfLog warning when extra arguments are provided

### LINKED ISSUE #1600: No SelfLog event when there are extra arguments
**Description**
I assumed there would be a SelfLog event when a logging statement provides more data than the given message template expects.  However, this does not seem to be the case.

**Reproduction**
Consider the following code.

```csharp
Serilog.Debugging.SelfLog.Enable(s => Console.WriteLine(s));
var logger = new LoggerConfiguration().WriteTo.Console().CreateLogger();
logger.Information("Argument missing {MissingArgument}");
logger.Information("Argument extra", 42);
```

**Expected behavior**
I expected it to have output something like the following.
```
2021-06-29T20:07:47.9719969Z Required properties not provided for: Argument missing {MissingArgument}
[15:07:47 INF] Argument missing {MissingArgument}
2021-06-29T20:07:47.9719969Z Extra properties provided
[15:07:48 INF] Argument extra
```

**Actual behavior**
This is the actual output.

```
2021-06-29T20:07:47.9719969Z Required properties not provided for: Argument missing {MissingArgument}
[15:07:47 INF] Argument missing {MissingArgument}
[15:07:48 INF] Argument extra
```

- Serilog 2.10.0
- Serilog.Sinks.Console 3.1.1

matantsach:
## Summary

Fixes #1600

When a message template has no properties but arguments are supplied (e.g. `logger.Information("Argument extra", 42)`), no `SelfLog` diagnostic was emitted, making it difficult to detect mismatched log calls at runtime.

## Changes

- **`PropertyBinder.ConstructNamedProperties`**: Added a `SelfLog.WriteLine` call when `NamedProperties` is `null` but `messageTemplateParameters.Length > 0`
- Added two unit tests:
  - `ProvidingMoreParametersThanNamedPropertiesInTheTemplateWillSelfLog` — verifies `SelfLog` fires when more args than named properties
  - `ProvidingParametersWhenTemplateHasNoPropertiesWillSelfLog` — verifies `SelfLog` fires for the exact scenario in #1600

## Performance

The added check (`messageTemplateParameters.Length > 0`) only executes in the early-return path when the template has no properties at all. On the normal hot path (templates with properties), there is zero overhead. This aligns with @jandresleiva's earlier benchmark analysis showing negligible impact.

## Testing

All existing tests pass. Two new tests added covering both the reported scenario and the related "more args than properties" case.

### DIFF
--- a/src/Serilog/Capturing/PropertyBinder.cs
+++ b/src/Serilog/Capturing/PropertyBinder.cs
@@ -99,7 +99,12 @@ class PropertyBinder
     {
         var namedProperties = template.NamedProperties;
         if (namedProperties == null)
+        {
+            if (messageTemplateParameters.Length > 0)
+                SelfLog.WriteLine("Parameters provided for message template with no properties: {0}", template);
+
             return NoProperties;
+        }
 
         var matchedRun = namedProperties.Length;
         if (namedProperties.Length != messageTemplateParameters.Length)
--- a/test/Serilog.Tests/Core/LogEventPropertyCapturingTests.cs
+++ b/test/Serilog.Tests/Core/LogEventPropertyCapturingTests.cs
@@ -142,6 +142,35 @@ public class LogEventPropertyCapturingTests
     }
 
     [Fact]
+    public void ProvidingMoreParametersThanNamedPropertiesInTheTemplateWillSelfLog()
+    {
+        var selfLogOutput = new List<string>();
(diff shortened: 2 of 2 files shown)


nblumhardt:
Looks great; thanks! (And thanks @jandresleiva for the earlier investigation :+1:)

PR #2227, opened 2026-04-04
Fix unreachable code in Serilog capturing and formatting

rustamque:
This PR addresses several static analyzer warnings related to unreachable code and potential `NullReferenceException` vulnerabilities:

1. **PropertyBinder.cs**: Fixed a logic bug where `result.Length` was checked instead of the actual valid properties count (`next`), causing the mismatch logging to be unreachable.
2. **PropertyValueConverter.cs**: Changed `value.ToString()` to `value?.ToString()` to safely allow the null-coalescing fallback to execute without throwing an NRE.
3. **JsonValueFormatter.cs**: Replaced `value.ToString() ?? ""` with `value?.ToString() ?? ""` to ensure the null-coalescing operator is actually reachable and functional.

Found by Linux Verification Center (linuxtesting.org) with SVACE.

### DIFF
--- a/src/Serilog/Formatting/Json/JsonValueFormatter.cs
+++ b/src/Serilog/Formatting/Json/JsonValueFormatter.cs
@@ -549,7 +549,7 @@ public class JsonValueFormatter : LogEventPropertyValueVisitor<TextWriter, bool>
     {
         Guard.AgainstNull(value);
 
-        FormatStringValue(value?.ToString() ?? "", output);
+        FormatStringValue(value.ToString() ?? "", output);
     }
 
     static void FormatStringValue(string str, TextWriter output)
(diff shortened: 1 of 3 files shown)


src/Serilog/Formatting/Json/JsonValueFormatter.cs
```diff
@@ -549,7 +549,7 @@ static void FormatLiteralObjectValue(object value, TextWriter output)
     {
         Guard.AgainstNull(value);
 
-        FormatStringValue(value.ToString() ?? "", output);
+        FormatStringValue(value?.ToString() ?? "", output);
```
Numpsy: Doesn't the Guard.AgainstNull mean that value will never be null here?

Numpsy:
And Stringify seems to only be called with none-null inputs as well,

src/Serilog/Formatting/Json/JsonValueFormatter.cs
rustamque: You right. Sorry, removed unnecessary checks

PR #2228, opened 2026-04-16
Short-circuit mixed positional token scan in MessageTemplate

### LINKED ISSUE #2211: Minor optimization in MessageTemplate positional token validation

HarryYardley:
Fixes #2211.

Stop scanning `MessageTemplate` tokens once both positional and named properties have been seen.

This does not change the existing behavior:
- templates with only positional properties still use positional binding
- templates with mixed positional and named properties still go through malformed-template handling and use named-property binding

This just avoids continuing the scan after the result is already known.

Tested with:

dotnet test .\test\Serilog.Tests\Serilog.Tests.csproj -c Release --no-restore --filter "FullyQualifiedName~MessageTemplateTests" -v minimal

These passed on net462, net48, net8.0, net9.0, and net10.0.

I also ran the full `Serilog.Tests` project locally after the change, and it passed.


nblumhardt:
Thanks for the PR. I don't think there's much to gain, here - the rare case these changes optimize will likely regress the common cases we care about due to increased branching. Thanks again, though!

PR #2234, opened 2026-05-01
Route optional interfaces through OptionalInterfaceForwardingSink for restricted sinks

### LINKED ISSUE #2231: RestrictedSink does not forward ISetLoggingFailureListener to inner sink
**Description**

`Serilog.Core.Sinks.RestrictedSink` — the wrapper applied by `LoggerSinkConfiguration.Sink(sink, restrictedToMinimumLevel)` whenever `restrictedToMinimumLevel` is anything other than `LevelAlias.Minimum` — does not implement `ISetLoggingFailureListener`. As a result, when an inner sink supports `ISetLoggingFailureListener` (directly or via `BatchingSink`), the failure listener installed by `FailureListenerSink` (used internally by `WriteTo.FallbackChain(...)` and `WriteTo.Fallible(...)`, introduced in #2108) never reaches it.

For synchronous sinks the gap is masked: `FailureListenerSink.Emit(...)` catches exceptions and notifies the listener directly. For **async/batched sinks** that surface failures only through `ILoggingFailureListener` (notably anything built on `Serilog.Core.Sinks.Batching.BatchingSink`), the listener stays unset and failed batches are silently dropped via `SelfLog` instead of being routed to the next stage of a fallback chain.

The companion wrapper `OptionalInterfaceForwardingSink` already forwards `ISetLoggingFailureListener` correctly — `RestrictedSink` should do the same.

**Reproduction**

```csharp
var listener = new MyFailureListener();

    .WriteTo.FallbackChain(
        // Primary: a batching sink with RestrictedToMinimumLevel set
        wt => wt.MyBatchingSink(restrictedToMinimumLevel: LogEventLevel.Information),
        // Fallback
        wt => wt.File("fallback.log"))
    .CreateLogger();

// When the primary batching sink fails to flush a batch, the failure is
// SelfLog'd and dropped. The fallback never receives the events because
// the listener was never installed on the batching sink — RestrictedSink
// swallowed the SetFailureListener call.
```

Removing `restrictedToMinimumLevel` (or using `LevelAlias.Minimum`) makes the fallback work as expected, because no `RestrictedSink` wrap is created.

**Expected behavior**

`RestrictedSink` should forward `SetFailureListener` to its inner sink when that sink implements `ISetLoggingFailureListener`, mirroring the forwarding already done by `OptionalInterfaceForwardingSink`.

- Serilog `dev` (4.3.2), reproduced on `net8.0` and `net9.0`.
- Discovered while wiring `Serilog.Sinks.RabbitMQ` against the new `WriteTo.FallbackChain` / `WriteTo.Fallible` extensions from #2108.

**Additional context**

Happy to open a PR — the fix is one method (mirror of `OptionalInterfaceForwardingSink.SetFailureListener`) plus a regression test.

ArieGato:
Closes #2231.

## Summary

`Serilog.Core.Sinks.RestrictedSink` — the wrapper applied by `LoggerSinkConfiguration.Sink(sink, restrictedToMinimumLevel)` whenever `restrictedToMinimumLevel` is anything other than `LevelAlias.Minimum` — hid the inner sink's optional interfaces (`IDisposable`, `IAsyncDisposable`, `ISetLoggingFailureListener`). The most visible consequence: the failure listener installed by `FailureListenerSink` (used internally by `WriteTo.FallbackChain(...)` and `WriteTo.Fallible(...)`, introduced in #2108) never reached the inner sink.

For synchronous sinks the gap is masked: `FailureListenerSink.Emit(...)` catches exceptions and notifies the listener directly. For **async/batched sinks** that surface failures only through `ILoggingFailureListener` (notably anything built on `BatchingSink`), the listener stayed unset and failed batches were silently dropped via `SelfLog` instead of being routed to the next stage of the fallback chain.

Per maintainer feedback, the fix routes optional interfaces through the existing `OptionalInterfaceForwardingSink` (the canonical mechanism, already used in the chained-sink path) rather than re-implementing forwarding inside each wrapper. Future optional interfaces will only need updates in one place.

## Changes

- `RestrictedSink` is now a plain `ILogEventSink` — its `Dispose`/`DisposeAsync`/`SetFailureListener` shims are removed.
- `LoggerSinkConfiguration.Sink(...)` wraps the resulting `RestrictedSink` in `OptionalInterfaceForwardingSink` whenever the inner sink supports any optional interface, mirroring the existing chained-sink wiring.
- Integration test `RestrictedToMinimumLevelAllowsFailureListenerPropagation` — pins the wiring through `FallbackChain` + `restrictedToMinimumLevel`.
- Companion test `DefaultLevelAllowsFailureListenerPropagation` — pins the un-wrapped baseline.
- Added small `ListenerTrackingSink` helper for the integration tests.

## Test plan

- [x] `dotnet test test/Serilog.Tests --filter "FullyQualifiedName~FailureListenerTests"` — 4/4 passing on `net8.0`, `net9.0`, `net10.0`
- [x] Full `Serilog.Tests` suite — 815/815 passing
- [x] `Serilog.ApprovalTests` passes (no public-API drift; `RestrictedSink` is internal)
- [x] CI on `windows-latest` covers the `net48`/`net462` legs

## Discovered via

Wiring `Serilog.Sinks.RabbitMQ` against the new `WriteTo.FallbackChain` / `WriteTo.Fallible` extensions from #2108. Setting `RestrictedToMinimumLevel = Information` on the primary sink silently broke the whole fallback story, with no compile-time or runtime indication.

### DIFF
--- a/src/Serilog/Configuration/LoggerSinkConfiguration.cs
+++ b/src/Serilog/Configuration/LoggerSinkConfiguration.cs
@@ -74,11 +74,19 @@ public class LoggerSinkConfiguration
             if (restrictedToMinimumLevel != LevelAlias.Minimum)
                 SelfLog.WriteLine("Sink {0} was configured with both a level switch and minimum level '{1}'; the minimum level will be ignored and the switch level used", sink, restrictedToMinimumLevel);
 
-            sink = new RestrictedSink(sink, levelSwitch);
+            sink = new RestrictedSink(logEventSink, levelSwitch);
+            if (!OptionalInterfaceForwardingSink.SupportsAll(sink) && OptionalInterfaceForwardingSink.SupportsAny(logEventSink))
+            {
+                sink = new OptionalInterfaceForwardingSink(sink, logEventSink);
+            }
         }
         else if (restrictedToMinimumLevel > LevelAlias.Minimum)
         {
-            sink = new RestrictedSink(sink, new(restrictedToMinimumLevel));
+            sink = new RestrictedSink(logEventSink, new(restrictedToMinimumLevel));
+            if (!OptionalInterfaceForwardingSink.SupportsAll(sink) && OptionalInterfaceForwardingSink.SupportsAny(logEventSink))
+            {
+                sink = new OptionalInterfaceForwardingSink(sink, logEventSink);
+            }
         }
 
         _addSink(sink);
--- a/src/Serilog/Core/Sinks/RestrictedSink.cs
+++ b/src/Serilog/Core/Sinks/RestrictedSink.cs
@@ -14,7 +14,7 @@
 
 namespace Serilog.Core.Sinks;
 
-sealed class RestrictedSink : ILogEventSink, IDisposable
+sealed class RestrictedSink : ILogEventSink, IDisposable, ISetLoggingFailureListener
 #if FEATURE_ASYNCDISPOSABLE
     , IAsyncDisposable
 #endif
@@ -36,6 +36,14 @@ sealed class RestrictedSink : ILogEventSink, IDisposable
             return;
 
         _sink.Emit(logEvent);
+    }
+
+    public void SetFailureListener(ILoggingFailureListener failureListener)
+    {
+        if (_sink is ISetLoggingFailureListener sfl)
+        {
+            sfl.SetFailureListener(failureListener);
+        }
     }
 
     public void Dispose()
--- a/src/Serilog/Core/Sinks/RestrictedSink.cs
+++ b/src/Serilog/Core/Sinks/RestrictedSink.cs
@@ -14,10 +14,7 @@
 
 namespace Serilog.Core.Sinks;
 
-sealed class RestrictedSink : ILogEventSink, IDisposable, ISetLoggingFailureListener
-#if FEATURE_ASYNCDISPOSABLE
-    , IAsyncDisposable
-#endif
+sealed class RestrictedSink : ILogEventSink
 {
     readonly ILogEventSink _sink;
     readonly LoggingLevelSwitch _levelSwitch;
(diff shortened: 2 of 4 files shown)


nblumhardt:
Thanks for checking this out.

Since `RestrictedSink` is a wrapper, `OptionalInterfaceForwardingSink` should have this covered, and `RestrictedSink`'s implementations of `Dispose`/`DisposeAsync` can be removed.

It's a little awkward, but this works in my copy:

```csharp
        var sink = logEventSink;
        if (levelSwitch != null)
        {
            if (restrictedToMinimumLevel != LevelAlias.Minimum)
                SelfLog.WriteLine("Sink {0} was configured with both a level switch and minimum level '{1}'; the minimum level will be ignored and the switch level used", sink, restrictedToMinimumLevel);

            sink = new RestrictedSink(logEventSink, levelSwitch);
            if (!OptionalInterfaceForwardingSink.SupportsAll(sink) && OptionalInterfaceForwardingSink.SupportsAny(logEventSink))
            {
                sink = new OptionalInterfaceForwardingSink(sink, logEventSink);
            }
        }
        else if (restrictedToMinimumLevel > LevelAlias.Minimum)
        {
            sink = new RestrictedSink(logEventSink, new(restrictedToMinimumLevel));
            if (!OptionalInterfaceForwardingSink.SupportsAll(sink) && OptionalInterfaceForwardingSink.SupportsAny(logEventSink))
            {
                sink = new OptionalInterfaceForwardingSink(sink, logEventSink);
            }
        }
```

In the long run, thie idea is that for any new added interfaces, an update to `OptionalInterfaceForwardingSink` should be sufficient to get it working consistently everywhere.

How's that look to you?

ArieGato:
> How's that look to you?

That looks great. I copied the code 1:1.

At the moment `!OptionalInterfaceForwardingSink.SupportsAll(sink)` always resolves to `true`. `RestrictedSink` doesn't implement all the interfaces and probably never will. I assume you want this as defense in depth for the future. So in the strange situation where all the interfaces Do get implemented on `RestrictedSink`, the bypasses will automatically be skipped.

ArieGato:
@nblumhardt  is it okay like this?

nblumhardt (approved):
Perfect, thanks!

Issue #2235, opened 2026-05-07
Destructuring does not work on Android.NET in Release build when logging to Seq


slepmog:
**The problem**: Logging to Seq from a Release build of an Android.NET application does not destructure objects.  
Objects that do not have a custom `.Destructure.ByTransforming<>` policy will be logged as their class names, objects that do will be logged as JSON-like strings.

Logging from a Debug Android build works fine.  
Logging from non-Android environments works fine too.

**Reproduction**
Use the standard Visual Studio template for an Android application. This will create a blank .NET9 Android app.
Install the Nuget packages: `Serilog`, `Serilog.Sinks.Seq`.  
Replace the contents of `public class MainActivity : Activity` with:

```c#
private class ClassWithSpecialTransforming
{
    public int IntProperty { get; set; }
    public string StringProperty { get; set; }
}

public class ClassWithoutSpecialTransforming
{
    public int IntProperty { get; set; }
    public string StringProperty { get; set; }
}

protected override void OnCreate(Bundle? savedInstanceState)
{
    base.OnCreate(savedInstanceState);

    SetContentView(Resource.Layout.activity_main);


    Serilog.Log.Logger = new Serilog.LoggerConfiguration()
        .Destructure.ByTransforming<ClassWithSpecialTransforming>(c => new { c.IntProperty, c.StringProperty })
        .WriteTo.Seq(
            serverUrl: "your.local.seq.server:5341"
        )
        .CreateLogger();


    Serilog.Log.Logger
        .ForContext("AdditionalObject", new ClassWithSpecialTransforming() { IntProperty = 42, StringProperty = "forty two" }, destructureObjects: true)
        .ForContext("MainObject", new ClassWithoutSpecialTransforming() { IntProperty = 42, StringProperty = "not forty two" }, destructureObjects: true)
        .Information("This is the object.");

    Serilog.Log.CloseAndFlush();
}
```

Start the application on the emulator (or a real device, does not matter) in the Debug configuration.
Observe that the following entry appears in Seq (as copied from Seq via *Export - Copy raw JSON*):

```json
{
  "@t": "2026-05-07T15:29:53.6870881Z",
  "@mt": "This is the object.",
  "@m": "This is the object.",
  "@i": "cb84e0d5",
  "AdditionalObject": {
    "IntProperty": 42,
    "StringProperty": "forty two"
  },
  "MainObject": {
    "$type": "ClassWithoutSpecialTransforming",
    "IntProperty": 42,
[...]

PR #2245, opened 2026-08-07
Allow opting back in to destructuring in trimmed builds (#2235)

MsfPablo:
Follow-up to #2235, where destructuring silently stops working in .NET Android **Release** builds.

### Root cause

`src/Serilog/Serilog.targets` sets the `Serilog.Capturing.IsStructureValueSupported` feature switch to `false` whenever `PublishTrimmed` is `true`. The `ILLink.Substitutions.xml` entry then stubs the property to `false`, so `PropertyValueConverter.TryConvertStructure` bails out and every destructured value is stringified — which is exactly the reported symptom (`"MainObject": "Namespace.ClassWithoutSpecialTransforming"`, and the transformed anonymous type rendered as `"{ IntProperty = 42, ... }"`).

The .NET Android SDK enables trimming for Release builds by default, so Android users get this with no opt-out — as the reporter found, no amount of `TrimmerRootAssembly`, `TrimmerRootDescriptor`, or `[DynamicallyAccessedMembers]` on their own types can re-enable it, because the check is compiled away rather than metadata-driven. It also explains why AOT was a red herring: plain trimming is enough.

### Change

The switch now honours a `SerilogIsStructureValueSupported` MSBuild property, and only defaults to `false` when the property is unset. Applications that do preserve the metadata of the types they destructure can opt back in:

```xml
<PropertyGroup>
  <SerilogIsStructureValueSupported>true</SerilogIsStructureValueSupported>
</PropertyGroup>
```

Behaviour is unchanged for everyone who does not set the property. This is the short-term escape hatch rather than a fix for trim-safe destructuring itself, so it may complement whatever longer-term approach you have in mind.

I do not have an Android device/emulator to verify the end-to-end repro, so this is reasoned from the build wiring; happy to adjust the property name or add a docs note wherever you prefer.


nblumhardt:
Sorry about the slow turnaround on this, I'll need some time clear to dig in, but sounds like a good move.

nblumhardt:
Hi @MsfPablo,

This looks good to me, but unfortunately it seems it fails the CI build.

Let me know if it's something I can help figure out.

PR #2247, opened 2026-09-05
fix: avoid OverflowException when alignment width is int.MinValue

dualfroz:
## Defect

`MessageTemplateParser` throws `System.OverflowException` while parsing a
message template whose alignment component is `-2147483648` (`int.MinValue`).

- File: `src/Serilog/Parsing/MessageTemplateParser.cs`, line 156 (pre-fix):
  `alignmentValue = new(direction, Math.Abs(width));`

When the alignment text parses to `int.MinValue`, `int.TryParse` succeeds and
`Math.Abs(int.MinValue)` throws `OverflowException` ("Negating the minimum
value of a twos complement number is invalid.").

This violates the documented contract of `MessageTemplateParser.Parse`:

> Where the template is not syntactically valid, text tokens will be returned.

The only exception the method is documented to throw is `ArgumentNullException`
for a null template. Every other malformed alignment (`+10`, `-aa`, `10-1`,
empty) is already returned as a `TextToken`; `-2147483648` is the single
numeric value that escapes that handling and crashes instead. Because message
templates are parsed on the hot path of every log call (via the template
cache), this turns a log statement such as
`log.Information("{Value,-2147483648}", x)` into an unhandled exception.

## Reproduction

```csharp
new Serilog.Parsing.MessageTemplateParser().Parse("{Hello,-2147483648}");
// throws System.OverflowException
```

## Fix

After parsing the alignment width, reject `int.MinValue` (whose magnitude is
not representable as a positive alignment width) and return the raw text as a
`TextToken`, matching the existing handling of other non-representable
alignment values.

## How to test

```
dotnet test test/Serilog.Tests/Serilog.Tests.csproj -f net9.0 \
  --filter "FullyQualifiedName~AlignmentOfIntMinValueIsParsedAsText"
```

New test: `Serilog.Tests.Parsing.MessageTemplateParserTests.AlignmentOfIntMinValueIsParsedAsText`.

- Before the fix the test fails with `OverflowException`.
- After the fix the template parses to `TextToken("{Hello,-2147483648}")`.

Full suite: `dotnet test test/Serilog.Tests/Serilog.Tests.csproj -f net9.0`
(816 passed).

## SDK / TFM

- SDK: .NET 10.0.400 (satisfies `global.json` `10.0.100` rollForward
  latestFeature).
- Test target framework: `net9.0` (test project multi-targets
  net8.0/net9.0/net10.0; net9.0 was used here).


nblumhardt:
Thanks :+1:

Issue #2249, opened 2026-09-12
Record batch transmission latency in BatchingSink via SelfMetrics


vkuropatka:
Delivery to a remote sink is the part of logging that's least under anyone's control. Even when the application and the sink are both behaving perfectly, how long a batch takes to land depends on a long list of things that live outside the process: network latency and packet loss, DNS or TLS handshakes when connections are re-established, a proxy or load balancer in the path, the ingestion service itself being slow, throttling or rate limiting on the receiving side, a cross-region hop, etc. Locally it can be just as indirect: CPU throttling in a container, thread pool starvation, or a GC pause stretching out an otherwise quick send.

From Serilog's side all of them look the same. Batches take longer to send. The queue in BatchingSink grows, and memory grows with it. In the end events are dropped, either at `QueueLimit`, or when `RetryTimeLimit` is over. Only this last step is visible today, through `SelfLog` or `ILoggingFailureListener`. The slow degradation before it is not visible at all.

It would be very useful to see the timing of batch delivery to be able to react in timely manner when latency is growing.

`SelfMetrics` already covers the pipeline, but not the delivery side of it. The same approach can be extended to `BatchingSink`. The main part would be a duration histogram around `_targetSink.EmitBatchAsync(...)` in `BatchingSink.LoopAsync()`, something like:

```
// SelfMetrics
public static readonly Histogram<double> BatchingEmitBatchDuration = Meter.CreateHistogram<double>(
    "serilog.batching.emit_batch.duration",
    unit: "ms",
    description: "The time taken by a batched sink to accept a batch of log events.");
```
Also I would suggest adding a tag for the target sink (`batchedSink.GetType().FullName`) for the case when there are multiple sinks. I would suggest to modify `BatchingSink` code like this:

```
var startTimestamp = 0L;
try
{
    if (_currentBatch.Count == 0)
    {
        await _targetSink.OnEmptyBatchAsync().ConfigureAwait(false);
    }
    else
    {
        isEagerBatch = false;
        startTimestamp = Stopwatch.GetTimestamp();
        await _targetSink.EmitBatchAsync(_currentBatch).ConfigureAwait(false);
        RecordEmitBatchDuration(startTimestamp, error: null);
        _currentBatch.Clear();
        _batchScheduler.MarkSuccess();
    }
}
catch (Exception ex)
{
    RecordEmitBatchDuration(startTimestamp, ex);
    // ... existing handling unchanged
}
```
[...]

PR #2251, opened 2026-09-18
Record batch transmission latency in BatchingSink (#2249)

vkuropatka:
Adds `serilog.batching.emit_batch.duration` to `SelfMetrics`: a histogram recorded around `IBatchedLogEventSink.EmitBatchAsync()`, so that the time taken to deliver a batch to a remote sink can be observed and its degradation detected before events start being dropped. Batches sent during normal operation are timed, and so is the final flush when the sink is disposed.

Every measurement is tagged with `serilog.batched_sink_type`, since a pipeline commonly has more than one batched sink. Failed batches are timed too, and carry `error.type`; its absence means success, following the OpenTelemetry convention.


src/Serilog/Core/Sinks/Batching/BatchingSink.cs
```diff
@@ -199,6 +205,38 @@ async Task LoopAsync()
...
+    void RecordEmitBatchDuration(long startTimestamp, Exception? error)
+    {
+        var tags = _metricTags;
+        if (error != null)
+        {
+            tags.Add(SelfMetrics.TagNames.ErrorType, error.GetType().FullName);
+        }
+
+        // Stopwatch.GetElapsedTime() is not available on all supported target frameworks.
+        var elapsedMilliseconds = (Stopwatch.GetTimestamp() - startTimestamp) * 1000.0 / Stopwatch.Frequency;
```
nblumhardt, line 232: The project convention is to add a constant like `FEATURE_STOPWATCHGETTIMESTAMP` to the targets supporting it in `Serilog.csproj`, and then use `#if`/`#else` to select between the new API and the manual fallback, here.

It's marginally more code, but it helps us track the relative quality impact of each downlevel platform we support (and helps us move forward quickly when old targets are dropped).

src/Serilog/Core/Sinks/Batching/BatchingSink.cs
```diff
@@ -46,6 +47,10 @@ sealed class BatchingSink : ILogEventSink, IDisposable, ISetLoggingFailureListen
     readonly FailureAwareBatchScheduler _batchScheduler;
     readonly Queue<LogEvent> _currentBatch = new();
     readonly Task _waitForShutdownSignal;
+    // Identifies the target sink among the measurements provided by batched sinks.
+    // It's necessary because a logging pipeline may include more than one batched sink
+    // so an aggregate across them won't really be useful.
```
nblumhardt: Comment makes sense, but since this is the typical purpose of metric tags, we probably don't need to call it out explicitly here.

src/Serilog/Core/Sinks/Batching/BatchingSink.cs
```diff
@@ -46,6 +47,10 @@ sealed class BatchingSink : ILogEventSink, IDisposable, ISetLoggingFailureListen
...
+    readonly TagList _metricTags;
```
nblumhardt: It's not clear we'll gain any useful perf by caching this, and although most operations on tag lists maintain by-value syntax, past a certain number of tags they do mutate shared reference state.

Inlining `TagList` creation to where we record the metrics will save review time on subsequent changes, let's do that.

src/Serilog/Core/Sinks/Batching/BatchingSink.cs
```diff
@@ -199,6 +205,38 @@ async Task LoopAsync()
...
+    // observed. Failed batches are timed, too: a sink that has begun timing out is the case most worth measuring.
+    async Task EmitCurrentBatchAsync()
+    {
+        var startTimestamp = Stopwatch.GetTimestamp();
+
+        try
+        {
+            await _targetSink.EmitBatchAsync(_currentBatch).ConfigureAwait(false);
+        }
+        catch (Exception ex)
```
nblumhardt: Using an exception filter (`when (RecordEmitBatchDuration(...)` and returning `false`) is marginally better in this kind of situation - keeps unhandled exception breakpoints pointing at the original throw location, and avoids marking the stack trace with the rethrow site.

nblumhardt (commented):
Looks great, thanks! Just a couple of minor comments.

The batch emit time is likely to vary based on batch size, so will be useful when investigating batching performance. Since we know this at the point we're recording the histogram, shall we add the batch size (`serilog.batched_event_count`?) as a tag on this, too?

src/Serilog/Core/Sinks/Batching/BatchingSink.cs
vkuropatka: That makes sense. I've inlined it.

src/Serilog/Core/Sinks/Batching/BatchingSink.cs
vkuropatka: Makes sense. Removed the comment.

vkuropatka:
As a tag, every distinct batch size becomes its own set of histogram series, and which sizes occur depends on the traffic, not on the configuration - full batches under load, anything from 1 upwards while idle or draining a backlog. That's unbounded cardinality from the metrics backend's point of view, and it thins the observations out as well, leaving too few per series for the quantiles to mean much.

And one more thing. The event count only partly explains the emit time: most of it comes from the size of the serialized payload, which `BatchingSink` can't see, because serialization happens inside the target sink, and the remote sink's own work also depends on what the batch contains - a batch of 50 events carrying destructured objects can be slower than 500 plain messages, and as a tag that case is indistinguishable from a fast small batch. If we could estimate the size of a batch in bytes here, a histogram over that would give a correlation worth having, but the event count on its own doesn't carry it. So I'd suggest leaving batch size out for now - unless you'd still like the count recorded, in which case a histogram is the cheap way to do it.

src/Serilog/Core/Sinks/Batching/BatchingSink.cs
vkuropatka: No problem, done.

src/Serilog/Core/Sinks/Batching/BatchingSink.cs
vkuropatka: Alright, done it this way.

